PR #68, opened 2022-03-17
👩‍💻 Fix the deploy script

fuzzybinary:
### What and why?

Since depoyments are made after a merge of a PR on main, the commit message wasn't the "Preparing" commit message expected. So we no longer check for it, and instead rely on a package name argument.

Also added a change to add the change log for the version to the release body.

### Review checklist

### DIFF
--- a/tools/releaser/bin/deployer.dart
+++ b/tools/releaser/bin/deployer.dart
@@ -14,15 +14,28 @@ class ReleaseInfo {
   final String commitSha;
   final String package;
   final String version;
+  final String changeLog;
 
-  ReleaseInfo(this.commitSha, this.package, this.version);
+  ReleaseInfo(
+    this.commitSha,
+    this.package,
+    this.version,
+    this.changeLog,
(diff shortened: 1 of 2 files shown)


tools/releaser/bin/deployer.dart
```diff
@@ -14,25 +14,38 @@ class ReleaseInfo {
...
+  var packageName = arguments.first;
+
   final gitDir = await getGitDir();
   if (gitDir == null) {
     Logger.root.shout('💥 Could not establish your current git directory.');
     exit(1);
   }
 
-  if (!(await _validateBranchState(gitDir))) exit(1);
+  //if (!(await _validateBranchState(gitDir))) exit(1);
```
xgouchet, line 46: Should this be commented?

fuzzybinary: No! Good catch!

xgouchet (approved):

PR #83, opened 2022-04-05
💥 Move DatadogTrackingHttpClient into its own package

fuzzybinary:
### What and why?

Because there are some "breaking" changes coming to HttpClient (two new methods added to the interface) I decided it would be a good idea to pull DatadogTrackingHttpClient into it's own package. That way we can update it and have it require newer versions of Flutter, while the base DatadogSdk can remain on lower versions.

This also establishes a Datadog "plugin" interface, which may or may not remain this way in the future.

### Review checklist


fuzzybinary:
One thing of note -- the current `datadog_tracking_http_client` package is not releasable because it depends on the version of `datadog_flutter_plugin` in the repo. The order for releasing this will be:

1. Release `datadog_flutter_plugin` 1.0.0-alpha.2
2. Update `datadog_tracking_http_client`'s pubspec.yaml to require 1.0.0-alpha.2
3. Release `datadog_tracking_http_client`

packages/datadog_tracking_http_client/.flutter-plugins
```diff
@@ -0,0 +1,2 @@
+# This is a generated file; do not edit or check into version control.
```
0xnm, line 1: seems it was checked into version control :)

packages/datadog_tracking_http_client/.flutter-plugins-dependencies
```diff
@@ -0,0 +1 @@
+{"info":"This is a generated file; do not edit or check into version control.","plugins":{"ios":[{"name":"datadog_flutter_plugin","path":"/Users/jeff.ward/Projects/dd-sdk-flutter/packages/datadog_flutter_plugin/","dependencies":[]}],"android":[{"name":"datadog_flutter_plugin","path":"/Users/jeff.ward/Projects/dd-sdk-flutter/packages/datadog_flutter_plugin/","dependencies":[]}],"macos":[],"linux":[],"windows":[],"web":[{"name":"datadog_flutter_plugin","path":"/Users/jeff.ward/Projects/dd-sdk-flutter/packages/datadog_flutter_plugin/","dependencies":[]}]},"dependencyGraph":[{"name":"datadog_flutter_plugin","dependencies":[]}],"date_created":"2022-04-05 10:15:03.720598","version":"2.10.3"}
```
0xnm, line 1: should this one be in VCS?

packages/datadog_tracking_http_client/.flutter-plugins
fuzzybinary: 🤔 Why isn't it in the default .gitignore I wonder...

0xnm (approved):

PR #94, opened 2022-04-13
🚀 Release of datadog_tracking_http_client 1.0.0-alpha.1

fuzzybinary:
Merge develop into main for release of datadog_tracking_http_client.


packages/datadog_tracking_http_client/example/android/build.gradle
```diff
@@ -0,0 +1,31 @@
+buildscript {
+    ext.kotlin_version = '1.6.10'
+    repositories {
+        google()
+        mavenCentral()
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:4.1.0'
```
0xnm, line 9: I think we can use newer version given Kotlin is already at `1.6.10`, so let's try to use `7.1.x`.

fuzzybinary: 👍 Since it's only in the example I'll do that as a seperate PR.

PR #98, opened 2022-04-15
Create version of the HTTP tracking client that supports 2.17

fuzzybinary:
### What and why?

Dart 2.17 will add two methods to HttpClient which we need to override. Currently, only developers using Flutter's `beta` channel will need this, but I'd rather stay ahead of it.

I'm keeping this on a release branch as a majority of users are going to be on Dart 2.16 until Flutter 2.12. for a while. Once Flutter 2.12 goes GA, we'll merge to master and likely close this release branch.

### How?

Change version solving to put Dart 2.17 as the minimum version. 

### Review checklist


packages/datadog_tracking_http_client/lib/src/tracking_http_client.dart
```diff
@@ -164,6 +164,16 @@ class DatadogTrackingHttpClient implements HttpClient {
     return request;
   }
 
+  @override
+  set connectionFactory(
+          Future<ConnectionTask<Socket>> Function(
+                  Uri url, String? proxyHost, int? proxyPort)?
+              f) =>
```
0xnm, line 171: should this be called just `f`?

0xnm (approved):

packages/datadog_tracking_http_client/lib/src/tracking_http_client.dart
fuzzybinary: Yes, it's a straight override of the HttpClient function which has the terrible name.

packages/datadog_tracking_http_client/lib/src/tracking_http_client.dart
```diff
@@ -164,6 +164,16 @@ class DatadogTrackingHttpClient implements HttpClient {
...
+      innerClient.connectionFactory = f;
+
+  @override
+  set keyLog(Function(String line)? callback) => innerClient.keyLog = callback;
```
ueman, line 175: Just FYI, you can do something like the following code, in order to support 2.16 and 2.17
```suggestion
  // ignore: override_on_non_overriding_member
  @override
  set connectionFactory(
      Future<ConnectionTask<Socket>> Function(
              Uri url, String? proxyHost, int? proxyPort)?
          f) {
    try {
      (innerClient as dynamic).connectionFactory = f;
    } on NoSuchMethodError catch (_) { }
  }

  // ignore: override_on_non_overriding_member
  @override
  set keyLog(Function(String line)? callback) {
    try {
      (innerClient as dynamic).keyLog = callback;
    } on NoSuchMethodError catch (_) { }
  }
```

fuzzybinary: Hrm, good call. I always get a little bit nervous about casting to dynamic, but might be a good stop-gap to make maintaining the two versions easier.

ueman: Yeah, I agree that it's not the best solution, but it's probably worth it. Also, you can easily test this, and it also just needs to be tested on Dart >= 2.17 as the methods aren't available on earlier versions.

fuzzybinary:
Actually re-changed my mind.  The published pubspec already only supports up to Dart 2.16, so I'm going to have to push a new version anyway to support 2.17, and I prefer having the safety of the checks by not casting to dynamic.

packages/datadog_tracking_http_client/lib/src/tracking_http_client.dart
ueman: You're welcome

fuzzybinary:
Note for anyone that might be waiting on this -- The CI failures are due to this issue: https://github.com/flutter/flutter/issues/102330, but it's unlikely to be fixed as it appears to not be caused by the Flutter framework.

I'm actively looking for a solution which bypasses the iOS os.activity framework.

kayayarai (commented):
I checked with @alai97 , and no docs review required on the READMEs involved at this time.

PR #100, opened 2022-04-18
✨ Adding a gRPC interceptor

fuzzybinary:
### What and why?

Adding a gRPC client interceptor, which checks adds span/trace headers and calls start / stop resource loading properly.  I need to check that some of the metadata is correct and add more tests before this is ready to be pulled in.

### How?

Generates a random trace / spanId as two 32-bit random numbers, then combines them into a BigInt. BigInt's performance is not the best, but after looking at the code for this particular use case, we'd be performing very similar if not identical operations so I think it's safe to use in this case.

Headers I need to check -- whether the 'path' is okay to send for the 'grpc.method' call or if I need to find another parameter, and if I've put the trace / span Ids in the proper metadata fields.

### Review checklist


packages/datadog_grpc_interceptor/README.md
```diff
@@ -0,0 +1,40 @@
+
+# Datadog gRPC Interceptor / Plugin
+
+> A plugin for use with the DatadogSdk, used to track performance of gRPC and enable Datadog Distributed Tracing.
```
0xnm, line 4:
```suggestion
> A plugin for use with the `DatadogSdk`, used to track performance of gRPC and enable Datadog Distributed Tracing.
```

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
```diff
@@ -0,0 +1,102 @@
...
+import 'package:datadog_flutter_plugin/datadog_flutter_plugin.dart';
+import 'package:datadog_grpc_interceptor/src/tracing_util.dart';
+import 'package:grpc/grpc.dart';
+import 'package:uuid/uuid.dart';
+
+class DatadogTracingHeaders {
+  static const traceId = 'x-datadog-trace-id';
+  static const parentId = 'x-datadog-parent-id';
+
+  static const sampled = 'x-datadog-sampled';
```
0xnm, line 16: I guess we can remove this one as the outcome in the APM discussion

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
```diff
@@ -0,0 +1,102 @@
...
+    final path = method.path;
+    final String fullPath;
+    if (host is InternetAddress) {
+      fullPath = '${host.host}:${_channel.port}$path';
+    } else {
+      fullPath = '$host:${_channel.port}$path';
+    }
+
+    bool shouldAppendTraces = _datadog.traces != null &&
+        _datadog.isFirstPartyHost(Uri.parse(fullPath));
```
0xnm, line 61: should we add it do the doc saying that host should be registered as first-party during the SDK configuration?

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
```diff
@@ -0,0 +1,102 @@
...
+    String? parentId;
+    if (shouldAppendTraces) {
+      traceId = generateTraceId();
+      parentId = generateTraceId();
+    }
+
+    final rumKey = uuid.v1();
+    _datadog.rum?.startResourceLoading(
+      rumKey,
+      RumHttpMethod.get,
```
0xnm, line 73: why it is always `GET`?

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
```diff
@@ -0,0 +1,102 @@
...
+        DatadogTracingHeaders.origin: 'rum',
+        DatadogTracingHeaders.samplingPriority: '1',
+        DatadogTracingHeaders.traceId: traceId!,
+        DatadogTracingHeaders.parentId: parentId!,
+      }));
+    }
+
+    final future = invoker(method, request, options);
+    future.then((v) {
+      _datadog.rum?.stopResourceLoading(rumKey, 200, RumResourceType.native);
```
0xnm, line 95: that is actually a question to think about: should we use `native` here or not. In RN we use `xhr` type (despite the fact that HTTP transport in RN is backed by native libs anyway).

packages/datadog_grpc_interceptor/lib/src/tracing_util.dart
```diff
@@ -0,0 +1,17 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'dart:math';
+
+final random = Random();
+
+String generateTraceId() {
+  final highBits = random.nextInt(1 << 32);
```
0xnm, line 10: I'm not quite sure about trace ID format, but this implementation is different from https://github.com/DataDog/browser-sdk/blob/bb0a5780ea2f6ca092262a01524b957fe4f7c966/packages/rum-core/src/domain/tracing/tracer.ts#L127, should we align? wdyt?

packages/datadog_grpc_interceptor/test/protos/helloworld.proto
```diff
@@ -0,0 +1,38 @@
...
+
+// The request message containing the user's name.
+message HelloRequest {
+  string name = 1;
+}
+
+// The response message containing the greetings
+message HelloReply {
+  string message = 1;
+}
```
0xnm, line 38:
```suggestion
}

```

packages/datadog_grpc_interceptor/test/src/generated/helloworld.pb.dart
```diff
@@ -0,0 +1,105 @@
+///
```
0xnm, line 1: if this is generated, should we commit it? same for files below

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
fuzzybinary: A yes, thanks!

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
fuzzybinary: Yes. Good call.

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
fuzzybinary: This is an interesting question. 

gRPC doesn't use Http Methods and isn't technically a "resource," but RUM doesn't really have a way of tagging something as an API request (that I know of) so I just defaulted it to GET. We have no way of knowing if a particular gRPC call is mutating or not. Though, technically, if you're running through an HTTP Proxy (instead of a server that supports HTTP2) it uses POST so though, so maybe I should change to that, but I'm not sure it matters.

In reality, this may be a larger discussion for both GraphQL and gRPC -- do we want to have a way for RUM to differentiate between resource requests (like images) and API calls (REST, GraphQL, gRPC)?

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
fuzzybinary: Another good question. When should we use xhr vs. native? To be honest, I think I used xhr in some places in the tracking_http_client.

packages/datadog_grpc_interceptor/lib/src/tracing_util.dart
fuzzybinary: Trace ID sent to APM is supposed to be a 64-bit unsigned int, converted to decimal (although, I think RUM wants it in hex... I should double check that).

I think it's the same in theory, just not in practice. JS is storing 63-bits in a Uint8Array than working down digit by digit through the combined 32-bit integers to create the string. This is doing the same, I'm just using Dart's BigInt abstraction to do the heavy lifting for me, but if you look at its internals it does exactly that (and allow me to use the full 64-bits as an unsigned int. Dart only supports 64-bit signed through it's native int type).

packages/datadog_grpc_interceptor/test/src/generated/helloworld.pb.dart
fuzzybinary: In this case yes, I think so, since this isn't generated through the Dart build system and requires a both a separate tool (protoc) and a global pub package to generate it.  Also I should add that to the README.

For what it's worth, most other repos I've seen commit their generated gRPC classes.

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
0xnm: I see. In this case maybe we can do RFC and collaborate with Browser team?

fuzzybinary: Yeah probably a good idea. Probably thinking about gRPC and GraphQL at the same time will make sense so I should coordinate with @xgouchet to see where he ended up on GraphQL.

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
0xnm: It was all `xhr` until not so long time ago: last September we added `native` type (https://github.com/DataDog/dd-sdk-android/pull/702 and https://github.com/DataDog/dd-sdk-ios/pull/605) to differentiate the calls made from native side vs cross-platform.

RN is using `xhr`, because calls made there are essentially backed by `XMLHttpRequest` (which is just an interface, and HTTP stack is implemented by OkHttp anyway on Android, for example).

With Flutter we may still have native modules in the user app, which will do `native` calls, but then the question is how do we categorize Flutter calls.

fuzzybinary: I'll leave it as native for now. I think that's the most accurate?

0xnm: Having `native` is fine imo for now. But let's also have a TODO item for a further discussion, if we want to separate native requests from flutter requests.

0xnm (approved):

PR #106, opened 2022-04-29
💥 Deprecate tracing.

fuzzybinary:
### What and why?

Because of issues with the os.activity library on iOS, and because tracing may be re-done through all RUM soon, we are deprecating tracing, and we'll likely remove it from alpha.4.

### How?

In most cases, we're keeping the code for now, but skipping any tests associated with tracing. This includes removing e2e monitors for tracing.

I've run all tests on Flutter 2.13 beta and they appear to work.

### Review checklist


packages/datadog_flutter_plugin/e2e_test_app/integration_test/test_utils.dart
```diff
@@ -42,14 +42,19 @@ Future<void> initializeDatadog([DatadogConfigCallback? configCallback]) async {
...
+    [double targetSeconds = 0.02]) async {
+  var stopwatch = Stopwatch();
+  stopwatch.start();
   await callback();
-  span?.finish();
+  stopwatch.stop();
+  // ignore: unused_local_variable
+  final elapsedSeconds = stopwatch.elapsedMicroseconds / 1000000.0;
+  // TODO: Determine best way to monitor this moving forward
+  if (elapsedSeconds > targetSeconds) {
```
0xnm, line 54: How precise is `Stopwatch`? Say we have 21 ms vs expected 20 ms, can we see that often?

0xnm (approved):
Fastest deprecation I've even seen :)

packages/datadog_flutter_plugin/e2e_test_app/integration_test/test_utils.dart
xgouchet: My guess is that it's at least accurate to the microsecond

xgouchet (approved):

PR #119, opened 2022-05-11
✨ Allow setting a sample rate for tracing

fuzzybinary:
### What and why?

This add support for setting a trace sampling rate for both the datadog_tracking_http_client and datadog_grpc_interceptor. A sample rate sets the probability that a given network request will include a sampling priority of '1' (meaning it should sample) and send trace and spanIds.

There is also a minor change to send a sampling priority of '0' on first part calls when network request does not pass the sample test.

### Review checklist


packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -179,15 +181,25 @@ class RumConfiguration {
   ///
   /// The sampling rate must be a value between `0.0` and `100.0`. A value of
   /// `0.0` means no RUM events will be sent, `100.0` means all sessions will be
-  /// kept
+  /// sent
   ///
   /// Defaults to `100.0`.
   double sampleRate;
```
0xnm, line 187: I think it is better to rename this one to `sessionSamplingRate` for the better clarity (but it is already part of public API though).

packages/datadog_flutter_plugin/test/rum/ddrum_test.dart
```diff
@@ -33,4 +37,89 @@ void main() {
     final other = ContentType.parse('application/octet-stream');
     expect(resourceTypeFromContentType(other), RumResourceType.native);
   });
+
+  test('Tracing sampling rate is clamped to 0..100', () {
+    final lowConfiguration = RumConfiguration(
+      applicationId: 'applicationId',
+      sampleRate: -12.3,
```
0xnm, line 44: from the description, `sampleRate` is for the session, so I guess here we should pass `tracingSampleRate`?

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -179,15 +181,25 @@ class RumConfiguration {
...
+  /// Defaults to `100.0`.
+  double tracingSamplingRate;
+
   RumConfiguration({
     required this.applicationId,
-    this.sampleRate = 100.0,
-  });
+    double sampleRate = 100.0,
+    this.tracingSamplingRate = 100.0,
+  }) : sampleRate = max(0, min(sampleRate, 100));
```
0xnm, line 202: I'm not 100% clear what this constructor does (but that is because I don't know Dart). I suppose it defines sampling values if they are not provided explicitly in the configuration argument. But then why we do `max-min` boundary only for `sampleRate`, but not for `tracingSamplingRate`?

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
fuzzybinary: Yeah, I'll deprecate it and map it to a new variable for at least one release.

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
fuzzybinary: No, you got it... this looks like a bug in both the test and the implementation. I was mixing up sample rate and tracingSampleRate, which is exactly why I should take your suggestion above and rename sample rate.

fuzzybinary:
I've renamed sessionRate and added tests that check for it being clamped, as well as properly testing that tracingSampling rate is properly clamped.

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -175,24 +177,42 @@ class RumConfiguration {
...
+
   /// Sets the sampling rate for RUM Sessions.
   ///
   /// The sampling rate must be a value between `0.0` and `100.0`. A value of
   /// `0.0` means no RUM events will be sent, `100.0` means all sessions will be
-  /// kept
+  /// sent
+  ///
+  /// Defaults to `100.0`.
+  double sessionSampleRate;
```
0xnm, line 194: very nit: wording is a big different between `sessionSampleRate` and `tracingSamplingRate`, we should choose between `sample` and `sampling` for the consistency.

0xnm (commented):
lgtm so far, but as I understand it will be some updates here with default rate at least.

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
fuzzybinary: Ugh good call. Thanks!

xgouchet (approved):
LGTM

maycmlee (approved):
👍  for docs.

PR #120, opened 2022-05-11
✨ Support Datadog iOS console logging in Flutter Developer console

fuzzybinary:
### What and why?

Override `consolePrint` to send console logging up to the Flutter developer console

Also fixes indentation in the main Datadog file from 2 spaces to 4. (Recommend turning on ignore whitespace when reviewing)

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/ios/Classes/SwiftDatadogSdkPlugin.swift
+++ b/packages/datadog_flutter_plugin/ios/Classes/SwiftDatadogSdkPlugin.swift
@@ -8,154 +8,162 @@ import Datadog
 import DatadogCrashReporting
 
 public class SwiftDatadogSdkPlugin: NSObject, FlutterPlugin {
-  let channel: FlutterMethodChannel
+    let channel: FlutterMethodChannel
 
-  // NOTE: Although these are instances, they are still registered globally to
-  // a method channel. That might be something we want to change in the future
-  public private(set) var logs: DatadogLogsPlugin?
-  public private(set) var tracer: DatadogTracesPlugin?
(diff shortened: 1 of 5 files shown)


packages/datadog_flutter_plugin/ios/Classes/SwiftDatadogSdkPlugin.swift
```diff
@@ -8,154 +8,162 @@ import Datadog
...
+            _ = ddConfigBuilder.trackURLSession(firstPartyHosts: Set(flutterConfig.firstPartyHosts))
+        }
+
+        if let customEndpoint = flutterConfig.customEndpoint,
+           let customEndpointUrl = URL(string: customEndpoint) {
+            _ = ddConfigBuilder
+                .set(customLogsEndpoint: customEndpointUrl)
+                .set(customTracesEndpoint: customEndpointUrl)
+                .set(customRUMEndpoint: customEndpointUrl)
+        }
```
maxep, line 150: I see that this is not a change from this PR, but shouldn't we have different URLs on `flutterConfig` as well? one for each product.

maxep (approved):
LGTM 👍

packages/datadog_flutter_plugin/ios/Classes/SwiftDatadogSdkPlugin.swift
fuzzybinary: 🤔 It's probably something I'll have to add? At the moment it's only used for integration testing where I haven't needed to set custom URLs per-product. Not sure about how it works with proxies though (and I think we're missing other information to get proxies to work.)

PR #123, opened 2022-05-12
Build Changes for Flutter 3

fuzzybinary:
### What and why?

This sets us up to build on latest stable Flutter (3.0) as well as configures new conditional runs of portions of the build similar to iOS.

In order to setup Flutter 3, we're pulling the 1.0.1-alpha release datadog_tracking_http_client package back into develop.

Note: Builds for this PR will fail until I switch over the bitrise build rules (which should happen after other branches are merged in)

### Review checklist


packages/datadog_tracking_http_client/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
-## Unreleased
+## 1.0.1-beta.1
```
0xnm, line 3: I'm curious why not `1.0.0-beta.2`. Does it mean that from now on we will continue only with `1.0.1`?

0xnm (approved):

packages/datadog_tracking_http_client/CHANGELOG.md
fuzzybinary: Because there's a possibility people will want / need to stay on Flutter 2.10. If I use beta.2 I can't continue support for the pre-Flutter 3.0 version line.

Until Flutter 2.10 is not being used, I'll likely have to maintain a 1.0.0 and a 1.0.1 release line. (TBH, it probably should have been 1.1.0, but that ship has sailed)

0xnm: Thanks, it is clear now!

tools/ci/src/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    Once set, the ENV will be available to `envman` running in caller process.
+    """
+    if not dry_run:
+        shell_output(f'envman add --key {key} --value "{value}"')
+    else:
+        print(f'[DRY-RUN] calling: `envman add --key {key} --value "{value}"`')
+
+
+# Copied from `tools/nightly-unit-tests/src/utils.py`
+# TODO: RUMM-1860 Share this code between both tools
```
ncreated, line 22:
```suggestion
```

tools/ci/src/ci_context.py
```diff
@@ -0,0 +1,115 @@
...
+
+@dataclass
+class TriggerENVs:
+    """
+    ENVs defined for a trigger in `bitrise.yml`.
+    """
+    DD_RUN_UNIT_TESTS: str
+    DD_RUN_INTEGRATION_TESTS: str
+    DD_RUN_SMOKE_TESTS: str
+    DD_RUN_TOOLS_TESTS: str
```
ncreated, line 29: We can get rid of SMOKE and TOOLS tests envs in this file.

PR #131, opened 2022-05-26
✨ Add simple automatic gesture detection

fuzzybinary:
### What and why?

This adds a widget that is capable of performing simple common gesture detection (mostly only capturing taps).

### How?

On detection of a tap (pointer down / pointer up), the listener queries the element tree for widgets that are and enabled and would respond to a tap gesture (Buttons, InkWell, GestureDetector). It then looks for a description of the widget by either taking inner text (for Buttons only) consulting any Semantics child widgets, or using the last RumUseActionAnnotation widget it saw.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
@@ -4,7 +4,6 @@
 
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
-import 'package:flutter/rendering.dart';
 import 'package:meta/meta.dart';
 
 import '../../datadog_flutter_plugin.dart';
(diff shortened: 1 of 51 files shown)


xgouchet (approved):

packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+  }
+
+  void _onPerformActionAt(Offset position, RumUserActionType action) {
+    final elementDescription = _getDetectingElementAtPosition(position);
+
+    if (elementDescription != null) {
+      widget.rum?.addUserAction(
+          RumUserActionType.tap, elementDescription.elementDescription);
```
0xnm, line 127: `action` argument should be used here

```suggestion
          action, elementDescription.elementDescription);
```

packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  void _onPointerUp(PointerUpEvent event) {
+    if (_lastPointerDownLocation != null && event.pointer == _lastPointerId) {
+      final distanceOffset = Offset(
+          _lastPointerDownLocation!.dx - event.localPosition.dx,
+          _lastPointerDownLocation!.dy - event.localPosition.dy);
+
+      final distanceSquared = distanceOffset.distanceSquared;
+      if (distanceSquared < _tapSlopSquared) {
```
0xnm, line 116: so for now we only detect `tap` events? In this case it is contrary to the `RumUserActionDetector` description, which says we can also detect swipes.

packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
```diff
@@ -0,0 +1,272 @@
...
+// Copyright 2019-Present Datadog, Inc.
+
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:meta/meta.dart';
+
+import '../../datadog_flutter_plugin.dart';
+
+// The distance a 'pointer' can move and still be considered a tap.
+const _tapSlop = 20;
```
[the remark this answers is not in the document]
fuzzybinary: It should be logical display units. I'll add it to the comment.

packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
```diff
@@ -0,0 +1,272 @@
...
+/// look for a child [Semantics] object, or an [Icon] with its [semanticsLabel]
+/// property set.
+///
+/// Alternately, you can enclose any Widget tree with a
+/// [RumUserActionAnnotation], which will use the provided description when
+/// reporting user actions detected in the child tree, without changing the
+/// Semantics of the tree.
+class RumUserActionDetector extends StatefulWidget {
+  @internal
+  static final elementMap = <RumUserActionDetector, Element>{};
```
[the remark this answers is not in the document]
fuzzybinary: Yes, it has to be static. Widgets and State are automatically connected by Flutter, and widgets / elements get recreated all the time, whereas State is long lived. A new Widget object will not always create a new State object, only the first time it's constructed.

If the widget tree above the RumUserActionDetector, it will create a new one, which it will connect with the existing StatefulElement / State object and call "didUpdateWidget." But, Widgets and State don't have a good way to get their Elements, so I need a global map of what Widgets map to what Elements, hence the static Map.

Technically, the static map could probably be on the State though.

packages/datadog_flutter_plugin/lib/src/rum/rum_gesture_detector.dart
fuzzybinary: Yes it only detects taps at the moment. I'd planned to do swipes but didn't get to it so I'll remove it from the docs until I actually get around to it.

0xnm (approved):
I have a concern about possible PII capture and that labels can be localized, but let's give it a try (at least it will be a foundation for the next work).

fuzzybinary:
Yeah, we'll have to see. That's why I limited it to Buttons, which would still have the localization problem though...

Issue #144, opened 2022-06-08
[`datadog_tracking_http_client`] Reduce dart dependency version


danielgomezrico:
**Is your feature request related to a problem? Please describe.**
When I include `datadog_tracking_http_client 1.0.1-beta.1` in my pubspec it fails because of:

```console
Because client depends on datadog_tracking_http_client >=1.0.1-beta.1 which requires SDK version >=2.17.0-0 <3.0.0, version solving failed.
pub get failed (1; Because client depends on datadog_tracking_http_client >=1.0.1-beta.1 which requires SDK version >=2.17.0-0 <3.0.0, version solving failed.)
``` 

**Describe the solution you'd like**
Reduce the min version of dart required by this package

**Additional context**
This version of dart is required for flutter 3.0 but not for flutter 2, so currently this is only usable from flutter 3

fuzzybinary:
I've updated the README for dart_http_tracking_client to clarify the Dart 2.16 / Flutter 2.10 support:  https://github.com/DataDog/dd-sdk-flutter/tree/develop/packages/datadog_tracking_http_client#flutter-210-support

Long term, the 1.0.x line will continue to support Dart 2.16 and 1.1.0 will be Dart 2.17

PR #145, opened 2022-06-08
Allow `datadog_tracking_http_client` to be used from flutter 2.10.3

danielgomezrico:
### What and why?

Without this this package can only be used by flutter 3 users

It fixes: 
- https://github.com/DataDog/dd-sdk-flutter/issues/144

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_tracking_http_client/pubspec.yaml
+++ b/packages/datadog_tracking_http_client/pubspec.yaml
@@ -6,7 +6,7 @@ homepage: https://datadoghq.com
 
 
 environment:
-  sdk: ">=2.17.0-0 <3.0.0"
+  sdk: ">=2.16.1 <3.0.0"
   flutter: ">=1.17.0"
 
 dependencies:


danielgomezrico:
I can not test with both plugins from git, having:
```YAML
datadog_flutter_plugin:
    git:
      path: packages/datadog_flutter_plugin
      url: https://github.com/danielgomezrico/dd-sdk-flutter.git
      ref: patch-1
  datadog_tracking_http_client:
    git:
      path: packages/datadog_tracking_http_client
      url: https://github.com/danielgomezrico/dd-sdk-flutter.git
      ref: patch-1
```

It breaks with:
```console
Because every version of datadog_tracking_http_client from git depends on datadog_flutter_plugin from hosted and client depends on datadog_flutter_plugin from git, datadog_tracking_http_client from git is forbidden.
So, because client depends on datadog_tracking_http_client from git, version solving failed.
pub get failed (1; So, because client depends on datadog_tracking_http_client from git, version solving failed.)
```

scopendo:
> I can not test with both plugins from git

Try adding a pubspec_overrides.yaml:

```yaml
dependency_overrides:

  datadog_flutter_plugin:
    git:
      path: packages/datadog_flutter_plugin
      url: https://github.com/danielgomezrico/dd-sdk-flutter.git
      ref: patch-1

  datadog_tracking_http_client:
    git:
      path: packages/datadog_tracking_http_client
      url: https://github.com/danielgomezrico/dd-sdk-flutter.git
      ref: patch-1
```

If that doesn't work, you should be able to use a relative path, just to prove the fix locally.

fuzzybinary:
Hi @danielgomezrico, see me comment on your issue, but the version constraint for 1.0.1-beta.1 is correct. It won't work below Flutter 3 / Dart 2.17. However, version 1.0.0-beta.1 should support Flutter 2.

fuzzybinary:
Hi @danielgomezrico -- I've updated documentation in the README for the datadog_tracking_http_client package to be clearer about our Flutter 2.10 / Dart 2.16 support - https://github.com/DataDog/dd-sdk-flutter/tree/develop/packages/datadog_tracking_http_client#flutter-210-support

To answer your question about why we override `dart:io.HttpClient` instead of `http.Client` -- we chose to hook in at the lower level `HttpClient` in order to catch all resources / http calls automatically, including requests from internal Flutter classes like Image.network. Overriding at this level will also catch calls from http.Client, but also catches other libraries like Dio and others. 

If you would rather only have specific resources tracked (e.g. only those called from a specific a `http.Client`) please let us know and we'll look into getting  an `http.Client` override package on the roadmap.

I'm going to close this PR as simply lowering the Dart version constraint won't work.

PR #166, opened 2022-07-08
💥 Remove deprecated traces feature

fuzzybinary:
### What and why?

We've had a deprecation warning in about traces for a while, makes sense to remove as we get closer to 1.0.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


packages/datadog_flutter_plugin/android/ndk_crash_reports_v2/network_information
```diff
@@ -0,0 +1 @@
+{"ev_size":40}{"connectivity":"network_not_connected"}
```
0xnm, line 1: how this file is used? it seems it shouldn't be here?

0xnm (approved):

packages/datadog_flutter_plugin/android/ndk_crash_reports_v2/network_information
fuzzybinary: Ugh, it shouldn't. Thanks for catching it.

PR #167, opened 2022-07-14
[datadog_grpc_interceptor Adds support for grpc hosts that don't have a URL scheme

acoutts:
Add support for grpc channels that don't have a URL scheme.

### DIFF
--- a/packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
+++ b/packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
@@ -64,7 +64,7 @@ class DatadogGrpcInterceptor extends ClientInterceptor {
       fullPath,
       {
         "grpc.method": method.path.split('/').lastWhere(
-              (e) => e == '/',
+              (e) => e.isNotEmpty,
               orElse: () => method.path,
             ),
         "grpc.path": method.path,
--- a/packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
+++ b/packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
@@ -60,10 +60,14 @@ class DatadogGrpcInterceptor extends ClientInterceptor {
     final rumKey = uuid.v1();
     _datadog.rum?.startResourceLoading(
       rumKey,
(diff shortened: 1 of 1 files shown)


packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
```diff
@@ -50,10 +60,14 @@ class DatadogGrpcInterceptor extends ClientInterceptor {
...
       rumKey,
-      RumHttpMethod.get,
+      RumHttpMethod.unknown,
       fullPath,
       {
-        "grpc.method": method.path,
+        "grpc.method": method.path.split('/').lastWhere(
+              (e) => e.isNotEmpty,
+              orElse: () => method.path,
+            ),
```
fuzzybinary, line 69: So, I'm not as well versed on gRPC -- I this generally correct that we should only support sending last portion of the path as the method name?

fuzzybinary (commented):
Thanks for the contribution @acoutts!

Would you be able fix the failing unit tests as well as add tests for this scenario?

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
```diff
@@ -50,10 +60,14 @@ class DatadogGrpcInterceptor extends ClientInterceptor {
     final rumKey = uuid.v1();
     _datadog.rum?.startResourceLoading(
       rumKey,
-      RumHttpMethod.get,
+      RumHttpMethod.unknown,
```
fuzzybinary, line 63: RumHttpMethod.unknown actually translates to GET under the hood. I'm removing it for the next release to avoid confusion See #170

packages/datadog_grpc_interceptor/lib/src/datadog_grpc_interceptor.dart
fuzzybinary: Looking into this -- it does not seem like this is correct in the general case. I'd rather leave the path as the full path, which should include the class and method name of the call.

fuzzybinary (changes requested):

fuzzybinary:
Closing in favor of #209

PR #169, opened 2022-07-19
Don't attempt to reinitialize Datadog on hot-reload

### LINKED ISSUE #141: SDK is already initialised
**Describe what happened**
An error message "Datadog SDK usage error: SDK is already initialized." appears when the app is restarted following hot reload.  Whilst this can be ignored, it seems to me that it would be better suppressed given that it is such a commonality.

```
flutter: 🔥 Datadog SDK usage error: SDK is already initialized.
flutter: 🔥 Datadog SDK usage error: The `Tracer` instance was already created. Use existing `Global.sharedTracer` instead of initializing the `Tracer` another time.
flutter: 🔥 Datadog SDK usage error: The `RUMMonitor` instance was already created. Use existing `Global.rum` instead of initializing the `RUMMonitor` another time.
```

**Additional context**
 - Dart/Flutter version: Dart 2.17.1 / Flutter 3.0
 - Datadog SDK version: datadog_flutter_plugin: ^1.0.0-beta.2

fuzzybinary:
### What and why?

Because hot reload / hot restart can recall `main`, you can get multiple requests to initialize Datadog, and the warning we spit to the console can get annoying. This removes that warning if your initialization hasn't changed.

If your initialization has changed, this shows a warning.

This fixes #141 

### Review checklist

- [x] Run unit tests
- [x] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
@@ -5,6 +5,7 @@
  */
 package com.datadoghq.flutter
 
+import android.util.Log
 import androidx.annotation.NonNull
 import com.datadog.android.Datadog
 import com.datadog.android.rum.GlobalRum
@@ -27,6 +28,7 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
 
     private lateinit var channel: MethodChannel
     private lateinit var binding: FlutterPlugin.FlutterPluginBinding
+    private var previousConfiguration: DatadogFlutterConfiguration? = null
 
     // Only used to shutdown Datadog in debug builds
     private val executor: ExecutorService = ThreadPoolExecutor(
@@ -54,7 +56,13 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
                 val configArg = call.argument<Map<String, Any?>>("configuration")
                 if (configArg != null) {
                     val config = DatadogFlutterConfiguration(configArg)
-                    initialize(config)
+                    if (!Datadog.isInitialized()) {
+                        initialize(config)
+                        previousConfiguration = config
+                    } else if (config != previousConfiguration) {
+                        // Maybe use DevLogger instead?
+                        Log.e(DATADOG_FLUTTER_TAG, MESSAGE_INVALID_REINITIALIZATION)
+                    }
                 }
                 result.success(null)
             }
@@ -119,7 +127,7 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         }
     }
 
-    fun invokePrivateShutdown(result: Result) {
+    internal fun invokePrivateShutdown(result: Result) {
         executor.execute {
             simpleInvokeOn("flushAndShutdownExecutors", Datadog)
             simpleInvokeOn("stop", Datadog)
@@ -149,3 +157,9 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         rumPlugin = null
     }
 }
+
+internal const val DATADOG_FLUTTER_TAG = "DatadogFlutter"
+
+internal const val MESSAGE_INVALID_REINITIALIZATION =
+    "🔥 Reinitialziing the DatadogSDK with different options, even after a hot restart, is not" +
+        "supported. Cold restart your application to change your current configuation."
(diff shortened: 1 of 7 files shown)


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -149,3 +157,9 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         rumPlugin = null
     }
 }
+
+internal const val DATADOG_FLUTTER_TAG = "DatadogFlutter"
+
+internal const val MESSAGE_INVALID_REINITIALIZATION =
+    "🔥 Reinitialziing the DatadogSDK with different options, even after a hot restart, is not" +
+        "supported. Cold restart your application to change your current configuation."
```
0xnm, line 165:
```suggestion
        " supported. Cold restart your application to change your current configuation."
```

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogSdkPluginTest.kt
```diff
@@ -157,6 +162,87 @@ class DatadogSdkPluginTest {
...
+                    "nativeCrashReportEnabled" to true,
+                    "loggingConfiguration" to null
+                )
+            )
+        )
+        val mockResultB = mock<MethodChannel.Result>()
+        plugin.onMethodCall(methodCallB, mockResultB)
+
+        // THEN
+        io.mockk.verify(exactly = 1) { Log.e(any(), match { it.contains("🔥") }) }
```
0xnm, line 243: I think we can remove `any()` and `match()` calls here and reference directly tag and message, because they are `internal` and should be available.

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/ReflectUtils.kt
```diff
@@ -58,6 +58,7 @@ private fun <T : Any> T.getDeclaredMethodRecursively(
         val lookingInClass = classesToSearch.removeAt(0)
         classesSearched.add(lookingInClass)
         method = try {
+            var methods = lookingInClass.declaredMethods
```
0xnm, line 61: is not used it seems

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogSdkPluginTest.kt
```diff
@@ -157,6 +162,87 @@ class DatadogSdkPluginTest {
         verify(mockResult).success(null)
     }
 
+    @Test
+    fun `M not issue warning W initialize called with same configuration`(
+        forge: Forge
+    ) {
+        // GIVEN
+        mockkStatic(Log::class)
```
0xnm, line 170: personally I'm not big fan of static mocks, because to mock static bytecode modification is needed (this is what `mockk` is doing under the hood, probably).

maybe we can do non-static wrapper around `Log`?

Anyway, is not a blocker for me, just a personal opinion :)

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogSdkPluginTest.kt
```diff
@@ -30,11 +30,15 @@ import io.flutter.embedding.engine.plugins.FlutterPlugin
...
 import io.flutter.plugin.common.MethodChannel
+import io.mockk.mockkStatic
 import io.opentracing.util.GlobalTracer
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.extension.ExtendWith
+import org.mockito.MockedStatic
+import org.mockito.Mockito
+import org.mockito.Mockito.mockStatic
```
0xnm, line 41: seems like those 3 are not used. And if Mockito has `mockStatic`, then why do we bring `mockk`?

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/ReflectUtils.kt
fuzzybinary: Ah, thanks. I was trying to figure out how to call something via reflection and forgot to remove this.

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogSdkPluginTest.kt
fuzzybinary: Ugh, yeah (I'm surprised ktlint and detekt didn't call this out....).

It's because Mockito's mockStatic doesn't work with Android. I don't know the particulars but when you try to use it throws an exception. mockk's version works with Android (and to be honest, I kinda like mockk better... I might switch everything over at some point).

PR #218, opened 2022-09-02
💥 Deprecate DdFlutterConfiguration.customEndpoint

fuzzybinary:
### What and why?

Custom endpoint was being used to set the logs and rum endpoint to the same endpoint, but we need capabilities to set each feature's endpoint separately.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/CHANGELOG.md
+++ b/packages/datadog_flutter_plugin/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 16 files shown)


fuzzybinary:
Something interesting is that iOS and Android approach this differently. iOS expects the entire endpoint, whereas Android only wants the host name and appends the path itself...

For now I'm putting a special case in the example code but I may look into unifying the APIs in the SDK at some point.

xgouchet (approved):

PR #220, opened 2022-09-07
✨ Add specifying application version during configuration

fuzzybinary:
### What and why?

Adding this options adds custom version support.
In order to be compliant with Datadog tags, we replace '+' with '-' automatically. The `datadog-ci flutter-symbols upload` will do this as well to be consistent.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


xgouchet (approved):

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -262,6 +263,25 @@ class DdSdkConfiguration {
...
+  /// By default, both iOS and Android sync their version numbers with the
+  /// current version in your pubspec, minus any build or pre-release
+  /// information. This property should only be used if you want to add this
+  /// additional information back in, or if the version in your pubspec does not
+  /// match your application version.
+  ///
+  /// Because 'version' is a Datadog tag, it needs to comply with the rules in
+  /// [Defining
+  /// Tags](https://docs.datadoghq.com/getting_started/tagging/#defining-tags)
+  /// Datadog documentation. We will automatically replace `+` with `-` for
```
0xnm, line 277: Should we say that if some other extra characters are presented in the version which don't comply with tag rules, they will be converted to underscores by the native SDKs?

0xnm (approved):

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
fuzzybinary: Does it do that for versions actually?

0xnm: yes, in case of logs version will be in the event in `ddtags` attribute which will be sanitized per rules.

For RUM it is different, because `version` is a part of query params there.

fuzzybinary: I think since we already link the documentation there, we should be okay. I'm basically only documenting the non-default replacement case.

PR #243, opened 2022-09-22
✅ Fix Flutter web tests, specifically around long tasks

fuzzybinary:
### What and why?

The browser sends any stall on the main thread more than 50ms, and this is not configurable. So for our integration tests on Flutter we check that all are above 50ms instead of 100ms, and check that at least one is above the 200ms threshold.

Flutter web tests still have some flakiness that I can't pin down but this fixes the major issue.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/integration_test_app/lib/auto_integration_scenarios/rum_auto_instrumentation_scenario.dart
+++ b/packages/datadog_flutter_plugin/integration_test_app/lib/auto_integration_scenarios/rum_auto_instrumentation_scenario.dart
@@ -18,7 +18,7 @@ class RumAutoInstrumentationScenario extends StatefulWidget {
 class _RumAutoInstrumentationScenarioState
     extends State<RumAutoInstrumentationScenario> {
   final images = [
-    'https://placekitten.com/300/300',
+    'https://imgix.datadoghq.com/img/about/presskit/kit/press_kit.png',
     'https://imgix.datadoghq.com/img/about/presskit/kit/press_kit.png'
   ];
 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -24,8 +24,8 @@
         },
         {
             "name": "auto instrumented app",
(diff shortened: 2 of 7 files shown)


plousada (approved):

packages/datadog_flutter_plugin/integration_test_app/lib/auto_integration_scenarios/rum_auto_instrumentation_scenario.dart
```diff
@@ -18,7 +18,7 @@ class RumAutoInstrumentationScenario extends StatefulWidget {
 class _RumAutoInstrumentationScenarioState
     extends State<RumAutoInstrumentationScenario> {
   final images = [
-    'https://placekitten.com/300/300',
+    'https://imgix.datadoghq.com/img/about/presskit/kit/press_kit.png',
```
0xnm, line 21: why do we reference the same image twice?

0xnm (approved):

packages/datadog_flutter_plugin/integration_test_app/lib/auto_integration_scenarios/rum_auto_instrumentation_scenario.dart
plousada: I had the same question but assumed it wasn't relevant for the scenario for the 2 images to be the same.

xgouchet (approved):

packages/datadog_flutter_plugin/integration_test_app/lib/auto_integration_scenarios/rum_auto_instrumentation_scenario.dart
fuzzybinary: Yeah the image isn't relevant to the test, and for web you have to grab an image from a server that has its CORS headers set, which placekitten doesn't.

PR #262, opened 2022-10-27
✨ Add DatadogClient for tracking http.Client calls

fuzzybinary:
### What and why?

This is an alternative tracking method to DatadogHttpClient, which overrides dart:io's HttpClient. DatadogClient can instead be composed with http.Client for use with other packages that provide the same interface (notably cronet_http and cupertino_http)

### How?

This works similarly to DatadogHttpClient, though the interface provided by http.Client is simpler. On a call to `send`, we intercept the call and add our own headers, passing through everything else about the call. Then we spy on the stream looking for errors and waiting for the 'close' call before sending out that the resource was stopped.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests


packages/datadog_tracking_http_client/README.md
```diff
@@ -16,6 +16,33 @@ final configuration = DdSdkConfiguration(
 )..enableHttpTracking()
 ```
 
+## Using http.Client wrapping
+
+This package also supplies a composable client usable with the [http pub package](https://pub.dev/packages/http) called `DatadogClient`. For most scenarios, Datadog recommends you use the HTTP tracking method above, but there are a few scenarios where using `DatadogClient` might make more sense:
+
+* If you are using native HTTP libraries like [cronet_http](https://pub.dev/packages/cronet_http) or [cupertino_http](https://pub.dev/packages/cupertino_http), which do not work with the above tracking method 
+* It you only want to track specific resource requests.
```
hestonhoffman, line 24:
```suggestion
* If you only want to track specific resource requests.
```

packages/datadog_tracking_http_client/README.md
```diff
@@ -16,6 +16,33 @@ final configuration = DdSdkConfiguration(
...
+  // specifying firstPartyHosts is still necessary to
+  // enable distributed tracing
+  firstPartyHosts: ['example.com'],
+)
+
+final httpClient = http.Client()
+final datadogClient = DatadogClient(datadogSdk: DatadogSdk.instance, innerClient: httpClient);
+```
+
+The `innerClient` parameter is optional. If it is not supplied, DatadogClient will create a `Client` for you with default options.  Now, you can use `datadogClient` to make any network calls, and they will be tracked.
```
hestonhoffman, line 44:
```suggestion
The `innerClient` parameter is optional. If it is not supplied, `DatadogClient` creates a `Client` for you with default options.  After configuring the `Client`, you can use `DatadogClient` to make any network calls, and they will be tracked.
```

hestonhoffman (approved):
Couple small edits

packages/datadog_tracking_http_client/README.md
```diff
@@ -16,6 +16,33 @@ final configuration = DdSdkConfiguration(
 )..enableHttpTracking()
 ```
 
+## Using http.Client wrapping
+
+This package also supplies a composable client usable with the [http pub package](https://pub.dev/packages/http) called `DatadogClient`. For most scenarios, Datadog recommends you use the HTTP tracking method above, but there are a few scenarios where using `DatadogClient` might make more sense:
+
+* If you are using native HTTP libraries like [cronet_http](https://pub.dev/packages/cronet_http) or [cupertino_http](https://pub.dev/packages/cupertino_http), which do not work with the above tracking method
```
0xnm, line 23:
```suggestion
* If you are using native HTTP libraries like [cronet_http](https://pub.dev/packages/cronet_http) or [cupertino_http](https://pub.dev/packages/cupertino_http), which do not work with the above tracking method. 
```

packages/datadog_tracking_http_client/lib/src/tracking_http.dart
```diff
@@ -0,0 +1,204 @@
...
+import 'package:datadog_flutter_plugin/datadog_internal.dart';
+import 'package:flutter/foundation.dart';
+import 'package:http/http.dart' as http;
+import 'package:uuid/uuid.dart';
+
+/// A composable client for use with the `http` package that supports tracking
+/// network requests and sending them to Datadog.
+///
+/// If the RUM feature is enabled, the SDK will send information about RUM
+/// Resources (calling startLoading, stopLoading, and stopLoadingWithErrorInfo)
```
0xnm, line 16:
```suggestion
/// Resources (calling startResourceLoading, stopResourceLoading, and stopResourceLoadingWithError)
```

packages/datadog_tracking_http_client/lib/src/tracking_http.dart
```diff
@@ -0,0 +1,204 @@
...
+          rumKey = null;
+        }
+      }
+
+      rethrow;
+    }
+
+    if (rumKey != null) {
+      try {
+        // Copy the response so we can spy on the stream. Use
```
0xnm, line 110: Is some sentence not finished here? Or it is ok?

packages/datadog_tracking_http_client/lib/src/tracking_http.dart
```diff
@@ -0,0 +1,204 @@
...
+      rumKey = _uuid.v1();
+      bool shouldSample = isFirstParty && rum.shouldSampleTrace();
+      final attributes =
+          _appendRequestHeaders(request, rumKey, isFirstParty, shouldSample);
+
+      final rumHttpMethod = rumMethodFromMethodString(request.method);
+      rum.startResourceLoading(
+          rumKey, rumHttpMethod, request.url.toString(), attributes);
+    } catch (e, st) {
+      datadogSdk.internalLogger.sendToDatadog(
```
0xnm, line 76: How often this and another `catch` blocks below can happen? If it can happen just because there is no network connection, won't we spam our telemetry too much?

0xnm (commented):
I'm not an expert in the different Flutter HTTP clients, but tried to add my questions/suggestions (quite generic though).

packages/datadog_tracking_http_client/lib/src/tracking_http.dart
fuzzybinary: Does appear there's an extra word but I don't think there was more to say.

0xnm (approved):

PR #265, opened 2022-10-28
✨ Add performance metric tracking

fuzzybinary:
### What and why?

Add a configuration option, `reportFlutterPerformance` that uses `SchedulerBinding.addTimingsCallback` to report build and raster times for views. These are sent to Datadog as "performance metrics" for each view.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
@@ -238,18 +238,7 @@ class DatadogRumPlugin(
                     val buildTimes = call.argument<List<Double>>(PARAM_BUILD_TIMES)
                     val rasterTimes = call.argument<List<Double>>(PARAM_RASTER_TIMES)
                     if(buildTimes != null && rasterTimes != null) {
-                        buildTimes.forEach {
-                            rum?._getInternal()?.updatePerformanceMetric(
-                                RumPerformanceMetric.FLUTTER_BUILD_TIME,
-                                it
-                            )
-                        }
-                        rasterTimes.forEach {
-                            rum?._getInternal()?.updatePerformanceMetric(
-                                RumPerformanceMetric.FLUTTER_RASTER_TIME,
-                                it
-                            )
-                        }
+                        updatePerformanceMetrics(buildTimes, rasterTimes)
                         result.success(null)
                     } else {
                         result.missingParameter(call.method)
@@ -265,6 +254,24 @@ class DatadogRumPlugin(
                 mapOf(
                     "methodName" to call.method
                 )
+            )
+        }
+    }
+
+    private fun updatePerformanceMetrics(
+        buildTimes: List<Double>,
+        rasterTimes: List<Double>
+    ) {
+        buildTimes.forEach {
+            rum?._getInternal()?.updatePerformanceMetric(
+                RumPerformanceMetric.FLUTTER_BUILD_TIME,
+                it
+            )
+        }
+        rasterTimes.forEach {
+            rum?._getInternal()?.updatePerformanceMetric(
+                RumPerformanceMetric.FLUTTER_RASTER_TIME,
+                it
             )
         }
     }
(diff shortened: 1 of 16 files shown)


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -231,6 +234,27 @@ class DatadogRumPlugin(
                         result.missingParameter(call.method)
                     }
                 }
+                "updatePerformanceMetrics" -> {
+                    val buildTimes = call.argument<List<Double>>(PARAM_BUILD_TIMES)
+                    val rasterTimes = call.argument<List<Double>>(PARAM_RASTER_TIMES)
+                    if(buildTimes != null && rasterTimes != null) {
```
0xnm, line 240: why do we check both in the same condition and not independently?

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/ContractHelpers.kt
```diff
@@ -57,6 +58,7 @@ data class Contract(
         return when(type) {
             SupportedContractType.STRING -> forge.anExtendedAsciiString()
             SupportedContractType.MAP -> forge.exhaustiveAttributes()
+            SupportedContractType.LIST -> forge.aList(0) {}
```
0xnm, line 61: We should generate a list with actual context or just use `emptyList` here. I guess it should be non-empty list though.

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogRumPluginTest.kt
```diff
@@ -454,6 +455,33 @@ class DatadogRumPluginTest {
         verify { mockResult.success(null) }
     }
 
+    @Test
+    fun `M call internal updatePerformanceMetrics W updatePerformanceMetrics is called`(
```
0xnm, line 459: we need to add the test where one metric is there, but another is missing, because the check is `if (buildTimes != null && rasterTimes != null)`

0xnm (commented):
lgtm overall, I've just added few questions

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
fuzzybinary: Simplicity more than anything. For the most part, these two parameters not being null is enforced by the type system in Dart, and contract checks in unit tests. If either one is null, it's a contract violation and reported to telemetry.  I could report which individual parameter violated the contract, but I haven't done that anywhere yet so this pattern is through the whole file.

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/ContractHelpers.kt
fuzzybinary: I'll use emptyList. I'm afraid of forging values since the caller could be looking for a List of any type. Since the contract tests don't look at the values anyway other than to cast them, an empty list should work fine.

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogRumPluginTest.kt
fuzzybinary: The contract tests do that. They test that the call succeeds if both parameters are present, and fails if either one is missing (it checks once for each parameter).  See: https://github.com/DataDog/dd-sdk-flutter/blob/6d789792ffe212ba8f60c609aa5a96a9a3409afd/packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/ContractHelpers.kt#L66

0xnm (approved):

PR #269, opened 2022-11-01
✨ Add better support to NavigationObserver for AppState scenarios

fuzzybinary:
### What and why?

NavigationObserver now automatically starts / stops views when the app goes into / returns from the background. 

By utilizing this in conjunction with a "pending route" when the app is in the background, we can more easily support "add to app" scenarios that use "prewarmed" flutter engines.

Update documentation to remove warnings about using prewarmed flutter engines.

### How?

When a Flutter View is dismissed and not visible, Flutter automatically sets the AppState to "detached," and sets the app state to "resumed" when it comes back. This is similar to what happens when an app is backgrounded (though backgrounded apps don't usually get the to the "detached" state.) By letting NavigationObserver automatically start and stop views this way, we can support pre-cached Flutter engines easily.

Also added examples of using predicates to remove FlutterViewController and FlutterActivity from sessions, though on iOS this results in the automatic view tracking missing the presenting view controller from coming back into focus. This is documented in the example readme.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests


xgouchet (approved):

examples/native-hybrid-app/README.md
```diff
@@ -22,20 +22,27 @@ void main() async {
...
+Depending on your settings, the automatic view tracking in the native iOS and Android SDKs will automatically track the presentation of the `FlutterViewController` and `FlutterActivity`/`FlutterFragment` when they appear, and then immediately add show a view load for your tracked Flutter view.
 
-For this reason, the example currently only uses non-prewarmed and non-cached versions of the Flutter engine.
+To avoid seeing the extra `FlutterViewController` and `FlutterActivity` views in your sessions, add a `UIKitRUMViewsPredicate` and `ComponentPredicate` to your iOS and Android code, respectively.  Very simple examples of these predicates can be found in the example code in [AppDelegate.swift](ios/iOS%20Flutter%20Hybrid%20Example/AppDelegate.swift) and [HybridApplication.kt](android/app/src/main/java/com/datadoghq/hybrid_flutter_example/HybridApplication.kt)
+
+### Restart iOS Views
+
+If you are using automatic view tracking on iOS and filtering `FlutterViewController` views with the predicate above, there is a known issue that transitioning back from the FlutterViewController does not restart the original view, so you have to perform this step manually.
```
ncreated, line 33:
```diff
- back from the FlutterViewController does not restart
+ back from the `FlutterViewController` does not restart
```

examples/native-hybrid-app/ios/iOS Flutter Hybrid Example/FirstViewController.swift
```diff
@@ -14,9 +15,29 @@ class FirstViewController: UIViewController {
...
-        GeneratedPluginRegistrant.register(with: flutterViewController.engine!);
+        let flutterEngine = (UIApplication.shared.delegate as! AppDelegate).flutterEngine
+        let flutterViewController = FlutterViewController(engine: flutterEngine, nibName: nil, bundle: nil)
+        flutterViewController.modalPresentationStyle = .overFullScreen
+        flutterViewController.pushRoute("/")
+        
+        DismissMethodCallHandler.shared.setDismissListener {
+            self.dismiss(animated: true) {
+                Global.rum.startView(viewController: self)
+            }
```
ncreated, line 26: Note that it leaks `self` as we capture a strong reference in dismiss listener. Using `[weak self]` could solve the problem if it's not intended.

ncreated (commented):
Glad it works 🥇. I left few smaller feedbacks 👌.

Doing a step back and referring to:
> on iOS this results in the automatic view tracking missing the presenting view controller from coming back into focus

we could re-call the predicate from [this code path](https://github.com/DataDog/dd-sdk-ios/blob/a3c14772ed7912760c06bbf4ff7b0ea594c90f2d/Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift#L173-L177):

```swift
private func applicationDidEnterBackground() {
    if let current = stack.last {
        stop(view: current)
    }
}
```
Because it's small impact and will additionally align iOS views tracking with Android, I think it could be a viable option to consider. Would it help  @fuzzybinary ?

fuzzybinary:
I just tested and you're correct - backgrounding and foregrounding re-starts the native view controller on the top.

I don't think re-calling the predicate will work -- but we may be able to take advantage of the fact that an unidentifiable view doesn't start:
https://github.com/DataDog/dd-sdk-ios/blob/a3c14772ed7912760c06bbf4ff7b0ea594c90f2d/Sources/Datadog/RUM/Instrumentation/Views/RUMViewsHandler.swift#L143-L145

If the Predicate can return a `RUMView` that we can then convert to a `RUMViewsHandler.View` that is not identifyable, it will take a slot up in the view stack but not have any view events tracked.

examples/native-hybrid-app/ios/iOS Flutter Hybrid Example/FirstViewController.swift
fuzzybinary: Probably not an issue as the block gets discarded once it's called, but definitely better practice to hold onto it weak.

fuzzybinary:
@ncreated Here's a pretty minimal change that we could use to do this: https://github.com/DataDog/dd-sdk-ios/commit/514f8efdb8d1e36b36b23d855a010a52ab9efccc

ncreated:
> @ncreated Here's a pretty minimal change that we could use to do this: [DataDog/dd-sdk-ios@514f8ef](https://github.com/DataDog/dd-sdk-ios/commit/514f8efdb8d1e36b36b23d855a010a52ab9efccc)

Oh, thanks for trying it out 💪!

However, this seems more impactful than I thought. I was thinking that it is possible to provide uniform abstraction over RUM views auto-tracking that will entirely hide OS-specific effects of background / foreground transitioning on iOS and Android. In other words: from the cross-platform SDK point of view both native SDKs will behave exactly the same. 

But it seems that the iOS-specifics of fe-/bg- transitioning must be also handled in the host (Flutter) app. For that reason I think it's better to only patch one component (Flutter SDK) instead of two (Flutter + iOS SDK). The solution provided in this PR is minimal and explained very well 👍.

PR #288, opened 2022-12-06
✨ Logs mapping API

fuzzybinary:
### What and why?

Allows mapping of logs to remove unwanted logs or remove sensitive information

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ Code style is enforced with the following libraries in the following languages:
 * Flutter - we use the included Flutter analyzer and linter. If you can, set up
   your IDE to format your Dart files on save, which will keep you in conformance
   with the linter
-* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to`
+* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to
   run as part of the build in XCode. If you want Swiftlint to autoformat your
   files for you and fix any potential errors, run `swiftlint --fix`
 * Android / Kotlin - We have both [ktlint](https://github.com/pinterest/ktlint)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,9 @@ them with Datadog. Running `./generate_env.sh` will create `example/.env` and
 `integration_test_app/.env`, both of which should be modified with your Client
 Id and Application Id from the Datadog RUM setup.
 
+Lastly, be sure to run `flutter pub run build_runner build` in projects that need
+their JSON models generated, which includes `packages/datadog_flutter_plugin`
+
 ## Code Style
 
 Code style is enforced with the following libraries in the following languages:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ them with Datadog. Running `./generate_env.sh` will create `example/.env` and
 Id and Application Id from the Datadog RUM setup.
 
 Lastly, be sure to run `flutter pub run build_runner build` in projects that need
-their JSON models generated, which includes `packages/datadog_flutter_plugin`
+their JSON models generated, which includes `packages/datadog_flutter_plugin`.
 
 ## Code Style
 
@@ -35,7 +35,7 @@ Code style is enforced with the following libraries in the following languages:
 * Flutter - we use the included Flutter analyzer and linter. If you can, set up
   your IDE to format your Dart files on save, which will keep you in conformance
   with the linter
-* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to
+* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to`
   run as part of the build in XCode. If you want Swiftlint to autoformat your
   files for you and fix any potential errors, run `swiftlint --fix`
 * Android / Kotlin - We have both [ktlint](https://github.com/pinterest/ktlint)
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
@@ -5,19 +5,22 @@
  */
 package com.datadoghq.flutter
 
+import android.os.Handler
+import android.os.Looper
 import android.util.Log
 import androidx.annotation.NonNull
 import com.datadog.android.Datadog
-import com.datadog.android._InternalProxy
 import com.datadog.android.event.EventMapper
+import com.datadog.android.log.model.LogEvent
 import com.datadog.android.rum.GlobalRum
-import com.datadog.android.telemetry.model.TelemetryConfigurationEvent
+import com.google.gson.JsonObject
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler
 import io.flutter.plugin.common.MethodChannel.Result
 import java.util.concurrent.ExecutorService
+import java.util.concurrent.Semaphore
 import java.util.concurrent.SynchronousQueue
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
@@ -175,21 +178,21 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
     }
 
     fun initialize(config: DatadogFlutterConfiguration) {
-        val configurationBuilder = config.toSdkConfiguration()
+        val configBuilder = config.toSdkConfigurationBuilder()
         val credentials = config.toCredentials()
 
-        _InternalProxy.setTelemetryConfigurationEventMapper(
-            configurationBuilder,
-            object : EventMapper<TelemetryConfigurationEvent> {
-                override fun map(event: TelemetryConfigurationEvent): TelemetryConfigurationEvent? {
-                    return mapTelemetryConfiguration(event)
+        if (config.attachLogMapper) {
+            configBuilder.setLogEventMapper(
+                object : EventMapper<LogEvent> {
+                    override fun map(event: LogEvent): LogEvent? {
+                        return mapLogEvent(event)
+                    }
                 }
-            }
-
-        )
+            )
+        }
 
         Datadog.initialize(
-            binding.applicationContext, credentials, configurationBuilder.build(),
+            binding.applicationContext, credentials, configBuilder.build(),
             config.trackingConsent
         )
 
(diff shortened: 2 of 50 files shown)


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -200,6 +215,84 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+    private fun mapLogEvent(event: LogEvent): LogEvent? {
+        val jsonEvent = event.toJson().asMap()
+
+        var modifiedJson: Map<String, Any?>? = null
+
+        val semaphore = Semaphore(1)
+        semaphore.acquire()
+
+        val handler = Handler(Looper.getMainLooper())
+        handler.post {
```
0xnm, line 228: ah, this is where the switch to the main thread comes

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -200,6 +215,84 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+
+                        override fun notImplemented() {
+                            Datadog._internal._telemetry.error(
+                                "mapLogEvent returned notImplemented."
+                            )
+                            semaphore.release()
+                        }
+                    }
+                )
+            } catch (t: Throwable) {
```
0xnm, line 258: should we narrow down it to `Exception`? Because if it is `Error`, there is something wrong with the runtime itself and we shouldn't prevent it from crashing (`Error` mostly likely will crash everything, unlikely that the code up in the stack will catch it).

But anyway, I'm ok with `Throwable` as well (as I understand it is so wide to release semaphore in any case).

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -200,6 +215,84 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+                Datadog._internal._telemetry.error("Attempting call mapLogEvent failed.", t)
+                semaphore.release()
+            }
+        }
+        // Stalls until the method channel finishes
+        semaphore.acquire()
```
0xnm, line 264: alternatively we can use `CountdownLatch(1)` which has `await` method which can take a timeout (to avoid cases waiting forever, if that can happen) and will reduce 2 `acquire` calls to the single `await` here (and in the `post` we need to call only `countDown`).

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -200,6 +215,84 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+
+        try {
+            var modifiedEvent: LogEvent? = null
+            modifiedJson?.let {
+                modifiedEvent = LogEvent.fromJsonObject(it.toJsonObject())
+            }
```
0xnm, line 270: minor: a bit more clean imo, but this is opinionated (benefits from smart cast + val vs var):

```suggestion
            val modifiedEvent = if (modifiedJson != null) {
                LogEvent.fromJsonObject(modifiedJson.toJsonObject())
            } else {
                null
            }
```

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -200,6 +215,84 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+
+
+            // Replace any modifyable info with the new event
+            return modifiedEvent?.let {
```
0xnm, line 274: minor: also here we can simply use `return if` instead of lambda usage.

0xnm (commented):
nice! I reviewed a bit the Kotlin part. Solution looks good to me!

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
fuzzybinary: 🤔 You may be right. I'm okay switching to Exception as that should probably catch anything that's recoverable anyway.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
fuzzybinary: 👍 I'll look into it.

CONTRIBUTING.md
```diff
@@ -25,14 +25,17 @@ them with Datadog. Running `./generate_env.sh` will create `example/.env` and
...
+
 ## Code Style
 
 Code style is enforced with the following libraries in the following languages:
 
 * Flutter - we use the included Flutter analyzer and linter. If you can, set up
   your IDE to format your Dart files on save, which will keep you in conformance
   with the linter
-* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to
+* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to`
```
0xnm, line 38:
```suggestion
* iOS / Swift - [Swiftlint](https://github.com/realm/SwiftLint) is configured to
```

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -272,6 +286,81 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+                    event.message = modifiedEvent.message
+                    event.ddtags = modifiedEvent.ddtags
+                    event.logger.name = modifiedEvent.logger.name
+
+                    event.additionalProperties.clear()
+                    event.additionalProperties.putAll(modifiedEvent.additionalProperties)
+                    event
+                }
+            } else {
+                null
```
0xnm, line 354: why if exception happens we return original event, but in this branch we return `null`? shouldn't we be consistent with all "unhappy" paths?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -272,6 +286,81 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         isRegistered.set(false)
     }
 
+    @Suppress("TooGenericExceptionCaught")
+    private fun mapLogEvent(event: LogEvent): LogEvent? {
```
0xnm, line 290: is it possible to add tests for this logic? or it will be too complicated? I can help if needed.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -272,6 +286,81 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+                    event.additionalProperties.clear()
+                    event.additionalProperties.putAll(modifiedEvent.additionalProperties)
+                    event
+                }
+            } else {
+                null
+            }
+        } catch (e: Exception) {
+            Datadog._internal._telemetry.error(
+                "Attempting deserialized mapped log event failed, returning unmodified event", e
```
0xnm, line 358:
```suggestion
                "Attempt to deserialize mapped log event failed, returning unmodified event", e
```

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -272,6 +286,81 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+            } catch (e: Exception) {
+                Datadog._internal._telemetry.error("Attempting call mapLogEvent failed.", e)
+                latch.countDown()
+            }
+        }
+        // Stalls until the method channel finishes
+        latch.await(1, TimeUnit.SECONDS)
+
+        try {
+            return if (modifiedJson != null) {
```
0xnm, line 337: indeed here because `modifiedJson` is `var` smart cast won't work, I missed that during my previous suggestion here.

Then maybe we can use something like `return modifiedJson?.let { ... } ?: event` or `return modifiedJson?.let { ... }` if returning `null` is ok?

Or we can live with `!!` as well (but I guess it will be flagged by tools like Detekt).

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -383,6 +385,9 @@ class DdSdkConfiguration {
   /// logging is disabled.
   LoggingConfiguration? loggingConfiguration;
 
+  //
```
0xnm, line 388: missing doc

packages/datadog_flutter_plugin/lib/src/logs/ddlog_event.dart
```diff
@@ -0,0 +1,143 @@
...
+    required this.architecture,
+  });
+
+  factory LogDevice.fromJson(Map<String, dynamic> json) =>
+      _$LogDeviceFromJson(json);
+  Map<String, dynamic> toJson() => _$LogDeviceToJson(this);
+}
+
+@commonJsonOptions
+class LogEventDd {
```
0xnm, line 45: if we end new properties on the native side to `LogEvent`, will things still work? Or will it be the crash? Also the question comes how to we keep the two in sync.

0xnm (changes requested):
I've checked Kotlin code and Dart as well and I have few questions/suggestions. But overall looks nice!

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
fuzzybinary: I'll give it a shot. The complexity of the different dependencies makes it difficult which is why I focused on the integration test.

0xnm (approved):

jtappa (approved):

maxep (approved):
LGTM!
I'm hoping to provide a `DictionaryCoder` like implementation as a shared utility, I will keep you posted!

PR #294, opened 2022-12-14
Support OTel b3s and b3m header injection

fuzzybinary:
### What and why?

This allows users to automatically generate OTel headers from any of the HTTP Interceptors (DatadogTrackingHttpClient, DatadogClient and the gRPC interceptor).

Additionally, this allows users to supply their own b3s or b3m headers in requests, and the interceptors will read them out and add those IDs to Datadog Resource attributes to properly support distributed tracing.

### How?

In the case of the DatadogTrackingHttpClient, we had to move assignment of the Datadog Resource atttributes to the `stopResourceLoading` call, since a user may not have set their b3 headers yet during the openUrl call.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests


fuzzybinary:
@maciejburda This is mostly done. I need to add for gRPC still but http should be ready to go.

xgouchet (approved):

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
```diff
@@ -130,87 +249,80 @@ void main() {
...
       when(() => mockDatadog.isFirstPartyHost(any())).thenReturn(true);
 
       final interceptor = DatadogGrpcInterceptor(mockDatadog, channel);
 
       final stub = GreeterClient(channel, interceptors: [interceptor]);
+      // Randomly generated
+      //  - 61ffb765f4f77e3b == 7061564389269667387
+      //  - 3eb5c1bcb46ab916 == 4518730817361066262
+      final options = CallOptions(metadata: {
+        'x-b3-traceid': '000000000000000061ffb765f4f77e3b',
```
maciejburda, line 262: I remember seeing somewhere that some implementations might be case sensitive. So it should be both `X-B3-TradeId` and value should use capital letters `000000000000000061FFB765F4F77E3B` and so on.

maciejburda (commented):
Gave it a first quick look.

For the public API do you allow to specify headers per endpoint for firstPartyHosts?

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
fuzzybinary: For grpc specifically, the headers need to be lowercase (I think because all gRPC metadata fields are lowercased)
https://github.com/openzipkin/b3-propagation#grpc-encoding

I'll fix the hex IDs to generate capital letters.

fuzzybinary:
@maciejburda 

> For the public API do you allow to specify headers per endpoint for firstPartyHosts?

No, I've added a ticket to remember to add that ability.

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
maciejburda: Yes you're - sorry for not double checking on my side.

maciejburda (approved):

PR #295, opened 2022-12-15
👷 Create a ./prepare.sh script

fuzzybinary:
### What and why?

The prepare script runs `flutter pub get` in all of the sub-packages, generates `.env` files, and runs code generation is any sub-packages that need it.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,13 +20,14 @@ Make sure you have installed the [Flutter
 SDK](https://docs.flutter.dev/get-started/install), and that `flutter doctor`
 passes without issues.
 
-You will also need to generate `.env` files for the various apps in order to use
-them with Datadog. Running `./generate_env.sh` will create `example/.env` and
-`integration_test_app/.env`, both of which should be modified with your Client
-Id and Application Id from the Datadog RUM setup.
+Next, run the `prepare.sh` script in the root of the repo. This will run
+`flutter pub get` on all of the packages in this repo, generate necessary files
+with `flutter pub run build_runner build`, and generate `.env` files for the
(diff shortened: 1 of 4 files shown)


xgouchet (approved):

CONTRIBUTING.md
```diff
@@ -20,13 +20,14 @@ Make sure you have installed the [Flutter
...
-`integration_test_app/.env`, both of which should be modified with your Client
-Id and Application Id from the Datadog RUM setup.
+Next, run the `prepare.sh` script in the root of the repo. This will run
+`flutter pub get` on all of the packages in this repo, generate necessary files
+with `flutter pub run build_runner build`, and generate `.env` files for the
+various apps in order to use them with Datadog. 
 
-Lastly, be sure to run `flutter pub run build_runner build` in projects that need
-their JSON models generated, which includes `packages/datadog_flutter_plugin`.
+Running `./generate_env.sh` or `./prepare.sh` will create `.env` files in the
```
0xnm, line 28: Shouldn't we mention `./prepare.sh` here only for the end user? Otherwise it may be confusing. Also `./generate_env.sh` is mentioned in `HybridApplication.kt` and in `AppDelegate.swift`, not sure if these mentions should be updated as well.

0xnm (approved):

CONTRIBUTING.md
fuzzybinary: I may mention you only really need to run `./prepare.sh` once, but you might re-run `./generate_env.sh` to point at different environments.

And actually, I think there's an oversight where `./prepare.sh` doesn't prepare the `examples` directory, which I will fix.

PR #296, opened 2022-12-16
✨ Add W3C `traceparent` header injection

fuzzybinary:
### What and why?

Support W3C `traceparent` header injection and extraction.

Add truncating > 63-bit trace and span ids to 63-bit decimal when they forwarded over to Datadog.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -14,10 +14,13 @@ enum TracingHeaderType {
   dd,
```
maciejburda, line 14: it was suggested by Open Telemetry folks to rename this one to `datadog`

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -37,29 +40,38 @@ class OTelHttpTracingHeaders {
...
   final BigInt value;
 
   const TracingUUID(this.value);
 
-  static TracingUUID? fromString(
+  static TracingUUID fromString(
       String? uuid, TraceIdRepresentation representation) {
     if (uuid == null) {
-      return null;
+      return TracingUUID(BigInt.zero);
```
maciejburda, line 69: I'd say null makes more sense for invalid data.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -75,13 +87,16 @@ class TracingUUID {
...
 
-    return null;
+    return TracingUUID(BigInt.zero);
   }
 
   String asString(TraceIdRepresentation representation) {
     switch (representation) {
       case TraceIdRepresentation.decimal:
         return value.toString();
+      case TraceIdRepresentation.decimal63bit:
```
maciejburda, line 97: I wonder if it makes sense to have `decimal_X_bit` representation. For header values I think we're want:
decimal, hexadecimal16Chars, hexadecimal32Chars. The fact that it's an integer from 1 to 2^63 is UUID generation implementation detail.

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
```diff
@@ -288,6 +296,48 @@ void main() {
...
     });
 
+    test('truncates b3m headers and sets attributes', () async {
+      when(() => mockDatadog.isFirstPartyHost(any())).thenReturn(true);
+
+      final interceptor = DatadogGrpcInterceptor(mockDatadog, channel);
+
+      final stub = GreeterClient(channel, interceptors: [interceptor]);
+
+      // 3c9bdc4fffcb6efe8f19eead81ad506b truncated will be 0f19eead81ad506b or
```
maciejburda, line 306: IMO we shouldn't be truncating values in SDKs, they should be fully processed.

maciejburda (changes requested):
Great effort! 🙌

I left some suggestions around naming and treating values properly.

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
fuzzybinary: Ah, I must have misread the spec. I thought we were expected to truncate.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
fuzzybinary: This was because I thought we should be tuncating. I'll remove it.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -14,10 +14,13 @@ enum TracingHeaderType {
   dd,
 
   /// Open Telemetry B3 [Single header](https://github.com/openzipkin/b3-propagation#single-headers).
-  b3s,
+  b3,
 
   /// Open Telemetry B3 [Multiple headers](https://github.com/openzipkin/b3-propagation#multiple-headers).
   b3m,
```
plousada, line 20: Should be `b3multi` according to OT team.

maciejburda (approved):
LGTM

Just curious - shouldn't we keep unit tests for b3?

fuzzybinary:
Yes, I only removed the 'extraction' tests (the ones that pull from user set headers). Injection is handled in the `for` loop for every type of header.

PR #301, opened 2023-01-10
✨ Support tracing header types per-host.

fuzzybinary:
### What and why?

Instead of setting header types on each client, now we set them as part of `firstPartyHosts` during configuration.  This allows users to set different header types per host.

One oddity is when there are multiple matches, the code merges the two header requests and adds both. For example if you set:

```dart
  firstPartyHostsWithTracingHeaders: {
    'host.com': { TracingHeaderType.datadog },
    'api.host.com': {TracingHeaderType.b3 },
  },
```

Calls to `api.host.com` will have both Datadog and b3 headers injected .

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


maciejburda (approved):
LGTM - very similar to how it was done on iOS

🎉🎯

Issue #297, opened 2022-12-16
Cancel an in-progress user action


christinachimi:
**Is your feature request related to a problem? Please describe.**
I wanted to record taps on the `BottomNavigationBar`. This is difficult to address currently; the DD SDK does pick up on the underlying `GestureDetector` but of course the name is "unknown". I'd like to handle this by just writing my own user events when the tap occurs, but this fails if the BottomNavigationBar is under the `RumUserActionDetector`, because the SDK is already in progress on the `GestureDetector` tap event. My current solution is to not put the `BottomNavigationBar` in the `RumUserActionDetector`; but this necessitates having multiple `RumUserActionDetector` to handle detection across the app, as opposed to just wrapping the `MaterialApp`.

**Describe the solution you'd like**
Ideally the SDK would be able to pick up on BottomNavigationBar taps, but I suspect this may be difficult, since the actual tap is known about by some other stateful widget above the nav. A solution that I think could be useful would be the ability to cancel the in-progress event so that I can write my own (or, perhaps, to opt a specific sub-portion of the tree out of automatic detection).

**Describe alternatives you've considered**
I looked into adding `BottomNavigationBar` as a supported type in the `RumUserActionDetector`, but the widget itself knows about its current state, not the tap. I also tried wrapping a few relevant widgets in `RumUserActionAnnotation` but so far can only get the current index, not the tapped index.

fuzzybinary:
Hi @christinachimi 

We are looking into supporting the ability to map or discard specific RUM events, which would allow you to at least discard the ones being automatically generated by the RumUserActionDetector that don't fit. Hopefully that support will be coming soon.

That said, it also sounds like RumUserActionDetector needs some improvement pertaining to how it works with BottomNavigationBar and other GestureDetectors. I'll look into how we can better support these.

Thanks for reporting!

fuzzybinary:
Canceling in progress use actions is now in as part of the release of 1.2.0. You can use the `rumActionEventMapper` to discard user actions you do not want sent to Datadog.

I'm going to leave this open to track improvements to `RumUserActionDetector` around `BottomNavigationBar`.

christinachimi:
Great news, thank you!

fuzzybinary:
Enhancements to RumUserActionDetector that support BottomNavigationBar are now live in 1.3.2

Issue #175, opened 2022-08-02
Cancelled (Dio) HTTP requests show as errors in RUM


scopendo:
**Describe what happened**

My application uses the `Dio` library to make HTTP requests.  This allows for requests to be cancelled, which is useful when performing a typeahead search to cancel a previously dispatched API call when the search input changes.

It appears that when a Dio-dispatched request is cancelled, that `datadog_tracking_http_client` records an error in RUM, which adds unwanted noise.

**Steps to reproduce the issue:**

Here is an example using Riverpod, whereby a search for staff members is dispatched when search input changes, with the request automatically being cancelled when the provider is no longer watched, which happens when the search input changes.

```dart
  ...
  @override
  Widget build(BuildContext context, WidgetRef ref) {
    ...
    final searchProvider = staffMemberSearchFamily(searchText.value);
    final matches = ref.watch(searchPod);
    ...
  }
  ...
```

```dart
final staffMemberSearchFamily = FutureProvider.family.autoDispose<List<PersonData>, String?>(
  (ref, searchText) async {
    // Dispatch a cancellable HTTP request, meaning that we instruct the HTTP
    // client to ignore the response.
    final cancelToken = CancelToken();
    ref.onDispose(cancelToken.cancel);

    return ref.watch(personApiPod).internalPersonSearch(
          searchText: searchText,
          cancelToken: cancelToken,
        );
  },
);
```

**Describe what you expected:**

I expected that programmatically-cancelled HTTP requests would not appear in RUM as errors.

**Additional context**

 - Dart/Flutter version: 2.17.6/3.0.5
 - Datadog SDK version: 1.0.0-beta.3
 - datadog_tracking_http_client version: 1.0.1-beta.1
 - Versions of any other relevant dependencies: Dio 4.0.6

fuzzybinary:
Hi @scopendo! 

Thanks for reporting this. I'll start looking into this. It looks like Dio calls `HttpClient.close(force: true)`, when you cancel, which likely throws an error on the Request or Response object that we'll have to handle.

Are the errors always the same? Could you send me over what errors you're seeing?

scopendo:
Sorry, I meant to add that before.  The vast majority of errors are the same:

> HttpException: HttpException: Connection closed before full header was received, uri = https://...

From the `error` attributes:

```
isCrash: false
type: HttpException
```

fuzzybinary:
Thanks @scopendo!

[...]

PR #305, opened 2023-01-13
✨ RUM Event Mapping / Scrubbing

### LINKED ISSUE #303: Support anonymization data before send
Are you requesting automatic instrumentation for a framework or library? Please describe.

Framework or library name : datadog_flutter_plugin
Library type: monitoring
Library version: 1.1.1
Is your feature request related to a problem? Please describe.

Hi, 

As part of a professional project dealing with sensitive data, I would like to be able to anonymize personal data before sending it to the Datadog platform please.

This function is already available for Javascript. 

documentation : https://docs.datadoghq.com/real_user_monitoring/browser/modifying_data_and_context/?tab=npm

Thanks

fuzzybinary:
### What and why?

This allows Flutter / Dart code to modify or drop certain events before they get sent to Datadog, similar to the log event mappers.

There are 4 mapper types:
  * View Event Mappers can be used to modify view events, but not discard them
  * Action Event Mappers can modify action events, changing the names of targets or discarding events before they're sent to Datadog.
  * Resource Event Mappers can modify or scrub information from network requests (especially from URLs) or can be used to discard tracking certain resources.
  * Error Event Mappers can modify error events to change error messages, add or remove stack traces or discard events. These can also be used to discard errors attached to specific resources by looking at the `error.resource` member

This should fix #303, #297 and #175 .

### How?

The process for sending these events to Dart involves serializing the events to a JSON equivalent (Dictionary or Map), sending these as a callback through Flutter Method Channels, then using `json_serializable` to deserialize them before issuing type safe callbacks to Dart mapper functions.

On the way back, the data is again serialized to a JSON equivalent (Map), and send back through the method channel where modifiable attributes are changed on the underlying events.

This allows us to ensure that the data users are seeing in the mappers is as close as possible to the actual events sent by the Native SDKs, with all extra information intact.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
@@ -5,16 +5,23 @@
  */
 package com.datadoghq.flutter
 
+import android.os.Handler
+import android.os.Looper
+import com.datadog.android.Datadog
+import com.datadog.android.log.model.LogEvent
 import com.datadog.android.rum.GlobalRum
 import com.datadog.android.rum.RumActionType
 import com.datadog.android.rum.RumErrorSource
 import com.datadog.android.rum.RumMonitor
 import com.datadog.android.rum.RumPerformanceMetric
 import com.datadog.android.rum.RumResourceKind
+import com.datadog.android.rum.model.ViewEvent
 import io.flutter.embedding.engine.plugins.FlutterPlugin
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
+import io.flutter.plugin.common.MethodChannel.Result
 import java.lang.ClassCastException
+import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 
 class DatadogRumPlugin(
@@ -273,6 +280,78 @@ class DatadogRumPlugin(
                 RumPerformanceMetric.FLUTTER_RASTER_TIME,
                 it
             )
+        }
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    internal fun mapViewEvent(event: ViewEvent): ViewEvent {
+        val jsonEvent = event.toJson().asMap()
+
+        var modifiedJson: Map<String, Any?> = jsonEvent
+        val latch = CountDownLatch(1)
+
+        val handler = Handler(Looper.getMainLooper())
+        handler.post {
+            try {
+                channel.invokeMethod(
+                    "mapViewEvent",
+                    mapOf(
+                        "event" to jsonEvent
+                    ),
+                    object : Result {
+                        override fun success(result: Any?) {
+                            modifiedJson = (result as? Map<String, Any?>) ?: jsonEvent
+                            latch.countDown()
+                        }
+
+                        override fun error(
+                            errorCode: String,
+                            errorMessage: String?,
+                            errorDetails: Any?
+                        ) {
+                            latch.countDown()
+                        }
+
+                        override fun notImplemented() {
+                            Datadog._internal._telemetry.error(
+                                "mapLogEvent returned notImplemented."
+                            )
+                            latch.countDown()
+                        }
+                    }
+                )
+            } catch (e: Exception) {
+                Datadog._internal._telemetry.error("Attempting call mapLogEvent failed.", e)
+                latch.countDown()
+            }
+        }
+
+        try {
+            // Stalls until the method channel finishes
+            if (!latch.await(1, TimeUnit.SECONDS)) {
+                Datadog._internal._telemetry.debug("logMapper timed out")
+                return event
+            }
+
+            if (modifiedJson.containsKey("_dd.mapper_error")) {
+                return event
+            }
+
+            // Pull out modifyable elements
+            (modifiedJson["view"] as? Map<String, Any?>)?.let {
+                event.view.name = it["name"] as? String
+                event.view.referrer = it["referrer"] as? String
+                event.view.url = it["url"] as String
+            }
+
+            return event
+        } catch (e: Exception) {
+            Datadog._internal._telemetry.error(
+                "Attempt to deserialize mapped log event failed, or latch await was interrupted." +
+                        " Returning unmodified event.",
+                e
+            )
+            return event
         }
     }
 }
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
@@ -356,13 +356,13 @@ class DatadogRumPlugin(
                 e
             )
         }
-        
+
         return event
     }
 
     internal fun mapViewEvent(event: ViewEvent): ViewEvent {
         var jsonEvent = event.toJson().asMap()
-        jsonEvent = extractExtraUserInfo(jsonEvent)
+        jsonEvent = normalizeExtraUserInfo(jsonEvent)
 
         return callEventMapper("mapViewEvent", event, jsonEvent) { encodedResult, event ->
             (encodedResult?.get("view") as? Map<String, Any?>)?.let {
@@ -377,7 +377,7 @@ class DatadogRumPlugin(
 
     internal fun mapActionEvent(event: ActionEvent): ActionEvent? {
         var jsonEvent = event.toJson().asMap()
-        jsonEvent = extractExtraUserInfo(jsonEvent)
+        jsonEvent = normalizeExtraUserInfo(jsonEvent)
 
         return callEventMapper("mapActionEvent", event, jsonEvent) { encodedResult, event ->
             if (encodedResult == null) {
@@ -403,7 +403,7 @@ class DatadogRumPlugin(
 
     internal fun mapResourceEvent(event: ResourceEvent): ResourceEvent? {
         var jsonEvent = event.toJson().asMap()
-        jsonEvent = extractExtraUserInfo(jsonEvent)
+        jsonEvent = normalizeExtraUserInfo(jsonEvent)
 
         return callEventMapper("mapResourceEvent", event, jsonEvent) { encodedResult, event ->
             if (encodedResult == null) {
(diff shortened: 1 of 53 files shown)


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -275,6 +286,217 @@ class DatadogRumPlugin(
...
+            if (modifiedJson?.containsKey("_dd.mapper_error") == true) {
+                return event
+            }
+
+            return completion(modifiedJson, event)
+        } catch (e: Exception) {
+            Datadog._internal._telemetry.error(
+                "Attempt to deserialize mapped log event failed, or latch await was interrupted." +
+                    " Returning unmodified event.",
+                e
```
xgouchet, line 352: Here it might be interesting to distinguish both cases to have a more accurate message. I'm not sure what the best practice is in dart, but I guess using two `catch` blocks, one for the "deserialzation failed" and one for the "latch await interrupted". 
Especially since the deserialization indicates a SDK issue and could be sent to telemetry, whereas latch await would mean a usage issue.

xgouchet (approved):
Left a comment but feel free to disregard, otherwise LGTM

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -275,6 +286,222 @@ class DatadogRumPlugin(
...
+            )
+        }
+        
+        return event
+    }
+
+    internal fun mapViewEvent(event: ViewEvent): ViewEvent {
+        var jsonEvent = event.toJson().asMap()
+        jsonEvent = extractExtraUserInfo(jsonEvent)
```
0xnm, line 365: I think using `var` and then re-assigning it to the result of `extractExtraUserInfo` method can be misleading (I was thinking `extractUserInfo` method returns user info by looking on the name, but turns out it moves data inside the event).

Maybe something like `val jsonEvent = normalize(event.toJson().asMap())`?. `normalize` is not a good name obviously, we need to come up with something better, but this is the idea.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -275,6 +286,222 @@ class DatadogRumPlugin(
...
+                    event.view.url = it["url"] as String
+                }
+
+                event
+            }
+        }
+    }
+
+    private fun extractExtraUserInfo(encodedEvent: Map<String, Any?>): Map<String, Any?> {
+        val reservedKeys = listOf("email", "id", "name")
```
0xnm, line 491: minor: doesn't really matter for such small collection, but for lookup it is better to use `setOf`

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -275,6 +286,222 @@ class DatadogRumPlugin(
...
+        // Pull out user information
+        val mutableEvent = encodedEvent.toMutableMap()
+        (mutableEvent["usr"] as? MutableMap<String, Any?>)?.let { usr ->
```
0xnm, line 494: what guarantees that `mutableEvent["usr"]` is a `MutableMap`? is this the type assigned by us somewhere or by the Flutter itself?

packages/datadog_flutter_plugin/lib/src/rum/ddrum_events.dart
```diff
@@ -0,0 +1,697 @@
...
+part 'ddrum_events.g.dart';
+
+// Currently excluded
+//  * ciTest
+//  * display
+//  * featureFlags
+//  * source
+//  * synthetics
+@commonJsonOptions
+class RumViewEvent {
```
0xnm, line 19: side question: what if we write a generator for these classes from the schema, the same way we do for iOS/Android. Do you think it is worth it? Probably it will be a little of added value, because most of the properties added are read-only, but I'm wondering about your opinion.

packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
```diff
@@ -157,3 +173,140 @@ class DdRumMethodChannel extends DdRumPlatform {
...
+      actionEventMapper,
+      (e) => e.toJson(),
+      RumActionEvent.fromJson,
+    );
+  }
+
+  Map<Object, Object?>? _mapResourceEvent(MethodCall call) {
+    final eventJson = call.arguments['event'] as Map;
+    return _callMapper<RumResourceEvent>(
+      'mapActionEvent',
```
0xnm, line 283: typo, wrong mapper is called

```suggestion
      'mapResourceEvent',
```

0xnm (commented):
looks good! I've left some comments/suggestions

packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
fuzzybinary: :+1: nice catch. Thankfully it's just for error messages.

packages/datadog_flutter_plugin/lib/src/rum/ddrum_events.dart
fuzzybinary: I thought about it, but I'm concerned about the level of effort versus the payoff. If I have to add a lot more events in the future I'll look into the generator.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
fuzzybinary: Yeah, it's a good call, I had similar thoughts. `normalizeUserInfo` actually works as a name though. I'll change it.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
fuzzybinary: Hrm, nothing guarantees it. `JsonObject.asMap` happens to return mutable maps but that doesn't mean it couldn't change. I'll cast to map and explicitly call `toMutableMap` instead.

0xnm (approved):

packages/datadog_flutter_plugin/ios/Classes/DatadogRumPlugin.swift
```diff
@@ -260,6 +264,258 @@ public class DatadogRumPlugin: NSObject, FlutterPlugin {
...
+                encodedResult = nil
+            } else if let result = result as? [String: Any] {
+                encodedResult = result
+            }
+
+            semaphore.signal()
+        }
+
+        if semaphore.wait(timeout: .now() + DispatchTimeInterval.milliseconds(250)) == .timedOut {
+            Datadog._internal.telemetry.debug(id: "event_mapper_timeout", message: "\(mapperName) timed out.")
```
ncreated, line 287: question/ Do you foresee running into time outs often? If so, is there more context we could add to understand it better?

ncreated (approved):
Looks good iOS-side 👍. I left few suggestions / questions - no blockers. Well done 🏅.

packages/datadog_flutter_plugin/ios/Classes/DatadogRumPlugin.swift
```diff
@@ -260,6 +264,258 @@ public class DatadogRumPlugin: NSObject, FlutterPlugin {
...
+            guard let encodedResult = encodedResult else {
+                return rumViewEvent
+            }
+
+            // Pull modifyable properties
+            var rumViewEvent = rumViewEvent
+            if let encodedView = encodedResult["view"] as? [String: Any?] {
+                rumViewEvent.view.name = encodedView["name"] as? String
+                rumViewEvent.view.referrer = encodedView["referrer"] as? String
+                rumViewEvent.view.url = encodedView["url"] as! String
```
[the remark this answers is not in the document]
fuzzybinary: Yeah, agreed. I'll fix.

packages/datadog_flutter_plugin/ios/Classes/DatadogRumPlugin.swift
fuzzybinary: I'm hoping we won't see it very often.  My own timing suggests that this should take very little time.

The one time we might see it would be if someone is stalling the main Dart thread. There's an improvement I'd like to make to the mappers to run in their own lightweight isolate, but that will require some extra work down the road.  This will tell me if I need to prioritize that.

packages/datadog_flutter_plugin/ios/Classes/DatadogRumPlugin.swift
```diff
@@ -260,6 +264,258 @@ public class DatadogRumPlugin: NSObject, FlutterPlugin {
...
+
+    func viewEventMapper(rumViewEvent: RUMViewEvent) -> RUMViewEvent {
+        let encoder = DictionaryEncoder()
+        guard var encoded = try? encoder.encode(rumViewEvent) else {
+            Datadog._internal.telemetry.error(
+                id: "datadog_flutter:view_mapping_encoding",
+                message: "Encoding a RUMViewEvent failed",
+                kind: "EncodingError",
+                stack: nil
+            )
```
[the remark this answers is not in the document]
fuzzybinary: I don't think it will be that common in Flutter, but probably something we want to look into adding anyway.

packages/datadog_flutter_plugin/ios/Classes/DatadogRumPlugin.swift
fuzzybinary: Did a bit more timing that suggests I might want to increase this for debug builds to 500ms.

ncreated: > This will tell me if I need to prioritize that.

Sounds good 👍

PR #351, opened 2023-02-28
✨ Add the ability to forward events from web views to mobile sessions

fuzzybinary:
### What and why?

Customers want the ability to track the full journey of a use, which may include interacting with WebViews which are already tracked by Datadog. This allows them to forward events from their WebViews to Mobile sessions.

This requires the use of the `webview_flutter` package, which is the main recommended package for Flutter.

### How?

We're utilizing the bridging interfaces Datadog has already written for WKWebView and Android's WebView. By utilizing an interface provided by `webview_flutter`, we're able to get the underlying platform controls and attach Datadog's observers.

Because `webview_flutter` requires Flutter 3.0, and we're currently committed to maintaining compatibility with Flutter 2.8, we're putting this into its own package.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


kayayarai:
opened https://datadoghq.atlassian.net/browse/DOCS-4980 to track docs review

packages/datadog_flutter_plugin/README.md
```diff
@@ -23,7 +23,7 @@ Your iOS Podfile must have `use_frameworks!` (which is true by default in Flutte
 
 ### Android
 
-On Android, your `minSdkVersion` must be >= 19, and if you are using Kotlin, it should be version >= 1.5.31.
+On Android, your `minSdkVersion` must be >= 19, and if you are using Kotlin, it should be version >= 1.6.21.
```
0xnm, line 26: side comment: Strictly speaking it should still be ok to have 1.4 or 1.5, because I think Kotlin should be able to keep binary compatibility for existing APIs 2 versions back, and during the dependency resolution Gradle will anyway update other Kotlin transitive dependencies to the most recent from the overall set of versions requested.

But let's keep it like that.

packages/datadog_flutter_plugin/doc/common_setup.md
```diff
@@ -26,7 +26,7 @@ You can replace `11.0` with any minimum version of iOS you want to support that
 
 ### Android
 
-For Android, your `minSdkVersion` version must be >= 19, and if you are using Kotlin, it should be a version >= 1.5.31. These constraints are usually in the head in your `app/build.gradle` file.
+For Android, your `minSdkVersion` version must be >= 19, and if you are using Kotlin, it should be a version >= 1.6.21.. These constraints are usually held in your `android/app/build.gradle` file.
```
0xnm, line 29: typo: double dot.

```suggestion
For Android, your `minSdkVersion` version must be >= 19, and if you are using Kotlin, it should be a version >= 1.6.21. These constraints are usually held in your `android/app/build.gradle` file.
```

packages/datadog_webview_tracking/android/src/main/kotlin/com/datadoghq/flutter/webview/DatadogFlutterWebViewPlugin.kt
```diff
@@ -0,0 +1,91 @@
...
+                )
+            }
+        }
+    }
+}
+
+internal const val DATADOG_FLUTTER_WEBVIEW_TAG = "DatadogFlutterWebView"
+
+internal const val JAVA_SCRIPT_NOT_ENABLED_WARNING_MESSAGE =
+    "You are trying to enable Datadog WebView tracking but the java script capability was not" +
```
0xnm, line 89:
```suggestion
    "You are trying to enable Datadog WebView tracking but the JavaScript capability was not" +
```

xgouchet (approved):

packages/datadog_webview_tracking/example/android/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+    repositories {
+        google()
+        mavenCentral()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:7.2.0'
```
0xnm, line 12: should it be `7.4.1` here? like in other sample apps.

packages/datadog_webview_tracking/example/android/detekt.yml
```diff
@@ -0,0 +1,1107 @@
+build:
```
0xnm, line 1: we also have the same file in the `datadog_flutter_plugin/example`, so maybe we can have a single file for all the projects later to avoid duplication.

packages/datadog_webview_tracking/example/integration_test/webview_scenario_test.dart
```diff
@@ -0,0 +1,86 @@
...
+        startView.viewEvents.first.rumEvent['application']['id'];
+    String expectedSessionId =
+        startView.viewEvents.first.rumEvent['session']['id'];
+
+    final browserView = session.visits[1];
+
+    for (var event in browserView.viewEvents) {
+      expect(event.rumEvent['application']['id'], expectedApplicationId);
+      expect(event.rumEvent['session']['id'], expectedSessionId);
+      expect(event.service, 'shopist-web-ui');
```
0xnm, line 75: fyi: we also have a dedicated page for the webview integration testing here https://github.com/DataDog/dd-sdk-android/blob/c618644f2b172549f16124f6cba105678b83f453/instrumented/nightly-tests/src/main/kotlin/com/datadog/android/nightly/activities/WebViewTrackingActivity.kt#L30 if it is needed to test something more complex.

packages/datadog_webview_tracking/example/ios/Tests/DatadogWebviewTrackingPluginTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        plugin.handle(call, result: { result in
+            resultStatus = .called(value: result)
+        })
+
+        switch resultStatus {
+        case .called(value: let value):
+            let error = value as? FlutterError
+            XCTAssertNotNil(error)
+            XCTAssertEqual(error?.code, "DatadogSdk:ContractViolation",
+                           "initwebView did not throw a contact violation when missing parameter 'allowedHosts'")
```
0xnm, line 55:
```suggestion
                           "initWebView did not throw a contact violation when missing parameter 'allowedHosts'")
```

packages/datadog_webview_tracking/example/ios/Tests/DatadogWebviewTrackingPluginTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        plugin.handle(call, result: { result in
+            resultStatus = .called(value: result)
+        })
+
+        switch resultStatus {
+        case .called(value: let value):
+            let error = value as? FlutterError
+            XCTAssertNotNil(error)
+            XCTAssertEqual(error?.code, "DatadogSdk:ContractViolation",
+                           "initwebView did not throw a contact violation when missing parameter 'webViewIdentifier'")
```
0xnm, line 32:
```suggestion
                           "initWebView did not throw a contact violation when missing parameter 'webViewIdentifier'")
```

0xnm (approved):
nice, lgtm! I've added few minor suggestions, but they are not blocking for me.

packages/datadog_webview_tracking/example/android/build.gradle
fuzzybinary: Yes probably. Thanks!

packages/datadog_webview_tracking/example/android/detekt.yml
fuzzybinary: Unfortunately each package has to "stand alone" because only the files in the package directory get deployed during `pub publish`. We're having some issues where if we reference files in gradle (or other tools) that are outside the package directory it causes build failures.

So... a little bit of duplication is necessary.

PR #368, opened 2023-03-17
🐛 Fix b3 and tracecontext header generation

fuzzybinary:
### What and why?

TraceId and SpanId were swapped in both b3 and tracecontext header generation.

Added more unit test coverage for these functions as well.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -169,8 +169,8 @@ Map<String, String> getTracingHeaders(
     case TracingHeaderType.b3:
       if (context.sampled) {
         final headerValue = [
-          context.spanId.asString(TraceIdRepresentation.hex32Chars),
-          context.traceId.asString(TraceIdRepresentation.hex16Chars),
+          context.traceId.asString(TraceIdRepresentation.hex32Chars),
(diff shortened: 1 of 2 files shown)


packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -169,8 +169,8 @@ Map<String, String> getTracingHeaders(
     case TracingHeaderType.b3:
       if (context.sampled) {
         final headerValue = [
-          context.spanId.asString(TraceIdRepresentation.hex32Chars),
-          context.traceId.asString(TraceIdRepresentation.hex16Chars),
+          context.traceId.asString(TraceIdRepresentation.hex32Chars),
+          context.spanId.asString(TraceIdRepresentation.hex16Chars),
```
0xnm, line 173: just curious: what is the reason of having 32 vs 16 chars?

fuzzybinary: 64 vs 128 bits. We only generate 63-bit ids but both b3 and traceheader trace Ids should be 128 bits (spans are still 64). We pad the beginning with 0's to bring it up to 128 bits.

xgouchet (approved):

PR #412, opened 2023-05-02
🐛 Fix rethrown exceptions not properly preserving stacktrace

fuzzybinary:
### What and why?

Dart has "rethrow" within catch blocks to preserve stack traces, and Futures will preserve stack traces if you "throw" the same exception so long as it is not inside an `async` block. The `close` method had its `onError` marked as async by accident, so it was not preserving the stack trace properly.

Fixes RUMM-3264

### Review checklist

- [x] Run unit tests
- [x] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/helpers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/helpers.dart
@@ -11,11 +11,13 @@ import 'internal_logger.dart';
 
 typedef WrappedCall<T> = FutureOr<T?> Function();
 
-void _handleError(Object error, StackTrace stackTrace, String methodName,
+// Returns true if the error was handled, false if the error should be re-thrown
+bool _handleError(Object error, StackTrace stackTrace, String methodName,
     InternalLogger logger, Map<String, Object?>? serializedAttributes) {
   if (error is ArgumentError) {
     logger.warn(InternalLogger.argumentWarning(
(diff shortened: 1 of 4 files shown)


xgouchet (approved):

packages/datadog_flutter_plugin/lib/src/helpers.dart
```diff
@@ -40,11 +43,16 @@ void wrap(
   try {
     var result = call();
     if (result is Future) {
-      result.catchError((dynamic e, StackTrace st) =>
-          _handleError(e, st, methodName, logger, attributes));
+      result.catchError((dynamic e, StackTrace st) {
+        if (!_handleError(e, st, methodName, logger, attributes)) {
+          throw e;
```
0xnm, line 48: should it be still `throw e` here and not `rethrow`? All the cases below with `_handleError` calls are using `rethrow`.

fuzzybinary: Okay updated. I thought `catchError` worked like `then`, which will preserve the stacktrace if you throw synchronously. It turns out, no, you should supply a `test` function to `catchError` that returns true if you will handle the exception, and passes it on to the Future otherwise.

Thanks for pointing this out!

0xnm (approved):

Issue #442, opened 2023-06-12
Expose a way to get the Session Id


almovsesanso:
It looks like there'a way to get the session id through the web SDK but I couldn't find a similar way to get it from Flutter. I need it for the same purpose as the web app, to allow the user to contact us when they're encountering a problem.

https://github.com/DataDog/browser-sdk/issues/390

fuzzybinary:
Hi @almovsesanso thanks for the suggestion. We're looking into it!

almovsesanso:
> Hi @almovsesanso thanks for the suggestion. We're looking into it!

Thanks @fuzzybinary. Do you know if there's an indirect way to get the session id currently using flutter?

fuzzybinary:
It's not a great workaround, but you can use a `RumViewEventMapper` and inspect `RUMViewEvent.session.id` and get the session after we send the first batch to Datadog.

utkarshagarwal:
@fuzzybinary as suggested we used view event mapper to extract the session id. It works perfectly for our use case for iOS simulator and android (devices and simulator both) but fails on iOS real device. Any particular reason you can think of?
The event mapper is not getting called at all on device.

fuzzybinary:
Not that I can think of, no, but I will look into it. I was able to get the callback to fire in a debug build on device, but I will try a release build on device later today.

Is there anything else strange about your setup? How are you checking that the mapper is firing?

fuzzybinary:
@utkarshagarwal I just tested our example application on an iPhone 11 and didn't have any issues getting a mapper callback. Note that the mappers are called immediately before we send data to Datadog, and there are situations (like low battery) where the SDK prefers to wait to send data, which is why this approach to getting session id isn't ideal....

almovsesanso:
> @utkarshagarwal I just tested our example application on an iPhone 11 and didn't have any issues getting a mapper callback. Note that the mappers are called immediately before we send data to Datadog, and there are situations (like low battery) where the SDK prefers to wait to send data, which is why this approach to getting session id isn't ideal....

Thanks for replying, Jeff.

Do you know if there's a way to one-off force the SDK to send events?

fuzzybinary:
Unfortunately no, there's no public `flush` or similar.

We are looking to add an official session id accessor very soon though. Sorry for the delay!

MauriMiguez:
Hi! Any update on this Issue? Do you have a timeline for this?

fuzzybinary:
[...]

PR #444, opened 2023-06-16
✨ Add sessionId getter and sessionStarted callback

fuzzybinary:
### What and why?

sessionStarted does not work on Flutter web, but the session gets updated properly in the UI of the integration test app on re-build of the overlay.

refs: RUMM-3345, #442

### How?

For mobile, the platform channel memoizes the last seen session id and returns it. In cases where the session is sampled out, sessionId returns an empty string.

### Review checklist

- [x] Run unit tests
- [x] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/CHANGELOG.md
+++ b/packages/datadog_flutter_plugin/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-
+* Add a `sessionId` getter to DdRum, as well as a `sessionStarted` callback. See [#442]
 
 ## 1.5.1
 
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
@@ -15,6 +15,7 @@ import com.datadog.android.rum.RumErrorSource
 import com.datadog.android.rum.RumMonitor
 import com.datadog.android.rum.RumPerformanceMetric
 import com.datadog.android.rum.RumResourceKind
+import com.datadog.android.rum.RumSessionListener
 import com.datadog.android.rum.model.ActionEvent
 import com.datadog.android.rum.model.ErrorEvent
 import com.datadog.android.rum.model.LongTaskEvent
@@ -84,6 +85,11 @@ class DatadogRumPlugin(
     ) {
         rum = RumMonitor.Builder()
             .sampleRumSessions(configuration.sampleRate)
+            .setSessionListener(object : RumSessionListener {
+                override fun onSessionStarted(sessionId: String, isDiscarded: Boolean) {
+                    sessionStarted(sessionId, isDiscarded)
+                }
+            })
             .build()
         GlobalRum.registerIfAbsent(rum!!)
     }
@@ -339,6 +345,19 @@ class DatadogRumPlugin(
     private fun stopSession(call: MethodCall, result: Result) {
         rum?.stopSession()
         result.success(null)
+    }
+
+    private fun sessionStarted(sessionId: String, isDiscarded: Boolean) {
+        val handler = Handler(Looper.getMainLooper())
+        handler.post {
+            channel.invokeMethod(
+                "rumSessionStarted",
+                mapOf(
+                    "sessionId" to sessionId,
+                    "sampled" to isDiscarded,
+                )
+            )
+        }
     }
 
     @Suppress("TooGenericExceptionCaught")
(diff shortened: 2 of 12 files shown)


xgouchet (approved):

packages/datadog_flutter_plugin/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-
+* Add a `sessionId` getter to DdRum, as well as a `sessionStarted` callback. See [#442]
```
0xnm, line 5:
```suggestion
* Add a `sessionId` getter to `DdRum`, as well as a `sessionStarted` callback. See [#442]
```

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -341,6 +347,19 @@ class DatadogRumPlugin(
...
     }
 
+    private fun sessionStarted(sessionId: String, isDiscarded: Boolean) {
+        val handler = Handler(Looper.getMainLooper())
+        handler.post {
+            channel.invokeMethod(
+                "rumSessionStarted",
+                mapOf(
+                    "sessionId" to sessionId,
+                    "sampled" to isDiscarded,
```
0xnm, line 357: shouldn't it be `!isDiscared` here? because if `isDiscarded = true`, it means that session is sampled out, meaning that `sampled` should be `false`.

packages/datadog_flutter_plugin/lib/src/rum/ddrum.dart
```diff
@@ -96,6 +96,28 @@ class DdRum {
 
   RumLongTaskObserver? _longTaskObserver;
 
+  /// Get the last active session id, or an empty string if a session hasn't started
+  ///
+  /// Session ids are updated when a new session is started in Datadog, usually
+  /// as a result of a user action. If you are using [stopSession], [sessionId] will
+  /// continue to return the last active session until a new session starts.
```
0xnm, line 103: Shouldn't we return `null` instead of there is no active session? what is the reasoning behind returning the last active session when it was stopped? What Browser SDK is doing in this case?

xgouchet (approved):

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
fuzzybinary: Yeah, I'm going to revisit the name of this variable. `sampled` here is supposed to mean "sampled out" which means `isDiscarded = true` but considering I actually messed it up twice while unit testing, I should have realized it's a bad name.

packages/datadog_flutter_plugin/ios/Classes/DatadogRumPlugin.swift
```diff
@@ -375,6 +375,18 @@ public class DatadogRumPlugin: NSObject, FlutterPlugin {
         return usrMember
     }
 
+    func rumSessionStarted(_ sessionId: String, _ sampled: Bool) {
```
0xnm, line 378: maybe it is worth to rename `sampled` here as well to be more descriptive? I would personally think its meaning is that session was sampled in.

packages/datadog_flutter_plugin/lib/src/rum/ddrum.dart
```diff
@@ -96,6 +96,28 @@ class DdRum {
...
+  String get sessionId => _platform.sessionId;
+
+  /// Get the current callback listening for `sessionStarted` events.
+  ///
+  /// This callback does not work for Flutter Web.
+  SessionStartedCallback? get sessionStarted => _platform.sessionStarted;
+
+  /// Set the method to be called when new sessions are started in Datadog RUM.
+  ///
+  /// If a session is sampled out, [sessionId] will be an empty string.
```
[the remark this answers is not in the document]
0xnm: I mean above there is the following statement:

```
If you are using [stopSession], [sessionId] will
  /// continue to return the last active session until a new session starts. Untracked
  /// sessions will have [DdRumSessionInfo.isSampledOut] set to `true`
  ///
  /// On web, sessionId is the currently active session, or `null` if the current session
  /// has been sampled out or if there is no active session.
```

So for the consistency, if session was stopped, shouldn't we return null or empty string instead? Not quite sure why we should return last active session when it is not active anymore.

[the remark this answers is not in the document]
fuzzybinary: You're not going to like the answer .... which is that I don't have any way to know that the session was stopped currently, unless I assume `stopSession` is always called from Flutter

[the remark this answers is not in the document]
fuzzybinary: Yeah, maybe we close this and delay it until we can either add the sessionId accessor to both native clients, or at least a callback on session stop as well.

fuzzybinary:
Closing for now in favor of changing some implementation details on the native side first.

PR #459, opened 2023-07-20
✨ Add log sample rate to loggers.

fuzzybinary:
### What and why?

There was a request to add `sampleRate` to Flutter loggers, similar to those that exist on Web and Native. Because of platform differences, Flutter handles log sampling itself, and keeps the sampling rate at the native level at 100%

This also differs slightly from Web, as each log is sampled, instead of each session.

refs: RUMM-3467

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/test/datadog_sdk_test.dart
+++ b/packages/datadog_flutter_plugin/test/datadog_sdk_test.dart
@@ -268,6 +268,28 @@ void main() {
         logger!.loggerHandle, loggingConfiguration));
   });
 
+  test('initialize with logging sample rate creates logger with sample rate',
+      () async {
+    when(() => mockLogsPlatform.createLogger(any(), any()))
+        .thenAnswer((_) => Future<void>.value());
+
+    final loggingConfiguration = LoggingConfiguration();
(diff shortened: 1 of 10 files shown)


packages/datadog_flutter_plugin/test/datadog_sdk_test.dart
```diff
@@ -268,6 +268,28 @@ void main() {
         logger!.loggerHandle, loggingConfiguration));
   });
 
+  test('initialize with logging sample rate creates logger with sample rate',
```
0xnm, line 271: It seems that this test is identical to the test above, I don't see any setup/assert for the sample rate here.

xgouchet (approved):

packages/datadog_flutter_plugin/test/datadog_sdk_test.dart
fuzzybinary: Ah, good call. I think this is left over from when I put the sample rate on the global configuration and it never got removed.

0xnm (approved):

plousada (approved):

PR #469, opened 2023-08-29
👩‍💻 Use melos for monorepo management

fuzzybinary:
### What and why?

Replacing some custom scripts and dependency overrides with [Melos](https://github.com/invertase/melos), which I'm hoping will simplify monorepo management in the future.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


bitrise.yml
```diff
@@ -39,6 +39,9 @@ workflows:
   _deploy_artifacts:
     steps:
     - deploy-to-bitrise-io: {}
+    - deploy-to-bitrise-io:
```
0xnm, line 42: same step twice?

0xnm (approved):

bitrise.yml
fuzzybinary: Different paths. The first is the default path that most of the steps use, and the second is where I'm putting the test output. I'll probably remove the default once I'm not using any bitrise steps.

Issue #451, opened 2023-07-04
Should be possible to use Dio interceptor to track request


Kiruel:
**Is your feature request related to a problem? Please describe.**
I would like to use Dio client to trace my request to datadog.

**Describe the solution you'd like**
I saw in the unofficial library this interceptor https://github.com/GetDutchie/datadog_flutter/issues/47#issuecomment-915049269 that can be fit my need.

**Describe alternatives you've considered**
I was trying using `startResourceLoading` and `stopResourceLoading` but with no result, the data is not send to datadog. And if I use `enableHttpTracking` he do not track the query title etc...

**Additional context**
I'm using Dio with graphql to perform some query etc... this is the code I use to configure the graphql:
```dart
@Riverpod(keepAlive: true)
GraphQLClient graphQlClient(GraphQlClientRef ref) => GraphQLClient(
      cache: GraphQLCache(
        store: kIsWeb
            ? InMemoryStore()
            : HiveStore(
                Hive.box<Map<dynamic, dynamic>?>(hiveBoxVersion),
              ),
      ),
      defaultPolicies: DefaultPolicies(
        mutate: Policies(fetch: FetchPolicy.networkOnly),
        query: Policies(fetch: FetchPolicy.networkOnly),
        subscribe: Policies(fetch: FetchPolicy.networkOnly),
        watchMutation: Policies(fetch: FetchPolicy.networkOnly),
        watchQuery: Policies(fetch: FetchPolicy.networkOnly),
      ),
      link: Link.from([
        ErrorLink(
          onGraphQLError: (request, forward, response) {
            if (response.errors?.first.extensions?['code'] == 'UNAUTHORIZED') {
              // Here capture errors from graphql
            }
            return;
          },
        ),
        DioLink(
          'GRAPHQL_API_URL',
          client: Dio(
            BaseOptions(
              baseUrl: 'BASE_URL',
              connectTimeout: const Duration(milliseconds: 5000),
              receiveTimeout: const Duration(milliseconds: 20000),
              contentType: Headers.jsonContentType,
            ),
          )
            ..interceptors.add(
              // here I would like to add interceptor to track my graphql query/mutation ...
            )
        ),
      ]),
    );
```

Kiruel:
Hi @fuzzybinary , thanks for quick reply. 

[...]

Issue #397, opened 2023-04-12
Customize RUM HttpTracking


ClaireDavis:
We are using RUM in our flutter application and have enabled http tracking using the `enableHttpTracking` method on the `DdSdkConfiguration`.  The problem is that our application also uses GraphQL and the automatic http tracking in the datadog framework records the resource.url_path which is the same for every GraphQL operation.  This makes the logs pretty meaningless.  I'd like to be able to include the operation name in logs but there doesn't appear to be a way to do so.  

I tried to use the `rumResourceEventMapper`, but when this is called, the event has no context on what the operation was.

Do you have any suggestions on how to achieve this within the current framework?  

current plugins +  versions:
  `datadog_flutter_plugin: ^1.3.2`
  `datadog_tracking_http_client: ^1.2.1`

fuzzybinary:
Hi @ClaireDavis ,

Unfortunately there's nothing you can do currently other than potentially tracking the resources manually with `DdRum.startResource` / `DdRum.stopResource` and adding attributes to those calls.

However, we're currently looking into ways to support GraphQL more directly. I'll update the issue when we have more information available.

ClaireDavis:
@fuzzybinary  thanks for the prompt reply!  

Any additional GraphQL support would be greatly appreciated - thanks for keeping us updated!  We'll look into using `startResource`/`stopResource` in the meantime

fuzzybinary:
HI @ClaireDavis,

A few quick questions for you:

First, what package(s) are you using for GraphQL support in Flutter? Are you using [graphql](https://pub.dev/packages/graphql)?

Second, we're hooking into the network libraries at a fairly low level, which makes it hard to return data to any listening callbacks. Would getting the HttpClientRequest / HttpClientResponse be enough to get the data you need, or would you need the body of the requests and responses as well?

ClaireDavis:
Hi @fuzzybinary - we are using [Ferry](https://ferrygraphql.com/), not graphql, but from my research, they operate in very similar ways and share a lot of the same underlying packages.

Our intermediate solution was to use an `Interceptor` and append the query name to the end of the `url`.  As long as we have access to the `Query` name, we should be fine, but I believe that is part of the `data` of the request

ClaireDavis:
Awesome! I'll check out what's available in 1.3.0.  I appreciate the support!

fuzzybinary:
[...]

PR #470, opened 2023-09-07
feat: Create a Datadog `gql_link` package

fuzzybinary:
### What and why?

This is the initial implementation of the `DatadogGqlLink` that supports distributed tracing and RUM2APM with GraphQL libraries that support `gql_link`. More testing in required on this link, including integration tests and checking it against an actual GraphQL endpoint.

This is incompatible with both versions of the tracking http client as they will interfere with one another.We will be adding an `ignoreUrls` parameter to `DatadogClient` and to `DatadogTrackingHttpClient` so that it can these trackers can be used together.

refs: RUM-967 #451 #397

### Review checklist

- [x] Run unit tests
- [x] Run integration tests


rtrieu (dismissed):
Looks good!

packages/datadog_gql_link/lib/src/datadog_gql_link.dart
```diff
@@ -0,0 +1,234 @@
...
+
+class _GraphQLAttributes {
+  static const operationType = 'operation_type';
+  static const operationName = 'operation_name';
+  static const errors = 'errors';
+}
+
+abstract interface class DatadogGqlListener {
+  void requestStarted(Request request, Map<String, Object?> attributes);
+  void responseRecieved(Response response, Map<String, Object?> attributes);
```
0xnm: small typo here

```suggestion
  void responseReceived(Response response, Map<String, Object?> attributes);
```

packages/datadog_gql_link/lib/src/datadog_gql_link.dart
```diff
@@ -0,0 +1,234 @@
...
+      ...userAttributes,
+      ...datadogAttributes,
+      '_dd': {
+        'graphql': {
+          ...internalAttributes,
+        }
+      },
+    };
+
+    // TODO: Assume `post` for now, but most links support `get` queries.
```
0xnm: it is better to add a ticket reference here, so that we don't forget

packages/datadog_gql_link/test/datadog_gql_link_test.dart
```diff
@@ -0,0 +1,839 @@
...
+      responseController.sink.add(response);
+      unawaited(responseController.sink.close());
+      await stream.drain();
+
+      // Then
+      verify(() => mockRum.stopResourceLoading(
+          capturedKey!, 418, RumResourceType.native, 66219, any()));
+    });
+
+    test('link calls listener on response recieved', () async {
```
0xnm:
```suggestion
    test('link calls listener on response received', () async {
```

0xnm (changes requested):
nice feature! I added few suggestions.

packages/datadog_gql_link/lib/src/datadog_gql_link.dart
fuzzybinary: Yes, good call. I'll add it.

0xnm (approved):

PR #474, opened 2023-09-15
feat!: Upgrade to Datadog iOS v2 

fuzzybinary:
### What and why?

Update the Datadog Flutter Plugin to use iOS v2.

As part of this, I'm trying to clean up some of the interfaces as well as move things into proper modules. I'm not sure we'll split the packages the way the native SDKs have, but this does make some of the internal logic cleaner.

This only updates the iOS and Flutter interfaces to v2. Android will be broken for now.

Still remaining as part of this conversion:
  * attachToExisting and tests
  * mappers (want to take another look at these for v2)
  * Tests and defaults for the new console logger.

refs: RUM-1035

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests


examples/native-hybrid-app/flutter_module/lib/main.dart
```diff
@@ -2,21 +2,19 @@
...
   WidgetsFlutterBinding.ensureInitialized();
 
-  final config = DdSdkExistingConfiguration(
-    loggingConfiguration: LoggingConfiguration(),
-    detectLongTasks: true,
-  )..enableHttpTracking();
+  // final config = DdSdkExistingConfiguration(
+  //   loggingConfiguration: LoggingConfiguration(),
+  //   detectLongTasks: true,
+  // )..enableHttpTracking();
```
maciejburda, line 15: do we want to bring it back?

packages/datadog_flutter_plugin/e2e_test_app/ios/Podfile
```diff
@@ -33,8 +33,10 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
   # Datadog Pod Overrides
-  pod 'DatadogSDK', :git => 'https://github.com/DataDog/dd-sdk-ios', :tag => '1.23.0'
-  pod 'DatadogSDKCrashReporting', :git => 'https://github.com/DataDog/dd-sdk-ios', :tag => '1.23.0'
+  pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios', :branch => 'develop'
```
maciejburda, line 36: We probably want to point to specific version `2.2.1`

packages/datadog_flutter_plugin/example/ios/Tests/FlutterSdkAttibutesTests.swift
```diff
@@ -27,7 +28,7 @@ class FlutterSdkAttributesTests: XCTestCase {
 
     func testAttributes_NestedTypes_AreEncodedProperly() {
         let flutterTypes: [String: Any?] = [
-            "arrayType": [ "My String Value", NSNumber(value: 32) ],
+            "arrayType": [ "My String Value", NSNumber(value: 32) ] as [Any],
```
maciejburda, line 31: is it needed?

packages/datadog_flutter_plugin/example/ios/Tests/FlutterSdkAttibutesTests.swift
```diff
@@ -3,8 +3,9 @@
 // Copyright 2019-2022 Datadog, Inc.
 
 import Foundation
+import Flutter
```
maciejburda, line 6: what does it change?

packages/datadog_flutter_plugin/example/lib/logging_screen.dart
```diff
@@ -20,6 +20,15 @@ class _LoggingScreenState extends State<LoggingScreen> {
   var _selectedLevel = LogLevel.debug;
   String? _message;
 
+  late DatadogLogger logger;
+
+  @override
+  void initState() {
+    logger =
+        DatadogSdk.instance.logs!.createLogger(DatadogLoggerConfiguration());
```
maciejburda, line 28: I've seen it some many times. It almost asks for some kind of `createDefaultDatadogLogger()` convenience function

packages/datadog_flutter_plugin/integration_test_app/ios/Podfile
```diff
@@ -33,8 +33,10 @@ target 'Runner' do
   flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
 
   # Datadog Pod Overrides
-  pod 'DatadogSDK', :git => 'https://github.com/DataDog/dd-sdk-ios', :tag => '1.23.0'
-  pod 'DatadogSDKCrashReporting', :git => 'https://github.com/DataDog/dd-sdk-ios', :tag => '1.23.0'
+  pod 'DatadogCore', :git => 'https://github.com/DataDog/dd-sdk-ios', :branch => 'develop'
```
maciejburda, line 36: same comment about the version

packages/datadog_flutter_plugin/lib/datadog_flutter_plugin.dart
```diff
@@ -265,8 +247,8 @@ class DatadogSdk {
     });
   }
 
-  /// Determine if the provided URI is a first party host as determined by the
-  /// value of [firstPartyHosts].
```
maciejburda, line 269: In Swift `///` allows introspection of a comment if you option click the property.

Not sure if it works this way in dart.

maciejburda (approved):
Looking good! I like how v2 comes into play. Really makes those APIs nicer to call.

Having only changes related to v2 migration isolated would make this PR slightly smaller, but not by much.

I'm wondering what would be the best approach with Android migration. If we merge this now we'll end up in a broken state on develop. Maybe worth making ios and android migration happen on a separate branch and merge it together to develop.

examples/native-hybrid-app/flutter_module/lib/main.dart
fuzzybinary: Yes. I've created RUM-1066 to track

packages/datadog_flutter_plugin/e2e_test_app/ios/Podfile
fuzzybinary: During development I track develop. I have a script that removes these overrides when we go to release.

packages/datadog_flutter_plugin/example/ios/Tests/FlutterSdkAttibutesTests.swift
fuzzybinary: Swift has started issuing warning for heterogeneous types. Or maybe swiftlint does. It's not strictly necessary but I don't love wrnings.

packages/datadog_flutter_plugin/example/ios/Tests/FlutterSdkAttibutesTests.swift
fuzzybinary: I moved from using and Objective-C initialization to doing everything in pure Swift. As a result, I think that Flutter is no longer globally imported in the module. Hence the need to add it back in.

packages/datadog_flutter_plugin/example/lib/logging_screen.dart
fuzzybinary: Yeah, converting a lot of this over it seems like folks might want a default logger.

packages/datadog_flutter_plugin/lib/datadog_flutter_plugin.dart
fuzzybinary: Yes triple slash is the Dart Doc indicator.

PR #486, opened 2023-10-02
feat: Add `attachToExisting` features back in for v2

fuzzybinary:
`attachToExisting` is how native apps that add a Flutter component can still interact with Datadog. It was commented out to simplify updating to v2. This change adds the functionality back in..

Also update the hybrid application to work with new interfaces.

refs: RUM-1066

### Review checklist

- [x] Run unit tests
- [x] Run integration tests


examples/native-hybrid-app/flutter_module/lib/main.dart
```diff
@@ -2,19 +2,21 @@
...
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  // final config = DdSdkExistingConfiguration(
-  //   loggingConfiguration: LoggingConfiguration(),
-  //   detectLongTasks: true,
-  // )..enableHttpTracking();
+  final config = DatadogAttachConfiguration(
+    detectLongTasks: true,
+    reportFlutterPerformance: true,
```
0xnm, line 16: it is quite a difference with what was before, did we have it enabled before?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -251,17 +253,23 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
-//
-//        return mapOf<String, Any>(
-//            "rumEnabled" to rumEnabled
-//        )
-//    }
+    private fun attachToExisting(): Map<String, Any> {
+        var loggingEnabled = false
+        if (Logs.isEnabled()) {
+            loggingEnabled = true
+        }
```
0xnm: it should be possible just to write `val loggingEnabled = Logs.isEnabled()`?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -251,17 +253,23 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
...
+
+        var rumEnabled = false
+        if (GlobalRumMonitor.isRegistered()) {
+            rumEnabled = true
```
0xnm:
```suggestion
        val rumEnabled = GlobalRumMonitor.isRegistered()
        if (rumEnabled) {
```

packages/datadog_flutter_plugin/ios/Classes/DatadogSdkPlugin.swift
```diff
@@ -294,21 +296,29 @@ public class DatadogSdkPlugin: NSObject, FlutterPlugin {
...
-//        return [
-//            "rumEnabled": rumEnabled
-//        ]
-//    }
-
-    // This is a way to work around https://github.com/flutter/flutter/issues/126671,
-    //
+    private func attachToExisting() -> [String: Any?] {
+        var rumEnabled = false
+        var logsEnabled = false
```
0xnm, line 301: we can probably check if logs and RUM is enabled at the init stage and avoid having `var`

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -254,104 +254,119 @@ class DatadogConfiguration {
...
+
+  /// Whether to report Flutter specific performance metrics (build and raster
+  /// times)
+  ///
+  /// This uses the [SchedulerBinding.addTimingsCallback] method to report build
+  /// and raster times for views, and has a documented negligible impact on
+  /// performance.
+  ///
+  /// Defaults to false
+  bool reportFlutterPerformance = false;
```
0xnm, line 295: should it be a part of this PR? it seems we didn't have it before?

packages/datadog_flutter_plugin/lib/src/datadog_sdk_method_channel.dart
```diff
@@ -72,6 +72,8 @@ class DatadogSdkMethodChannel extends DatadogSdkPlatform {
         .invokeMapMethod<String, Object?>(
             'attachToExisting', <String, Object?>{});
 
+    print('channelResponse $channelResponse');
```
0xnm: some debugging leftover it seems

packages/datadog_flutter_plugin/lib/src/rum/ddrum.dart
```diff
@@ -99,39 +102,61 @@ class DatadogRum {
...
+    _init(
+      core: core,
+      detectLongTasks: config.detectLongTasks,
+      longTaskThreshold: config.longTaskThreshold,
+      reportFlutterPerformance: config.reportFlutterPerformance,
+    );
+  }
+
+  DatadogRum._(DatadogSdk core, DatadogRumConfiguration configuration)
+      : tracingSamplingRate = configuration.tracingSamplingRate,
```
0xnm: I think in native SDKs we try to use `trace` instead of `tracing`, e.g. `traceSampleRate`, so we need to align naming here as well.

packages/datadog_flutter_plugin/lib/src/rum/ddrum.dart
```diff
@@ -87,9 +87,12 @@ class DatadogRum {
     return DdRumPlatform.instance;
   }
 
-  final sampleRandom = Random();
+  /// The sampling rate for tracing resources.
+  ///
+  /// See [RumConfiguration.tracingSamplingRate]
```
0xnm: I cannot find `RumConfiguration` file in the branch https://github.com/DataDog/dd-sdk-flutter/tree/feature/v2, is it a correct reference? Also it is strange to see notion of `tracing` for RUM here. Should it be in the HTTP plugin support instead only? But I guess it is coming from v1 anyway.

examples/native-hybrid-app/flutter_module/lib/main.dart
fuzzybinary: In the example, no I don't think we did, but probably better to have it enabled.

packages/datadog_flutter_plugin/ios/Classes/DatadogSdkPlugin.swift
fuzzybinary: Yeah, that's fair. I'll fix.

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
fuzzybinary: We didn't, but that was an oversight. I think adding it here makes sense.

packages/datadog_flutter_plugin/lib/src/rum/ddrum.dart
fuzzybinary: Good catch. The configuration was indeed renamed and I missed the doc reference.

Yes, it does only refer to http tracing at the moment, and coming from v1. However, we have several packages that do HTTP tracing and I think it might make sense for them to share the configuration?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
fuzzybinary: Yes. I thought I needed more code here like in RUM.

examples/native-hybrid-app/flutter_module/.vscode/launch.json
```diff
@@ -0,0 +1,13 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
xgouchet, line 4: > [!NOTE]
JSON is a data-only format, and using `//` style comments could break some parsing tools (apparently not VSCode 😅 ).
I usually go with an attribute that is ignored and won't break the parsing or code style in editors, e.g.:
```
{
  "comment": "My comment about something"
}
```

xgouchet (approved):

maycmlee (approved):

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -254,104 +254,119 @@ class DatadogConfiguration {
...
+  }
+
+  /// A list of first party hosts and the types of tracing headers Datadog
+  /// should automatically inject on resource calls. This is used in conjunction
+  /// with Datadog network tracking packages like `datadog_tracking_http_client`
+  ///
+  /// This property only affects network requests made from Flutter, and is not
+  /// shared with or populated from existing the SDK.
+  ///
+  /// See [DdSdkConfiguration.firstPartyHostsWithTracingHeaders] for more information
```
0xnm: probably a wrong reference?

```suggestion
  /// See [DatadogConfiguration.firstPartyHostsWithTracingHeaders] for more information
```

packages/datadog_flutter_plugin/lib/src/rum/ddrum.dart
```diff
@@ -87,9 +87,12 @@ class DatadogRum {
     return DdRumPlatform.instance;
   }
 
-  final sampleRandom = Random();
+  /// The sampling rate for tracing resources.
```
0xnm:
```suggestion
  /// The sample rate for tracing resources.
```

0xnm (commented):
lgtm, just some final suggestions

examples/native-hybrid-app/flutter_module/.vscode/launch.json
fuzzybinary: This is technically `jsonc`, which is literally just json with comments.

packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
fuzzybinary: Thanks, I've made task to do a full scan over documentation for renames.

0xnm (approved):

PR #488, opened 2023-10-06
feat: Fix portions of web functionality

fuzzybinary:
### What and why?

Some features of the Flutter SDK that could work in Web weren't working.  While we still don't have full feature parity for Flutter Web, this should bring us closer.

In addition, we're adding web integration tests to the CI to ensure that we at least keep current levels of functionality moving forward.

### Review checklist

- [ ] Run unit tests
- [ ] Run integration tests

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/logs/ddlogs_web.dart
+++ b/packages/datadog_flutter_plugin/lib/src/logs/ddlogs_web.dart
@@ -11,6 +11,7 @@ import 'package:js/js.dart';
 import '../../datadog_flutter_plugin.dart';
 import '../web_helpers.dart';
 import 'ddlogs_platform_interface.dart';
+import 'ddweb_helpers.dart';
 
 class DdLogsWeb extends DdLogsPlatform {
   final Map<String, Logger> _activeLoggers = {};
@@ -19,6 +20,7 @@ class DdLogsWeb extends DdLogsPlatform {
     init(_LogInitOptions(
       clientToken: configuration.clientToken,
       env: configuration.env,
+      proxyUrl: configuration.loggingConfiguration?.customEndpoint,
       site: siteStringForSite(configuration.site),
       service: configuration.service,
       version: configuration.versionTag,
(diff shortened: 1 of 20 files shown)


packages/datadog_flutter_plugin/lib/src/logs/ddlogs_web.dart
```diff
@@ -84,7 +86,19 @@ class DdLogsWeb extends DdLogsPlatform {
...
+      error = JSError();
+      error.stack = convertWebStackTrace(errorStackTrace);
+      error.message = errorMessage;
+      error.name = errorKind ?? 'Error';
+    }
+    logger?.log(
+      message,
+      valueToJs(attributes, 'attributes'),
+      webLogLevel,
+      error,
```
0xnm, line 100: error can be uninitialized here if `if` block above is not executed? what will be the behavor in this case, will we have a crash?

packages/datadog_flutter_plugin/lib/src/web_helpers.dart
```diff
@@ -52,3 +53,30 @@ dynamic valueToJs(Object? value, String parameterName) {
   throw ArgumentError(
       'Could not convert ${value.runtimeType} to javascript.', parameterName);
 }
+
+String? convertWebStackTrace(StackTrace? stackTrace) {
+  if (stackTrace == null) return null;
+
+  final dartLineRegex =
+      RegExp(r'(?<file>.+) (?<location>\d+:\d+)\s*(?<function>.+)');
```
0xnm: should we extract it a static field or simply class property? it is same expression everytime, so maybe we can create the object once and compile the expession, no need to create another one with each method invocation.

packages/datadog_flutter_plugin/lib/src/logs/ddlogs_web.dart
fuzzybinary: Dart defaults to `null` on a nullable value, and `log` supports null errors so we should be good.

packages/datadog_flutter_plugin/lib/src/web_helpers.dart
fuzzybinary: Yes, good call. I originally tried to make it `const` but apparently RegExp doesn't have a const constructor.

packages/datadog_flutter_plugin/lib/src/logs/ddlogs_web.dart
fuzzybinary: (And this case should be handled in the integration tests)

0xnm (approved):

PR #526, opened 2023-12-05
fix:  Fix Android losing data on backgrounding

fuzzybinary:
### What and why?

Android is recreating the main DatadogSdkPlugin on background / foregrounding the application, which recreates the DatadogRumPlugin, which loses its previous reference to the RUMMonitor.  This is causing telemetry to report that RUM is being used when it's not enabled, and likely leading to lost data.

The fast fix is to have DatadogRumPlugin be a singleton, which is what's done here. I've made DatadogLogsPlugin one as well, although that was likely not needed.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogsPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogsPlugin.kt
@@ -24,7 +24,7 @@ import java.util.concurrent.TimeUnit
 import org.json.JSONArray
 import org.json.JSONObject
 
-class DatadogLogsPlugin : MethodChannel.MethodCallHandler {
+class DatadogLogsPlugin internal constructor() : MethodChannel.MethodCallHandler {
     companion object LogParameterNames {
         const val LOG_LEVEL = "logLevel"
         const val LOG_MESSAGE = "message"
@@ -38,6 +38,10 @@ class DatadogLogsPlugin : MethodChannel.MethodCallHandler {
 
         // See DatadogSdkPlugin's description of this same member
         private var previousConfiguration: Map<String, Any?>? = null
+
+        val instance: DatadogLogsPlugin by lazy {
+            DatadogLogsPlugin()
+        }
 
         // For testing purposes only
         internal fun resetConfig() {
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
@@ -62,8 +62,9 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         SynchronousQueue()
     )
 
-    val logsPlugin: DatadogLogsPlugin = DatadogLogsPlugin()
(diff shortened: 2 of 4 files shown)


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogsPlugin.kt
```diff
@@ -39,6 +39,10 @@ class DatadogLogsPlugin : MethodChannel.MethodCallHandler {
         // See DatadogSdkPlugin's description of this same member
         private var previousConfiguration: Map<String, Any?>? = null
 
+        val instance:DatadogLogsPlugin by lazy {
```
0xnm: this will prevent us from having multiple SDK instances (one of the major feature of SDK v2 arch)?

fuzzybinary: Yes, but unfortunately not supported in Flutter right now and other issues prevent it at the moment anyway. (it wasn't a goal of Flutter v2)

0xnm: Ok, fair. It is just for the future as well, because once we support it in Flutter, we will need to get rid of singleton.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -62,8 +62,9 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         SynchronousQueue()
     )
 
-    val logsPlugin: DatadogLogsPlugin = DatadogLogsPlugin()
-    val rumPlugin: DatadogRumPlugin = DatadogRumPlugin()
+    // Because Flutter recreates the plugin, we have to rely on singletons for the contained plugins
+    val logsPlugin: DatadogLogsPlugin = DatadogLogsPlugin.instance
+    val rumPlugin: DatadogRumPlugin = DatadogRumPlugin.getInstance()
```
0xnm: is there a reason why logs is using property to get a singleton and RUM is calling a method?

0xnm (dismissed):

xgouchet (dismissed):

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
fuzzybinary: Yes, though it might not be a good reason? `DatadogRumPlugin` can take a parameter to its constructor which is used in unit tests, while `DatadogLogsPlugin` does not, so I took separate approaches to creating / getting the instances (`lazy` vs. `getInstance()`)

Do you think I should try to standardize them more?

0xnm: This is not a blocker for me definitely, I was just curious of the reason :) So as you wish.

fuzzybinary: Nope, nevermind, I switched to using `by lazy` for both, since constructors need to be marked as internal anyway.

0xnm (approved):

PR #540, opened 2024-01-11
feat: Add an accessor for current session Id

fuzzybinary:
### What and why?

There are use cases where clients want to know the current Session Id. This provides an accessor to the current Id.

refs: #442 RUM-731

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
@@ -116,6 +116,7 @@ class DatadogRumPlugin internal constructor(
             when (call.method) {
                 "enable" -> enable(call, result)
                 "deinitialize" -> deinitialize(call, result)
+                "getCurrentSessionId" -> getCurrentSessionId(call, result)
                 "startView" -> startView(call, result)
                 "stopView" -> stopView(call, result)
                 "addTiming" -> addTiming(call, result)
@@ -190,6 +191,14 @@ class DatadogRumPlugin internal constructor(
         rum = null
 
         result.success(null)
+    }
+
+    private fun getCurrentSessionId(call: MethodCall, result: MethodChannel.Result) {
+        rum?.let {
+            it.getCurrentSessionId { sessionId ->
+                result.success(sessionId)
+            }
+        }
     }
 
     fun attachToExistingSdk(monitor: RumMonitor) {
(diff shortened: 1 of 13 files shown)


xgouchet (dismissed):

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -192,6 +193,16 @@ class DatadogRumPlugin internal constructor(
         result.success(null)
     }
 
+    private fun getCurrentSessionId(call: MethodCall, result: MethodChannel.Result) {
+        rum?.let {
+            it.getCurrentSessionId { sessionId ->
+                result.success(sessionId)
+            }
+        } ?: run {
```
0xnm: what is the reason to use `run` here? I think it is possible to write just `?: result.invalidOperation("RUM is not enabled")`

packages/datadog_flutter_plugin/example/lib/rum_screen.dart
```diff
@@ -27,6 +35,14 @@ class _RumScreenState extends State<RumScreen> {
     super.dispose();
   }
 
+  void _getCurrentSessionId() async {
+    _currentSessionId = await DatadogSdk.instance.rum?.getCurrentSessionId();
+
+    if (mounted) {
+      setState(() {});
+    }
```
0xnm, line 43: just out of curiosity: what this thing is doing?

0xnm (dismissed):

packages/datadog_flutter_plugin/example/lib/rum_screen.dart
fuzzybinary: In order to let Flutter know that the State (the object) has changed, you have to call `setState` so it marks the attached Widget as dirty. Usually, you put the modification of the state in the callback, but it's not necessary. Because I modified `currentSessionId` I have to then call `setState` to re-render. But, since it's an async method, there's a possibility the Widget / State has been detached from the tree, so I have to check `mounted` first (which checks if the widget is in the tree) to avoid an exception.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
fuzzybinary: I actually don't need to call the result at all, as `onMethodCall` already checks for the existence of `rum`.

0xnm (dismissed):

xgouchet (approved):

0xnm (approved):

PR #543, opened 2024-01-15
Fix tracecontext header by lowercase the value

tatashidayat:
### What and why?

We've observed that the generated traceparent header from DD is UpperCase. It makes Otel doesn't recognize it as a valid traceparent.

|Uppercase (DD generated)|Lowecase (Proposed)|
|:---|:---|
|Uppercase|Lowercase|

image

reference:
* https://www.w3.org/TR/trace-context/#traceparent-header-field-values

### How?
LowerCase the traceparent header value

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -208,7 +208,7 @@ Map<String, String> getTracingHeaders(
         'o:rum',
         'p:$spanString',
       ].join(';');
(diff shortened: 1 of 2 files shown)


fuzzybinary:
Hi @tatashidayat,

Thanks for bringing this to our attention! It looks like uppercasing was being done erroneously on all trace ids, and they should be all lowercase, so I've opened #544 to address it across the board.

I'll close this in favor of that.

Thanks again!

fuzzybinary:
This is fixed in `2.2.0`

tatashidayat:
Thanks @fuzzybinary, I'll try 2.2.0

Issue #554, opened 2024-02-02
Required headers to connect RUM and Traces are not being added when using Flutter web


Gummiees:
**Describe what happened**

I have set up a Datadog RUM **Flutter web only** application successfully, but when connecting RUM and traces [following the docs](https://docs.datadoghq.com/real_user_monitoring/mobile_and_tv_monitoring/setup/flutter/#setup), it does not set the headers.

When opening the DevTools and checking the console, Datadog has printed multiple warnings (one per first party host), saying:

```
datadog-rum-slim.js:1 Datadog Browser SDK: Allowed Tracing Urls parameters should be a string, RegExp, function, or an object. Ignoring parameter 
```

<details>
<summary>Full Chrome printed log</summary>

```
datadog-rum-slim.js:1 Datadog Browser SDK: Allowed Tracing Urls parameters should be a string, RegExp, function, or an object. Ignoring parameter Object
(anonymous) @ datadog-rum-slim.js:1
(anonymous) @ datadog-rum-slim.js:1
wn @ datadog-rum-slim.js:1
(anonymous) @ datadog-rum-slim.js:1
l @ datadog-rum-slim.js:1
(anonymous) @ datadog-rum-slim.js:1
initialize @ ddrum_web.dart:27
initialize @ datadog_flutter_plugin_web.dart:77
runBody @ async_patch.dart:84
_async @ async_patch.dart:127
initialize @ datadog_flutter_plugin_web.dart:55
initialize @ datadog_flutter_plugin.dart:168
(anonymous) @ async_patch.dart:45
runUnary @ zone.dart:1661
handleValue @ future_impl.dart:162
handleValueCallback @ future_impl.dart:846
_propagateToListeners @ future_impl.dart:875
[_completeWithValue] @ future_impl.dart:647
(anonymous) @ future_impl.dart:721
_microtaskLoop @ schedule_microtask.dart:40
_startMicrotaskLoop @ schedule_microtask.dart:49
(anonymous) @ async_patch.dart:181
Promise.then (async)
_scheduleImmediateWithPromise @ async_patch.dart:179
_scheduleImmediate @ async_patch.dart:151
_scheduleAsyncCallback @ schedule_microtask.dart:69
_rootScheduleMicrotask @ zone.dart:1466
scheduleMicrotask @ zone.dart:1681
[_asyncCompleteWithValue] @ future_impl.dart:720
[_asyncComplete] @ future_impl.dart:680
[...]

PR #556, opened 2024-02-02
fix: Error setting firstPartyHosts in Flutter Web

fuzzybinary:
### What and why?

The Browser SDK is checking if the type supplied to firstPartyHosts is a RegExp, but the Dart RegExp returns `false` for that check.  Fix it so that we create a JavaScript RegExp.

Note, this only works because the RegExp is very simple in this case. More complicated RegExes may not convert properly.

refs: #554

### How?

### Review checklist


git-thuerk-done (approved):
looks good for docs

0xnm (approved):

Issue #568, opened 2024-02-28
[iOS] Variant doesn't appear to be applied to logs


ellis-cot:
### Describe the bug

Since symbolication appears to use the variant to find the correct symbols, this causes "Symbolication failed" errors for logs with stack traces.

We're obfuscating our build and uploading the symbols/maps with the `--flavor` argument as recommended here: https://docs.datadoghq.com/real_user_monitoring/error_tracking/mobile/flutter/#advanced-configuration---flavors-and-build-numbers. 
We also supply the `flavor` parameter to the DatadogConfiguration object.

Android logs are tagged with the variant and are symbolicated correctly, we get an error message for iOS logs.

### Reproduction steps

Obfuscate a build and upload the symbols with the `--flavor` parameter.
Initialise the DataDog SDK instance with flavor set.
Create a log that includes a stack trace.
See if the log has been symbolicated.

### Expected behavior

Stack traces are symbolicated and the logs are tagged with the variant

### Affected SDK versions

2.2.0

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Flutter Version

3.16.5

### Setup Type

Flutter application

### Device Information

iOS

fuzzybinary:
Hi @ellis-cot,

Thanks for reporting. This does seem to be the case. I'll need to add variant reporting to the iOS SDK and I'll release a fix to Flutter ASAP.

Sorry for the inconvenience!

ellis-cot:
Marvellous, thanks for the swift response @fuzzybinary

fuzzybinary:
This should now be available in 2.4.0. Please let me know if you see any issues.

PR #569, opened 2024-02-29
fix: Add test for variant on logs

fuzzybinary:
### What and why?

iOS was not properly adding `flavor` / `variant` to logs. Add an integration test that checks for this.

Also fix `source_type` to use an internal value over attempting to set `error.source_type`

This is dependent on DataDog/dd-sdk-ios#1702

refs: #568

### How?

### Review checklist


examples/simple_example/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>11.0</string>
+  <string>12.0</string>
```
0xnm: is this intended change?

0xnm (dismissed):

examples/simple_example/ios/Flutter/AppFrameworkInfo.plist
fuzzybinary: No, good catch. However, with that said Flutter is going to keep changing it so I might as well let it go through. I'm not sure when iOS intends on dropping iOS 11 support but I'm thinking Flutter might need to sooner rather than later.

0xnm (approved):

PR #578, opened 2024-03-15
feat: Support global log attributes

fuzzybinary:
### What and why?

This allows attributes to be added, changed, and removed from all loggers managed by Datadog.

refs: RUM-3071

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogLogsPluginTest.kt
+++ b/packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogLogsPluginTest.kt
@@ -215,6 +215,13 @@ class DatadogLogsPluginTest {
     }
 
     private val contracts = listOf(
+        Contract("addGlobalAttribute", mapOf(
+            "key" to ContractParameter.Type(SupportedContractType.STRING),
+            "value" to ContractParameter.Type(SupportedContractType.MAP),
+        )),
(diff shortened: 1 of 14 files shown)


packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogLogsPluginTest.kt
```diff
@@ -215,6 +215,13 @@ class DatadogLogsPluginTest {
     }
 
     private val contracts = listOf(
+        Contract("addGlobalAttribute", mapOf(
+            "key" to ContractParameter.Type(SupportedContractType.STRING),
+            "value" to ContractParameter.Type(SupportedContractType.MAP),
```
0xnm, line 220: should it be `Any` here instead of `MAP`?

0xnm (dismissed):

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogLogsPluginTest.kt
fuzzybinary: The contracts don't really have a concept of "Any" so I give it the most complicated type I have.

At some point I may want to go back and add "Any" as a supported type (this is true in iOS as well)

0xnm (approved):

PR #592, opened 2024-04-19
fix: Fix iOS build

fuzzybinary:
### What and why?

Use `trackUserInteractions` over `trackInteractions` in iOS Telemetry object.
Add missing `addFeatureFlagEvaluations` for Android.

### Review checklist


maxep (dismissed):

maxep:
We send both on iOS, I guess we should do the same and drop it!

fuzzybinary:
The build issue is that we did drop it from the Telemetry object, even though we send it to the backend. So I just needed to update to the new parameter.

maxep:
ah yea, sorry my bad..

0xnm (approved):

PR #601, opened 2024-04-30
feat: Support 128-bit trace ids

fuzzybinary:
### What and why?

Adding 128-bit trace id support for distributed tracing.

`datadog_tracking_http_client` and `datadog_grpc_interceptor` didn't change code (though their tests changed) But to be safe I'm changing them to require 2.5.0 in future releases. I'll also be releasing a point release of both packages to restrict them to <2.5.0.

refs: RUM-1831

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -87,15 +87,15 @@ class TracingId {
   const TracingId(this.value);
 
   static TracingId fromString(
-      String? uuid, TracingIdRepresentation representation) {
-    if (uuid == null) {
+      String? id, TracingIdRepresentation representation) {
+    if (id == null) {
       return TracingId(BigInt.zero);
     }
 
     switch (representation) {
       case TracingIdRepresentation.lowDecimal:
       case TracingIdRepresentation.decimal:
-        final value = BigInt.tryParse(uuid);
+        final value = BigInt.tryParse(id);
         if (value != null) {
           return TracingId(value);
         }
@@ -103,17 +103,17 @@ class TracingId {
       case TracingIdRepresentation.hex:
       case TracingIdRepresentation.hex16Chars:
       case TracingIdRepresentation.hex32Chars:
-        final value = BigInt.tryParse(uuid, radix: 16);
+        final value = BigInt.tryParse(id, radix: 16);
         if (value != null) {
           return TracingId(value);
         }
         break;
       case TracingIdRepresentation.highHex16Chars:
         // Take only the last 16 chars
-        if (uuid.length > 16) {
-          uuid = uuid.substring(uuid.length - 16);
+        if (id.length > 16) {
+          id = id.substring(id.length - 16);
         }
-        final value = BigInt.tryParse(uuid, radix: 16);
+        final value = BigInt.tryParse(id, radix: 16);
         if (value != null) {
           return TracingId(value);
         }
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -5,7 +5,6 @@
 import 'dart:math';
 
 import 'package:flutter/foundation.dart';
-import 'package:flutter/material.dart';
 
 import '../../datadog_internal.dart';
 
@@ -27,9 +26,11 @@ enum TracingHeaderType {
 class DatadogHttpTracingHeaders {
   static const traceId = 'x-datadog-trace-id';
   static const parentId = 'x-datadog-parent-id';
-
+  static const tags = 'x-datadog-tags';
   static const samplingPriority = 'x-datadog-sampling-priority';
   static const origin = 'x-datadog-origin';
+
+  static const traceIdTag = '_dd.p.tid';
 }
 
 class OTelHttpTracingHeaders {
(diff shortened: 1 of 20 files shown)


marco-saia-datadog (dismissed):

fuzzybinary:
@ganeshnj Can I get your review on this to be sure I've got everything correct?

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -46,66 +47,140 @@ class W3CTracingHeaders {
   static const tracestate = 'tracestate';
 }
 
-enum TraceIdRepresentation {
+/// Controls how we print a TracingId
+enum TracingIdRepresentation {
```
ganeshnj, line 51: /informational

are these types publicly exposed? If yes, it is breaking change. I don't think we are going to a major version but just to let customers.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -46,66 +47,140 @@ class W3CTracingHeaders {
...
+  // to be able to store all 64-bits properly.
   final BigInt value;
 
-  const TracingUUID(this.value);
+  const TracingId(this.value);
 
-  static TracingUUID fromString(
-      String? uuid, TraceIdRepresentation representation) {
+  static TracingId fromString(
+      String? uuid, TracingIdRepresentation representation) {
```
ganeshnj: /nit

Not a UUID - just an id.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -46,66 +47,140 @@ class W3CTracingHeaders {
...
+        }
+        break;
+      case TracingIdRepresentation.highHex16Chars:
+        // Take only the last 16 chars
+        if (uuid.length > 16) {
+          uuid = uuid.substring(uuid.length - 16);
+        }
+        final value = BigInt.tryParse(uuid, radix: 16);
+        if (value != null) {
+          return TracingId(value);
```
ganeshnj: /question

why do we need this? There is no valid id where we would take the high 64 bits? You could introduce some convenience property/method on the TracingId type to get low or high bits in the required format.

packages/datadog_flutter_plugin/test/tracing/tracing_headers_test.dart
```diff
@@ -6,10 +6,35 @@ import 'package:datadog_flutter_plugin/src/tracing/tracing_headers.dart';
...
+    const low = 0x01445feed89934bb;
+    const high = 0x01222f00d89934ba;
+
+    var combined = BigInt.from(high);
+    combined = (combined << 64) + BigInt.from(low);
+
+    final tracingId = TracingId(combined);
+
+    expect(tracingId.asString(TracingIdRepresentation.hex),
+        '1222f00d89934ba01445feed89934bb');
```
ganeshnj, line 21: /blocking

the generated id looks off here to me. 

Taking from one of my examples
`6634a5530000000035ad5c814c5a1e78`

first 32 bits: timestamp: 6634a553 (hex) -> 1714726227 (dec) ->  Friday, 3 May 2024 08:50:27
second 32 bits: 00000000
last 64 bits: 35ad5c814c5a1e78 (pure random)

we should add some test case on `_generateTraceId` method.

ganeshnj: For reference https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Tests/NetworkInstrumentation/TraceIDGeneratorTests.swift#L17

packages/datadog_gql_link/test/datadog_gql_link_test.dart
```diff
@@ -728,10 +728,11 @@ query UserInfo($id: ID!) {
...
                 () => mockRum.startResource(any(), any(), any(), captureAny()))
             .captured[0] as Map<String, dynamic>;
-        var traceId =
-            BigInt.parse(capturedAttrs[DatadogRumPlatformAttributeKey.traceID]);
+        var traceId = BigInt.parse(
+            capturedAttrs[DatadogRumPlatformAttributeKey.traceID],
+            radix: 16);
         expect(traceId, isNotNull);
-        expect(traceId.bitLength, lessThanOrEqualTo(63));
+        expect(traceId.bitLength, lessThanOrEqualTo(127));
```
ganeshnj: /question
why 127? Is it because of the reserved sign bit? I'm curious if you use two BitInt to hold lower and higher part, would it actually make 100% compliant?

The problem here, the SDKs will send full 128 bits and we will always lose one bit.

ganeshnj (changes requested):
Overall looks good. my main concern is around using 127 bit only support - this will cause data loss when communicating with native SDK side if we are reading a value from the SDK side.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
fuzzybinary: No they're not. They're only "publicly" exposed on the `datadog_internal.dart` portion of the package which is not part of our public interface.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
fuzzybinary: 👍 Changed the type but not the parameter. Will fix.

packages/datadog_flutter_plugin/test/tracing/tracing_headers_test.dart
fuzzybinary: This essentially only testing just the various string representations that can be created from a `BigInt`, and that low / high bits get masked out and padded as appropriate. The low / high bits that are being passed in aren't conformant to the trace id spec.

That said, you're right should add a test that checks that `TracingId` creates a conformant Id when asked.

packages/datadog_gql_link/test/datadog_gql_link_test.dart
fuzzybinary: 🤔 I thought I saw that the top bit on the trace id needed to be zero? I must be mistaken.  I'll change it to allow the full 128 bits.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
fuzzybinary: To be honest, the parsing code probably isn't necessary but added for completeness.

packages/datadog_flutter_plugin/test/tracing/tracing_headers_test.dart
```diff
@@ -6,10 +6,46 @@ import 'package:datadog_flutter_plugin/src/tracing/tracing_headers.dart';
...
+  test('traceId generates proper values', () {
+    final nowSeconds = (DateTime.now().millisecondsSinceEpoch ~/ 1000);
+    final traceId = TracingId.traceId();
+
+    final traceIdString = traceId.asString(TracingIdRepresentation.hex);
+    int traceSeconds = int.parse(traceIdString.substring(0, 8), radix: 16);
+    expect(traceSeconds, closeTo(nowSeconds, 1));
+    expect('00000000', traceIdString.substring(8, 16));
+    expect(traceIdString.substring(16), isNot('0000000000000000'));
+  });
```
ganeshnj, line 43: 🤩

ganeshnj (approved):
LGTM - please do one last check before merge
- having only one header type enabled, the distributed traces are working fine end to end and you can see in APM.
- with default (datadog + tracecontext) works fine end to end

fuzzybinary:
Confirmed in demo environment:
Only Datadog traces enabled: 
datadog trace

Datadog + tracecontext:
image

Issue #596, opened 2024-04-23
No implementation found for method addError on channel datadog_sdk_flutter.rum


androidmitry:
### Stack trace

Fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: MissingPluginException(No implementation found for method addError on channel datadog_sdk_flutter.rum)
       at MethodChannel._invokeMethod(platform_channel.dart:332)
       at ._willHandleError(helpers.dart:14)

### Reproduction steps

Add the datadog_flutter_plugin package, release to App Store

### Volume

0,0021 (1-2 users per day)

### Affected SDK versions

2.4.0

### Does the crash manifest in the latest SDK version?

Yes

### Flutter Version

3.19.5

### Setup Type

Flutter Application 

### Device Information

OS - Android 
per version:
Android 12 - 88% 
Android 10 - 7%
Android 14 - 3%
Android 13 - 2%

per device:
Samsung - 93%
Oneplus - 7%

### Other relevant information

Device states: background 60%

fuzzybinary:
Hi @androidmitry ,

Thanks for the report, I'll look into this as soon as I can.

Can you give me anymore information about a possible reproduction? Have you been able to reproduce locally at all? Is there anything strange about your setup that might be disconnecting the MethodChannel from our plugin?  We tend to wrap every call we make to try to avoid crashes, so I'm very concerned that this is causing a crash....

androidmitry:
Hi @fuzzybinary , unfortunately thats all information I have so far. I wasn't able to reproduce it. We had some custom platform code, but it was removed. Whats interesting is that number of reports is decreasing. I will update the issue if crash goes away.

fuzzybinary:
@androidmitry Yeah if you can keep me posted I would appreciate it.  

I'm seen issues in the past where the method channel can get disconnected from the plugin, but I've fixed those, and most threw errors in the native layer, not Dart.

nirmal0707:
This issue is also occurring in version 2.1.0, and we have encountered the MissingPluginException from the Android channels datadog_sdk_flutter.rum and datadog_sdk_flutter.logs in the production release. Due to consecutive RUM events, the error count is excessively high. Below are some error messages we've received:

1. MissingPluginException(No implementation found for method addError on channel datadog_sdk_flutter.rum)
2. MissingPluginException(No implementation found for method createLogger on channel datadog_sdk_flutter.logs)
[...]

Issue #580, opened 2024-03-19
Is it not possible to use data dog from flutter isolate ?


Den-creator:
### Question

By calling DatadogSdk.instance.initialize in isolate I've receive such error:  

[ERROR:flutter/runtime/dart_isolate.cc(1097)] Unhandled exception:
'package:flutter/src/services/platform_channel.dart': Failed assertion: line 542 pos 7: '_binaryMessenger != null || BindingBase.debugBindingType() != null': Cannot set the method call handler before the binary messenger has been initialized. This happens when you call setMethodCallHandler() before the WidgetsFlutterBinding has been initialized. You can fix this by either calling WidgetsFlutterBinding.ensureInitialized() before this or by passing a custom BinaryMessenger instance to MethodChannel().
#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      MethodChannel.setMethodCallHandler (package:flutter/src/services/platform_channel.dart:542:7)
#3      DatadogSdkMethodChannel.initialize (package:datadog_flutter_plugin/src/datadog_sdk_method_channel.dart:56:21)
#4      DatadogSdk.initialize (package:datadog_flutter_plugin/datadog_flutter_plugin.dart:168:21)
<asynchronous suspension>

But WidgetsFlutterBinding.ensureInitialized() can not be called in isolate...

fuzzybinary:
Yes this is limitation of the Datadog SDK at the moment, you can only use it from the main isolate at the moment.

If you have a need for this, please raise it as a feature request. It's definitely doable we just haven't had any requests for the functionality.

Den-creator:
@fuzzybinary thank you for you reply !

dballance:
@fuzzybinary - We need this for our impl. We've been stuck back on the 3rd party plugin but need to upgrade now due to the Apple Privacy changes.

How can we request this change?

For clarity - we run a single FlutterEngine with two isolates. On Android - one isolate runs within a foreground service and runs even when the UI isolate is terminated. On iOS it's similar - but the runtime model for iOS means both isolates are always running. 

We primarily log from one isolate (background) but that isolate is always the *second* isolate to be established.

[...]

PR #626, opened 2024-06-24
fix: Do not use singletons for RUM / Logs plugins

fuzzybinary:
### What and why?

Under certain circumstances, Flutter can create two FlutterEngines which each have their own method channels. If this happens, we could end up in a situation where RUM and Logs were not properly attached to a MethodChannel, resulting in error messages and lost calls.

refs: #596 #580 RUM-491 RUM-4438

### How?

We're removing the `DatadogRumPlugin` and `DatadogLogsPlugin` singletons and replacing them with instances to avoid this.  Both plugins will attach to existing Datadog Logs and RUM instances during initialization if they have already been inititalized.

However, this breaks our current mapper implementation. Since Datadog expects the mapper during initialization, and holds it during the life of the application, we have to move the mappers to a companion object of the `*Plugin` classes to avoid issues with multiple or disconnected `MethodChannels`.

This potentially also helps us with multiple isolate tracking, but is just the first step of that.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
+++ b/packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
@@ -1,0 +1,113 @@
+package com.datadoghq.flutter
+
+import android.os.Handler
+import android.os.Looper
+import com.datadog.android.Datadog
+import com.datadog.android.log.LogsConfiguration
+import com.datadog.android.log.model.LogEvent
+import io.flutter.plugin.common.MethodChannel
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+/**
+ * This is a helper class that that simplifies event mapping / scrubbing for Log events.
+ *
+ * Since it is possible for Flutter engines to shut down and be recreated, it is possible that the
+ * object used as the event mapper to be lost. This class is used as a static instance that will
+ * perform mapping and call the Flutter mapping methods using the last provided MethodChannel.
+ */
+internal class DatadogLogEventMapper {
+    private val channels: MutableSet<MethodChannel> = mutableSetOf()
+
+    fun addChannel(channel: MethodChannel) {
+        channels.add(channel)
+    }
+
+    fun removeChannel(channel: MethodChannel) {
+        channels.remove(channel)
+    }
+
+    fun attachMapper(config: LogsConfiguration.Builder): LogsConfiguration.Builder {
+        config.setEventMapper { event -> mapLogEvent(event) }
+
+        return config
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    internal fun mapLogEvent(event: LogEvent): LogEvent? {
+        val jsonEvent = event.toJson().asFlutterMap()
+        var modifiedJson: Map<String, Any?>? = null
+
+        val latch = CountDownLatch(1)
+
+        val handler = Handler(Looper.getMainLooper())
+        handler.post {
+            try {
+                val channel = channels.first()
+                channel.invokeMethod(
+                    "mapLogEvent",
+                    mapOf(
+                        "event" to jsonEvent
+                    ),
+                    object : MethodChannel.Result {
+                        override fun success(result: Any?) {
+                            @Suppress("UNCHECKED_CAST")
+                            modifiedJson = result as? Map<String, Any?>
+                            latch.countDown()
+                        }
+
+                        override fun error(
+                            errorCode: String,
+                            errorMessage: String?,
+                            errorDetails: Any?
+                        ) {
+                            // No telemetry needed, this is likely an issue in user code
+                            latch.countDown()
+                        }
+
+                        override fun notImplemented() {
+                            Datadog._internalProxy()._telemetry.error(
+                                "mapLogEvent returned notImplemented."
+                            )
+                            latch.countDown()
+                        }
+                    }
+                )
+            } catch (e: Exception) {
+                Datadog._internalProxy()._telemetry.error("Attempting call mapLogEvent failed.", e)
+                latch.countDown()
+            }
+        }
+
+        try {
+            // Stalls until the method channel finishes
+            if (!latch.await(1, TimeUnit.SECONDS)) {
+                Datadog._internalProxy()._telemetry.debug("logMapper timed out")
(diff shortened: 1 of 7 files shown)


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
```diff
@@ -0,0 +1,113 @@
...
+    internal fun mapLogEvent(event: LogEvent): LogEvent? {
+        val jsonEvent = event.toJson().asFlutterMap()
+        var modifiedJson: Map<String, Any?>? = null
+
+        val latch = CountDownLatch(1)
+
+        val handler = Handler(Looper.getMainLooper())
+        handler.post {
+            try {
+                val channel = channels.first()
```
0xnm: it is a possible crash point here - if there is no channel by some reason, it will throw `NoSuchElementException`. So maybe you want to use `firstOrNull`? Also, to avoid concurrency, maybe it is better to get the channel reference before making `handler.post` call?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumEventMapper.kt
```diff
@@ -0,0 +1,330 @@
...
+        mapperName: String,
+        event: T,
+        encodedEvent: Map<String, Any?>,
+        completion: (Map<String, Any?>?, T) -> T?
+    ): T? {
+        var modifiedJson: Map<String, Any?>? = encodedEvent
+        // Any valid channel should do here since they should be connected to the same initialization code,
+        // but calling on multiple could result in weird behavior and performance issues. While "first" may
+        // not be the actual first registered channel, it should be fine for our purposes.
+        val channel = channels.first()
```
0xnm: should we handle a situation when there is no channels registered by some reason?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
```diff
@@ -82,25 +67,27 @@ class DatadogRumPlugin internal constructor(
...
     fun attachToEngine(flutterPluginBinding: FlutterPlugin.FlutterPluginBinding) {
         channel = MethodChannel(flutterPluginBinding.binaryMessenger, "datadog_sdk_flutter.rum")
         channel.setMethodCallHandler(this)
+        eventMapper.addChannel(channel)
 
         binding = flutterPluginBinding
+
+        if (GlobalRumMonitor.isRegistered()) {
+            rum = GlobalRumMonitor.get()
+        }
```
0xnm, line 86: why it is needed to keep the reference, when we can simply call `GlobalRumMonitor.get()` each time?

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * This is a helper class that that simplifies event mapping / scrubbing for Log events.
+ *
+ * Since it is possible for Flutter engines to shut down and be recreated, it is possible that the
+ * object used as the event mapper to be lost. This class is used as a static instance that will
+ * perform mapping and call the Flutter mapping methods using the last provided MethodChannel.
+ */
+internal class DatadogLogEventMapper {
+    private val channels: MutableSet<MethodChannel> = mutableSetOf()
```
0xnm: also the thread-safe access collection should be used here if it is possible to access it from the different threads

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
fuzzybinary: Yeah I'll move it out for concurrency and use `firstOrNull` instead. I originally figured `first` would get caught in the try / catch there but probably better not to rely on that.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogRumPlugin.kt
fuzzybinary: It's not now, but it's a smaller change to keep the reference for this PR. Do you think it's an issue to keep it the way it is?

0xnm: not a big deal, but if SDK was stopped, `GlobalRumMonitor.get()` will properly return no-op instance, while in this case Flutter will be holding kind of dead instance.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    @Suppress("TooGenericExceptionCaught")
+    internal fun mapLogEvent(event: LogEvent): LogEvent? {
+        val jsonEvent = event.toJson().asFlutterMap()
+        var modifiedJson: Map<String, Any?>? = null
+
+        val latch = CountDownLatch(1)
+
+        val handler = Handler(Looper.getMainLooper())
+        val channel = channels.firstOrNull() ?: return event
```
0xnm, line 46: just fyi: set/map doesn't have by default a guarantee for the items/keys ordering, so `first` may return any item from the set, not necessary first added. Not sure if it matters for you.

Upd: nevermind, you already have a similar comment in the `DatadogRumPlugin` 👍

0xnm (approved):

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogLogEventMapper.kt
fuzzybinary: Yeah, I'm not sure I love the compromise but since the every engine is only ever working of a single codebase, any still active channel should hit the same code.

PR #628, opened 2024-06-27
feat: Add configurable trace header injection

fuzzybinary:
### What and why?

Currently, the libraries that allow Datadog distributed always inject their sampling decision, which has downstream effects and prevents tail based sampling. Allow header injection to be configured so that only sampled traces inject their tracing headers.

refs: RUM-3538

### How?

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -225,16 +225,14 @@ Map<String, String> getTracingHeaders(
 
   switch (headersType) {
     case TracingHeaderType.datadog:
-      if (context.sampled) {
+      if (shouldInjectHeaders) {
         headers[DatadogHttpTracingHeaders.traceId] =
             context.traceId.asString(TracingIdRepresentation.lowDecimal);
         headers[DatadogHttpTracingHeaders.tags] =
             '${DatadogHttpTracingHeaders.traceIdTag}=${context.traceId.asString(TracingIdRepresentation.highHex16Chars)}';
         headers[DatadogHttpTracingHeaders.parentId] =
             context.spanId.asString(TracingIdRepresentation.decimal);
-      }
-      headers[DatadogHttpTracingHeaders.origin] = 'rum';
-      if (shouldInjectHeaders) {
+        headers[DatadogHttpTracingHeaders.origin] = 'rum';
         headers[DatadogHttpTracingHeaders.samplingPriority] = sampledString;
       }
       break;
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -6,6 +6,7 @@ import 'dart:math';
 
 import 'package:flutter/foundation.dart';
 
+import '../../datadog_flutter_plugin.dart';
 import '../../datadog_internal.dart';
 
 /// The type of tracing header to inject into first party requests.
@@ -213,11 +214,14 @@ Map<String, Object?> generateDatadogAttributes(
 
 Map<String, String> getTracingHeaders(
   TracingContext context,
-  TracingHeaderType headersType,
-) {
+  TracingHeaderType headersType, {
+  TraceContextInjection contextInjection = TraceContextInjection.all,
+}) {
   var headers = <String, String>{};
 
   final sampledString = context.sampled ? '1' : '0';
+  bool shouldInjectHeaders =
+      context.sampled || contextInjection == TraceContextInjection.all;
 
   switch (headersType) {
     case TracingHeaderType.datadog:
@@ -230,7 +234,9 @@ Map<String, String> getTracingHeaders(
             context.spanId.asString(TracingIdRepresentation.decimal);
       }
       headers[DatadogHttpTracingHeaders.origin] = 'rum';
-      headers[DatadogHttpTracingHeaders.samplingPriority] = sampledString;
+      if (shouldInjectHeaders) {
+        headers[DatadogHttpTracingHeaders.samplingPriority] = sampledString;
+      }
       break;
     case TracingHeaderType.b3:
       if (context.sampled) {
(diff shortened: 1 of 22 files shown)


packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -230,7 +234,9 @@ Map<String, String> getTracingHeaders(
             context.spanId.asString(TracingIdRepresentation.decimal);
       }
       headers[DatadogHttpTracingHeaders.origin] = 'rum';
-      headers[DatadogHttpTracingHeaders.samplingPriority] = sampledString;
+      if (shouldInjectHeaders) {
```
ganeshnj: :blocking

the configuration not only applies to sampling priortiy but also applies to all other types of headers.

In L228, the `context.sampled` should be replaced with `shouldInjectHeaders` ie 

Something like https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift#L78

Here in this case, if customer has set `all` in config and `context.sampled` is false, we are not setting headers.

ganeshnj: I think this applies other type of headers as well except trace context.

ganeshnj (changes requested):

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
fuzzybinary: 👍 this a change in previous functionality then, at least in Flutter. Previously we only wrote the Datadog  headers for the trace id, tags, and parent id when the trace was sampled in, but wrote origin and sampling priority in all cases.

I double check the logic but I think the Datadog headers are the only ones where I set this logic.

fuzzybinary: @ganeshnj running a comparison in the iOS code we have:
  * [Datadog Header](https://github.com/DataDog/dd-sdk-ios/blob/72c95e5edadc6c3bfc2dbaf563f2e92854c7e5cd/DatadogInternal/Sources/NetworkInstrumentation/Datadog/HTTPHeadersWriter.swift#L78) -- `traceId`, `traceIdTag`, `parentId` headers are written regardless of sample state when TraceContextInjection.all (incorrect in current Flutter code, but different from Flutter's previous implementations)
  * [b3](https://github.com/DataDog/dd-sdk-ios/blob/72c95e5edadc6c3bfc2dbaf563f2e92854c7e5cd/DatadogInternal/Sources/NetworkInstrumentation/B3/B3HTTPHeadersWriter.swift#L136) -- unsampled only writes the sampling string "0" when TraceContextInjection.all
  * [b3multi](https://github.com/DataDog/dd-sdk-ios/blob/72c95e5edadc6c3bfc2dbaf563f2e92854c7e5cd/DatadogInternal/Sources/NetworkInstrumentation/B3/B3HTTPHeadersWriter.swift#L125) -- unsampled only writes the `multipleSampled` header when TraceContextInjection.all
  * [tracecontext](https://github.com/DataDog/dd-sdk-ios/blob/72c95e5edadc6c3bfc2dbaf563f2e92854c7e5cd/DatadogInternal/Sources/NetworkInstrumentation/W3C/W3CHTTPHeadersWriter.swift#L90) -- always writes a the full traceparent and trace state when TraceContextInjection.all

So I think everything in Flutter matches iOS except for the Datadog headers.  Can we confirm that that's the expected behavior on Datadog's traces before making the change?

packages/datadog_flutter_plugin/test/tracing/tracing_headers_test.dart
```diff
@@ -72,110 +73,229 @@ void main() {
...
+      context,
+      TracingHeaderType.datadog,
+      contextInjection: TraceContextInjection.sampled,
+    );
 
     expect(headers['x-datadog-trace-id'], isNull);
     expect(headers['x-datadog-parent-id'], isNull);
-    expect(headers['x-datadog-sampling-priority'], '0');
-    expect(headers['x-datadog-origin'], 'rum');
+    expect(headers['x-datadog-sampling-priority'], isNull);
```
ganeshnj, line 199: this is they key part, now upstream systems can decide their own.

packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
```diff
@@ -213,24 +214,27 @@ Map<String, Object?> generateDatadogAttributes(
...
         headers[DatadogHttpTracingHeaders.parentId] =
             context.spanId.asString(TracingIdRepresentation.decimal);
+        headers[DatadogHttpTracingHeaders.origin] = 'rum';
+        headers[DatadogHttpTracingHeaders.samplingPriority] = sampledString;
       }
-      headers[DatadogHttpTracingHeaders.origin] = 'rum';
-      headers[DatadogHttpTracingHeaders.samplingPriority] = sampledString;
       break;
     case TracingHeaderType.b3:
       if (context.sampled) {
```
ganeshnj, line 240: I think we can simplify B3 headers but that needs to be done across the board, not related to this work stream. Basically sending all the headers like we do for datadog headers because currently when a trace is not sampled and trace injection is all

the behavior is going to something like below when all trace headers are enabled.
```
        B3 = 0;
        Traceparent = "00-668e538e00000000f1b3bab076ddd791-3464ad4c72428ec2-00";
        Tracestate = "dd=p:3464ad4c72428ec2;s:0";
        "X-Amzn-Trace-Id" = "Root=1-668e538e-3ce2240c32d121f82c7fc8d3";
        "X-B3-Sampled" = 0;
        "X-Datadog-Parent-Id" = 3775332931489402562;
        "X-Datadog-Sampling-Priority" = 0;
        "X-Datadog-Tags" = "_dd.p.tid=668e538e00000000";
        "X-Datadog-Trace-Id" = 17416469451043100561;
```
I don't think anyone would go this far to have a case like this far tbh.

ganeshnj (approved):

PR #634, opened 2024-07-24
fix: Prevent error when using logging from background isolate

fuzzybinary:
### What and why?

When creating a logger from a background isolate, Flutter would throw because WidgetsFlutterBinding wasn't initialized. This is specifically because we were adding a callback handler, which is not supported from background isolates.

Add checks to both logs and RUM and prevent initialization of things we know won't work from background isolates.

This is not full support for background isolates but usable as a work around.

refs: #580

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
@@ -26,7 +26,9 @@ class DdRumMethodChannel extends DdRumPlatform {
       internalLogger: core.internalLogger,
     );
 
-    methodChannel.setMethodCallHandler(callbackHandler.handleMethodCall);
+    if (ServicesBinding.rootIsolateToken != null) {
+      methodChannel.setMethodCallHandler(callbackHandler.handleMethodCall);
+    }
 
     return methodChannel.invokeMethod('enable', {
       'configuration': configuration.encode(),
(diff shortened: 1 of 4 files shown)


packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
```diff
@@ -26,7 +26,9 @@ class DdRumMethodChannel extends DdRumPlatform {
       internalLogger: core.internalLogger,
     );
 
-    methodChannel.setMethodCallHandler(callbackHandler.handleMethodCall);
+    if (ServicesBinding.rootIsolateToken != null) {
```
dballance, line 29: This potentially could cause issues when using multiple flutters - but I understand this is not fully supported. 

I have zero clue if it actually will impact multiple flutters - I kinda think it WON'T simply because using two FlutterEngines here means we're not in a spawned isolate ever - but just wanted to mention it.

It might be worthwhile, if DD wants to support multiple flutter engines directly, to add a test / example based off of https://github.com/flutter/samples/tree/main/add_to_app/multiple_flutters. 

Appreciate the effort here regardless! 🎉

fuzzybinary: Thanks for the input. I'll see if I can run a quick test with multiple flutters, but my guess there is that each engine will have its own root Isolate and therefore its own `rootIsolateToken`.

plousada (approved):

PR #636, opened 2024-08-05
fix: Wait 1ms during route changes on web

fuzzybinary:
### What and why?

The Browser SDK reads `location` during a call to `startView`. Unfortunately, the RouterObserver is called before the browser's location has changed, leading to a disconnect in what we send in the event.

Wait 1ms after a route change and, if the route hasn't changed in that 1ms, then call startView.

refs: RUM-5416

### Review checklist


packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
```diff
@@ -112,7 +113,20 @@ class DatadogNavigationObserver extends RouteObserver<ModalRoute<dynamic>>
...
       _currentView = viewInfo;
       if (viewInfo != null) {
-        datadogSdk.rum?.startView(viewInfo.name, null, viewInfo.attributes);
+        if (kIsWeb) {
+          // On web, Flutter is informing us of the change in Route before the
+          // browser has had a chance to update the `location`. Wait a frame so
+          // the Browser SDK can properly capture the location
+          Future.delayed(const Duration(milliseconds: 1)).then((_) {
+            // Make sure the view wasn't overwritten in the last ~1ms
+            if (_currentView == viewInfo) {
```
mariusc83, line 122: where is this `viewInfo` coming from ? The way I see it the `viewInfo` is the constant passed to the Future lambda and the `_viewInfo` is the variable one that changes based on the callback in your observer. Here you are making sure that this `lambda` still applies for the right View right ?

mariusc83: Shouldn't you make the `_currentView` volatile in that case ? Is this concept available in dart or is more like in C where there is no copy of the variable in the Thread local memory ?

mariusc83 (dismissed):

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
fuzzybinary: Right, I'm making sure that the Future that `_currentView` hasn't been replaced in the 1ms it took for this Future to execute.

Dart isn't multithreaded, so it has no concept of `volatile`.

mariusc83 (approved):

Issue #639, opened 2024-08-08
Android builds are broken with Flutter 3.24


fuzzybinary:
### Describe the bug

When building with Flutter 3.24, you get the following error message:

> What went wrong:
> Execution failed for task ':app:checkReleaseAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
>      15 issues were found when checking AAR metadata:
> [ snip ]

This appears to be because the datadog-flutter-sdk's compile version is 33, where Flutter now only supports 34+.

### Reproduction steps

* Upgrade to Flutter 3.24
* Build the example app on Android

### Expected behavior

Example application builds

### Affected SDK versions

2.6.0

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Flutter Version

3.24.0

### Setup Type

Flutter Application

fuzzybinary:
Fixed in 2.7.0

PR #640, opened 2024-08-08
feat: Upgrade Android build to SDK 34

fuzzybinary:
### What and why?

Flutter 3.24 appears to have removed support for SDK 33, as we get multiple compiler errors when using it.

This updates all of the Android module to set `compileSdkVersion` to 34.

refs: #639

### Review checklist

### DIFF
--- a/examples/native-hybrid-app/android/app/build.gradle
+++ b/examples/native-hybrid-app/android/app/build.gradle
@@ -11,7 +11,7 @@ plugins {
 
 android {
     namespace 'com.datadoghq.hybrid_flutter_example'
-    compileSdk 33
+    compileSdk 34
 
     defaultConfig {
         applicationId "com.datadoghq.hybrid_flutter_example"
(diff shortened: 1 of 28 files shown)


examples/native-hybrid-app/android/app/build.gradle
```diff
@@ -11,7 +11,7 @@ plugins {
 
 android {
     namespace 'com.datadoghq.hybrid_flutter_example'
-    compileSdk 33
+    compileSdk 34
```
ambushwork, line 14: I noticed that `compileSdk` version is upgraded to 34, but `targetSdk` stays at 32, is that wanted?

ambushwork (approved):

examples/native-hybrid-app/android/app/build.gradle
fuzzybinary: It should be fine, especially in example apps. In reality I may want to switch all the apps to use `flutter.targetSdkVersion` in the future but that can be a separate change.

Issue #652, opened 2024-10-08
RUM Resource Event Mapper is not called sometimes


hakivin:
### Describe the bug

In our 2 Flutter applications, we are using the Datadog SDK to monitor Real User Monitoring (RUM) events. We have implemented a resource event mapper to filter query parameters from the RUM resource events. 

```dart
RumResourceEvent? _mapRumResourceEvent({
  required RumResourceEvent event,
}) {
  String finalUrl = event.resource.url;
  if (event.resource.type == RumResourceType.native) {
    final uri = Uri.tryParse(event.resource.url);
    if (uri != null) {
      if (_isExcludedResource(event.resource)) {
        return null;
      }
      // remove query params
      final url = '${uri.origin}${uri.path}';
      finalUrl = _maskUrl(url, method: event.resource.method);
    }
  }
  return event.copyWith(
    resource: event.resource.copyWith(
      url: finalUrl,
    ),
  );
}
```

We have observed that the resource event mapper is not consistently called for all resource events.

![Image](https://github.com/user-attachments/assets/e911bce6-b88b-497f-a4a8-0cffdf82dbd2)

We would appreciate any insights into why this issue happened. If there are any additional steps we can take to debug or resolve this issue, please let us know.

Thank you.

### Reproduction steps

We are unable to reliably reproduce this issue since the issue is intermitten

### Affected SDK versions

2.5.0, 2.6.0

### Did you confirm if the latest SDK version fixes the bug?

No

### Flutter Version

3.19.6

fuzzybinary:
Hi @hakivin,

Do you know if this seems to happen more often on iOS or Android?

Also, if you haven't yet, can you please open a support issue with Datadog and point them to this issue? I'd like to take a look and see if I'm seeing any telemetry from your org that might point to the potential problem.

Thanks!

hakivin:
it is only happened in Android. 

I have just submitted a support ticket. Thank you.

fuzzybinary:
Hi @hakivin ,

Really quick, do you have any background services or multiple flutter engines being created on Android? I'm seeing one error I expected, but a second that I wasn't.

hakivin:
I don't think our apps are creating multiple Flutter engines or running any background services. May I know about the error you're seeing? I suspect it might be related to a background process as well.

hakivin:
[...]

PR #653, opened 2024-10-10
fix: Add additional timeout for mappers

fuzzybinary:
### What and why?

We're getting reports that event mappers still time out, even with 1s of available time. This is likely because of a sudden influx of calls on the method channel, which is hopefully temporary and can be resolved within the new timeout (5s).
 
Additionally, we're seeing reports of `notImplemented` errors when calling mappers.  This appears to be caused by users or third party libraries creating multiple Flutter engines (some of which don't initialize Datadog) or some weird (and very difficult to reproduce cases) where it appears that Flutter does not unregister the plugin.   To fix this, we'll attempt to call other mappers if the first one returns `NotImplemented`, and remove the offending method channel from the list.

Add formerly missing mapper "not implemented" telemetry to iOS.

Add an integration test for user information.

refs: RUM-6457 #652

### Review checklist

### DIFF
--- a/packages/datadog_common_test/lib/src/decoders/rum_decoder.dart
+++ b/packages/datadog_common_test/lib/src/decoders/rum_decoder.dart
@@ -6,6 +6,21 @@ import 'dart:io';
 
 import '../../datadog_common_test.dart';
 
+class RumUser {
+  String? email;
+  String? id;
+  String? name;
+
+  RumUser(this.email, this.id, this.name);
+
+  static RumUser fromJson(Map<String, dynamic> json) {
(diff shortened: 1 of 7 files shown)


packages/datadog_common_test/pubspec.yaml
0xnm, line 1: seems a change to rollback

0xnm (dismissed):

packages/datadog_common_test/pubspec.yaml
fuzzybinary: Yeup, fixed.

0xnm (approved):

PR #655, opened 2024-10-17
feat: Add `addViewLoadingTime` API

fuzzybinary:
### What and why?

Add methods to call `addViewLoadingTime` on the native SDKs.

refs: RUM-5841

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/ddrum_web.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/ddrum_web.dart
@@ -114,6 +114,11 @@ class DdRumWeb extends DdRumPlatform {
   }
 
   @override
+  Future<void> addViewLoadingTime(bool overwrite) async {
+    // NOOP - Not supported by the Browser SDK
+  }
+
+  @override
   Future<void> addAction(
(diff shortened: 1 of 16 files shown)


packages/datadog_flutter_plugin/lib/src/rum/ddrum_web.dart
```diff
@@ -113,6 +113,11 @@ class DdRumWeb extends DdRumPlatform {
     DD_RUM?.addTiming(name);
   }
 
+  @override
+  Future<void> addViewLoadingTime(bool overwrite) async {
+    // NOOP - Not supported by the Browser SDK
```
0xnm, line 118: maybe it makes sense to make a log entry in such cases?

0xnm (dismissed):

packages/datadog_flutter_plugin/lib/src/rum/ddrum_web.dart
fuzzybinary: The support matrix for web is pretty shallow, if I added a log here I'd be adding it for a lot of features that we don't support.

I think once more things are supported out of the box, I'll revisit and add more logs to things we don't support.

0xnm (approved):

PR #661, opened 2024-11-05
fix max int generation on web to use `-1 >> 0`

a-h-mzd:
### What and why?

When running flutter apps compiled to JS, `1 << 32` results in _0_ instead of the max int allowed for random number generation.

### How?

According to [this](https://dart.dev/guides/language/numbers#bitwise-operations) we can use `-1 >> 0` to achieve what is needed.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
+++ b/packages/datadog_flutter_plugin/lib/src/tracing/tracing_headers.dart
@@ -145,6 +145,8 @@ class TracingId {
 
   TracingId.spanId() : value = _generateSpanId();
 
+  static const _maxInt = kIsWeb ? -1 >> 0 : 1 << 32;
+
   /// Generate a 128-bit Trace Id.
   ///
   /// The trace is generated within the range:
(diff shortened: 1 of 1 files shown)


fuzzybinary:
Hi @a-h-mzd ,

Thanks for taking the time to create a PR.

Are you seeing something behave incorrectly? I'm pretty sure this tracer is actually only used in native platforms, not in Web where we rely on the Datadog Browser SDK's tracing implementation.

a-h-mzd:
@fuzzybinary thank you for the quick response.

`TracingContext.generateTracingContext` is actually used by the plugins(like _datadog_gql_link_ and _datadog_grpc_interceptor_). I have also created a new package which contains an interceptor for **Dio** based on the implementation found in _datadog_gql_link_. I will create a PR for it soon.

fuzzybinary:
👍 I'll still have to double check that the Browser SDK doesn't handle those calls internally anyway. It would be better if we had a failing test or repro case to work from to start.

a-h-mzd:
It’s actually pretty easy to test. Just try to print `1 << 32` and run the code in the browser or just compile it to JS. You’ll see that `1 << 32` is converted to _0_.

fuzzybinary:
Yes, I'm more referring to showing that using the `grpc` or `gql` interceptors on web does not produce the desired results. The actual feature is that distributed tracing in these packages works on web. `1 << 32` is an implementation detail and the cause of the bug.

a-h-mzd:
A similar commit(054b038ec0586ff93dc8b94a9d2241c597b46072) to this PR has landed in the main branch and therefor this PR is no longer needed.

PR #662, opened 2024-11-05
feat: Initial release of datadog_dio_interceptor plugin

a-h-mzd:
### What and why?

Currently there are no Dio interceptors in the repo. 

### How?

This PR aims to introduce a Dio interceptor to make it easier for anyone using the Dio package to attach Datadog-specific headers to the requests.

### Review checklist

### DIFF
--- a/packages/datadog_dio_interceptor/.gitignore
+++ b/packages/datadog_dio_interceptor/.gitignore
@@ -1,0 +1,29 @@
+# Miscellaneous
+*.class
+*.log
+*.pyc
+*.swp
+.DS_Store
+.atom/
+.buildlog/
+.history
+.svn/
+
+# IntelliJ related
+*.iml
+*.ipr
+*.iws
+.idea/
+
+# The .vscode folder contains launch configuration and tasks you configure in
+# VS Code which you may wish to be included in version control, so this line
+# is commented out by default.
+#.vscode/
+
+# Flutter/Dart/Pub related
+# Libraries should not include pubspec.lock, per https://dart.dev/guides/libraries/private-files#pubspeclock.
+/pubspec.lock
+**/doc/api/
+.dart_tool/
+.packages
+build/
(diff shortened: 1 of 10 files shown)


a-h-mzd:
@fuzzybinary, currently there are no tests implemented for this package. I plan to add the tests soon.

rtrieu:
Created DOCS-9428 for docs review.

fuzzybinary:
@rtrieu Hold off on docs review until I can take a look at the code.

fuzzybinary:
Hi @a-h-mzd,

We'll need tests before we can accept, but, also, is there a reason this was needed? All cases I've seen of Dio either use the default `HttpClient` provided with `HttpOverrides`, or accept a `Client` from the `http` package.  Our interceptor provides ways to override both.

Is there a case I don't know of where Dio will not use something we already intercept, or is there additional functionality this would provide?

a-h-mzd:
Hi @fuzzybinary,

You're partially correct. Dio for native uses the default client provided by the overrides, but Dio for browser takes a different approach and uses `XMLHttpRequest` directly to make the requests. As far as I know, this is done to ensure that the UI thread does not freeze, no matter how heavy and time-consuming the API call is.

And as for tests, I will try to write some as soon as I get the chance.

fuzzybinary:
@a-h-mzd Did you double check that the Browser SDK doesn't already intercept these without going through any Dart / Flutter code? If I remember correctly, it intercepts all XMLHttpRequests by default (and gets more information out of them than we do). If it doesn't that may be the real bug, not the lack of a Dio interceptor.

I appreciate the contributions, but I would always start with an issue in the repo first. Show what you want to do and the fact that it doesn't work, then we can work together to determine what the solution should be (which, in reality, may be this interceptor provided your issue checks out).

a-h-mzd:
@fuzzybinary that is a good point. I will create an issue right away.

a-h-mzd:
@fuzzybinary thank you for your support on [this issue](https://github.com/DataDog/dd-sdk-flutter/issues/663). Do you think it’s worth to keep this PR open?

fuzzybinary:
No I think it's better to close it. Our support for Dio flows through HttpOverrides at the moment.

Issue #663, opened 2024-11-05
Datadog headers not added to requests made with Dio on web


a-h-mzd:
### Question

When following the configuration steps in the documentation, we are instructed to use the enableHttpTracking function from the datadog_tracking_http_client package to track HTTP requests and add the necessary Datadog tracking headers. Based on my review of the code, it appears that this function overrides HttpOverrides.global with a custom implementation. I can confirm that requests are indeed being tracked, as they appear in the Datadog portal. However, Datadog-related headers are not being attached to these requests.

Please note that this issue is specific to web, and I haven’t tested it on other platforms. I suspect this approach might work on other platforms, as dio for native uses HttpOverrides to create the client for making requests. However, for browser environments, dio uses XMLHttpRequest directly for requests, which could be the reason headers aren't being attached.

fuzzybinary:
Hi @a-h-mzd, thanks for the issue!

If requests are appearing in the portal, than the Browser SDK is intercepting them properly. Can you double check that you've properly setup `firstPartyHosts` properly, as this is the configuration option that tells Datadog which requests to attach tracing headers to.

a-h-mzd:
It's mentioned in the documentation that `firstPartyHosts` must look like this:
```dart
final configuration = DatadogConfiguration(
  // configuration
  firstPartyHosts: ['example.com'],
)..enableHttpTracking()
```
And that is exactly how my current setup looks like.

mathieux51:
@fuzzybinary for context [here](https://help.datadoghq.com/hc/en-us/requests/1889210) is the request that was made to support.

fuzzybinary:
Hi @mathieux51 ,

Your support request had all the right information, thanks! It turns out there are a few issue in distributed tracing in Flutter Web, but Dio has nothing to do with it. 

First, the regex we supply to Web is made to look for the host name only, but the Browser SDK checks against the full URL.
Second, we never send the `traceSampleRate` option to Web.

I'm working on those issues now and adding a set of integration tests for tracing on Flutter web.

mathieux51:
@fuzzybinary amazing news 🎉 . I'm very happy you're working on this! Thanks for your help 🙏

a-h-mzd:
@fuzzybinary, I just tested #664 using our project, and it is working perfectly fine now. The traces are appearing in the portal as expected. Thank you for the fast and great support!

a-h-mzd:
[...]

PR #664, opened 2024-11-07
fix: Fix distributed tracing for web

fuzzybinary:
### What and why?

Web was not properly adding the trace sample rate, nor was it properly checking the `firstPartyHost` against the requested URL correctly. Both of these are fixed.

As part of this, I added web as a platform on the tracing package.  Web doesn't actually use our http tracking module, but the tests for http tracking live in that module.

Additionally, add sanity tests for the Dio, which we assume works because it uses HttpOverrides on mobile (which our plugin intercepts) and XMLHttpRequest on web (which the Browser SDK intercepts).  Adding an integration test to ensure it's working on all supported platforms and that we don't break it moving forward.

refs: #663

### Review checklist


a-h-mzd:
Hey @fuzzybinary,
Do you have an ETA on this?

0xnm (approved):

fuzzybinary:
@a-h-mzd There are a few things I'm looking into getting in before we do our next release, but hopefully soon.

Issue #667, opened 2024-11-12
Add ability to add custom attributes to RumUserActionAnnotation


orevial:
### Feature description

Currently Datadog default implementation uses `RumUserActionDetector` to detect buttons & other actions taps.
Also it gives `RumUserActionAnnotation` which allow providing custom description, to either override (for things like standard buttons or Semantics widgets) or implement tap-like strategies for custom elements.

When using `RumUserActionAnnotation` it would be great if we could also add an `attributes` map to this widget so we can further customize the tap action with custom attributes, just like we can do with `DatadogSdk.instance.rum?.addAction()`.

Say I have an avatar image of a user, and the tap on this image will redirect to the user profile. 

Currently what I can do is pass the user id and/or name as the action description, e.g.: 

```dart
RumUserActionAnnotation(
  description: 'Avatar(user 123)'
  child: UserAvatar(),
)
```

However this is not ideal because it means that each avatar tap will result in a different action in Datadog. 
What I'd like to do instead is use a generic name as the description, and then use custom attributes to add additional data (e.g. for debugging purposes). Something along those lines: 

```dart
RumUserActionAnnotation(
  description: 'Avatar(user)'
  attributes: {
    'userId': '123',
    'userName': 'John Doe',
  },
  child: UserAvatar(),
)
```

### Proposed solution

```dart
@immutable
class RumUserActionAnnotation extends StatelessWidget {
  final String description;
  final Widget child;
  final Map<String, Object?> attributes;

  const RumUserActionAnnotation({
    super.key,
    required this.description,
    required this.child,
    this.attributes = const {},
  });

  @override
  Widget build(BuildContext context) {
    return child;
  }
}
```

Then in `RumUserActionDetector` we should check if attributes exist, and if they do we should give it to `addAction` method: 

```dart
  void _onPerformActionAt(Offset position, RumActionType action) {
    final elementDescription = _getDetectingElementAtPosition(position);

    if (elementDescription != null) {
      widget.rum?.addAction(
         action, 
         elementDescription.toString(),
         // Also give the optional attributes
         attributes,
);
    }
  }
```

fuzzybinary:
@orevial This looks good to me. Would you be up for implementing / submitting a PR?

orevial:
Yes I think so, I will try to submit a PR in the next few days 🙂

orevial:
[...]

PR #669, opened 2024-11-13
feat: add ability to add custom attributes to RumUserActionAnnotation

orevial:
Context can be found in issue #667

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
@@ -20,6 +20,7 @@ class _ElementDescription {
   final Element element;
   final String elementName;
   final String elementDescription;
+  final Map<String, Object?>? attributes;
 
   // Whether we can potentially do better than this element further down the
   // tree. Used for non-specific widgets like `GestureDetector` and `InkWell`
@@ -47,6 +48,7 @@ class _ElementDescription {
     required this.elementDescription,
     // ignore: unused_element
     this.tryForBetter = false,
+    this.attributes,
   });
 }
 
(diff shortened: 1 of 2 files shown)


packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
```diff
@@ -148,16 +148,19 @@ void main() {
     verifyNoMoreInteractions(mockRum);
   });
 
-  testWidgets('tap gesture detector with annotation reports description',
+  testWidgets(
+      'tap gesture detector with annotation reports description and attributes',
```
fuzzybinary: Let's leave the existing tests alone since the basic version of this API remains unchanged. Instead, let's add a new tests that checks that when attributes are added they're forwarded properly.

fuzzybinary (changes requested):
Thanks @orevial! This looks good.

In addition to the minor test changes, can you [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)?  Once that's setup, I'll have to reopen the PR so our build system can see it properly, but then we should be able to get this merged in.

packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
```diff
@@ -405,15 +412,25 @@ Element? _findGestureDetectorElement(
 class RumUserActionAnnotation extends StatelessWidget {
   final String description;
   final Widget child;
+  final Map<String, Object?>? attributes;
```
fuzzybinary, line 418: One more comment -- you did such a good job with the doc strings on your other PR it made me realize we should add some documentation for this as well. You should be good to just add it to the class's doc string.

orevial:
@fuzzybinary commit signed and request changes made, let me know if you need me to reopen the PR or something else !

fuzzybinary:
Looks good. I've reopend as #674 for us to merge in once CI checks pass.

Thanks for the contribution!

PR #671, opened 2024-11-14
feat: add ability to add custom gesture detectors to RumUserActionDetector

orevial:
This PR addresses the problem raised in [this comment](https://github.com/DataDog/dd-sdk-flutter/issues/667#issuecomment-2473206701) from issue #667.

Implemented solution is the one proposed by @fuzzybinary in [this comment](https://github.com/DataDog/dd-sdk-flutter/issues/667#issuecomment-2473829225).

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
@@ -59,6 +59,17 @@ class _ElementDescription {
 /// Flutter widgets, including [ElevatedButton], [TextButton],
 /// [CupertinoButton], [BottomNavigationBar], [TabBar], [InkWell], and
 /// [GestureDetector].
+/// You can also provide a custom detection logic by passing [customGestureDetector]
+/// parameter to detect your custom tappable widgets. Example:
+///
+/// ```dart
+/// customGestureDetector: (widget) {
+///    if (widget is CustomButton) {
+///      return RumGestureDetectorInfo('CustomButton');
+///    }
+///    return null;
+/// }
+/// ```
 ///
 /// For most Button types, the detector will look for a [Text] widget child,
 /// which it will use for the description of the action. In other cases, it will
@@ -79,10 +90,14 @@ class RumUserActionDetector extends StatefulWidget {
   /// The Widget tree to detect gestures in.
   final Widget child;
 
+  /// A function that provides custom gesture detection logic.
+  final CustomGestureElementDetector? customGestureDetector;
+
   const RumUserActionDetector({
     super.key,
     required this.rum,
     required this.child,
+    this.customGestureDetector,
   });
 
   @override
@@ -290,7 +305,14 @@ class _RumUserActionDetectorState extends State<RumUserActionDetector> {
     String? elementName;
     bool searchForBetter = false;
     bool searchForText = true;
-    if (widget is ButtonStyleButton) {
+
+    final customWidgetDetected =
+        this.widget.customGestureDetector?.call(widget);
+    if (customWidgetDetected != null) {
+      elementName = customWidgetDetected.elementName;
+      searchForBetter = customWidgetDetected.searchForBetter;
+      searchForText = customWidgetDetected.searchForText;
+    } else if (widget is ButtonStyleButton) {
       if (widget.enabled) {
         elementName = 'Button';
       }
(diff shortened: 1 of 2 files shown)


packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
```diff
@@ -77,10 +77,13 @@ class RumUserActionDetector extends StatefulWidget {
   /// The Widget tree to detect gestures in.
   final Widget child;
 
+  final CustomGestureElementDetector? customGestureDetector;
```
fuzzybinary, line 94: Can we add a doc string to this?

packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
```diff
@@ -283,6 +286,14 @@ class _RumUserActionDetectorState extends State<RumUserActionDetector> {
...
     bool searchForText = true;
+
+    final customWidgetDetected =
+        this.widget.customGestureDetector?.call(widget);
+    if (customWidgetDetected != null) {
+      elementName = customWidgetDetected.elementName;
+      searchForBetter = customWidgetDetected.searchForBetter;
+      searchForText = customWidgetDetected.searchForText;
+    }
     if (widget is ButtonStyleButton) {
```
fuzzybinary: This should probably be `else if`, otherwise the custom widget information will be overwritten if the widget happens to also be a `ButonStyleButton`.

packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
```diff
@@ -417,3 +428,47 @@ class RumUserActionAnnotation extends StatelessWidget {
...
+}
+
+/// This function type allows you to define custom criteria for gesture detection
+/// within your widget tree. It accepts a `Widget` as input and returns
+/// a `RumGestureDetectorInfo` if the widget meets the specified criteria, or
+/// `null` otherwise.
+///
+/// This function is used by the [RumUserActionDetector] to detect custom gesture detectors.
+///
+/// Example use:
```
fuzzybinary, line 497: Good work on this documentation!

packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
```diff
@@ -514,4 +537,63 @@ void main() {
...
+      _TestWidget(),
+    ));
+
+    final button = find.byType(_TestWidget);
+    await tester.tap(button);
+
+    verify(() => mockRum.addAction(RumActionType.tap, 'TestWidget(unknown)'));
+  });
+
+  testWidgets('tap custom widget reports tap to RUM and search for better text',
```
fuzzybinary: I think we should also have tests for the `searchForBetter` and `searchForText` options work as expected.

fuzzybinary (changes requested):
Thanks @orevial this looks good except for a few minor changes and additional tests needed.

One interesting thing that's lacking is what "better" means in the context of `searchForBetter`. It might be worth adding to the documentation that in this case it means until a custom element returns `false` for `searchForBetter` or until the RumUserActionDetector locates one of its supported gesture detectors (that is not a GestureDetector or InkWell)

As with the previous PR, can you [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)?

packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
orevial: It would be a bit messed up to do custom detection logic on Material components, but I'll gladly change this to an `else if` 🙂

fuzzybinary:
Looks good. I've reopend as #674 for us to merge in once CI checks pass.

Thanks for the contribution!

packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
fuzzybinary: Agreed, but that doesn't mean a custom component couldn't inherit from a builtin component, in which case it would match both the custom logic and the checks below would overwrite it

orevial:
@fuzzybinary PR rebased, commit signed, doc and tests added

packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
```diff
@@ -540,4 +590,181 @@ void main() {
...
+
+      final text = find.byType(_DescriptiveWidget);
+      await tester.tap(text);
+
+      verify(() => mockRum.addAction(
+          RumActionType.tap, 'DescriptiveWidget($annotation)'));
+      verifyNoMoreInteractions(mockRum);
+    });
+
+    // testWidgets(
```
fuzzybinary: Did you forget to remove these?

orevial: Definitely did 😅

orevial: Fixed.

fuzzybinary (approved):

PR #679, opened 2024-11-19
feat: allow RumUserActionAnnotation in subtrees

orevial:
### What and why?

PR #671 added the ability to add custom gesture detectors which is great when dealing with reusable elements like design systems buttons or other reusable tappable elements. 
However when doing this we need a way to actually give a custom description to these custom widgets. 

Take this custom button for example: 

```dart
class CustomButton extends StatelessWidget {
  final String debugActionLabel;
  final Map<String, Object?>? attributes;
  final String? text;
  final VoidCallback? onPressed;

  const CustomButton({
    required this.debugActionLabel,
    this.attributes,
    this.text,
    this.onPressed,
    super.key,
  });

  @override
  Widget build(BuildContext context) {
    return RumUserActionAnnotation(
      description: debugActionLabel,
      attributes: attributes,
      child: ElevatedButton(
        onPressed: onPressed,
        child: text != null ? Text(text!) : null,
      ),
    );
  }
}
```

With this PR, the `CustomButton` can actually display `debugActionLabel` as its description instead of going further and find the text of the `ElevatedButton` instead.

I've added a complete example in the plugin example app.

### How?

When visiting children tree for "interesting" node, the algorithm now stops if it detects a `RumUserActionAnnotation` widget because it is considered the best (most descriptive) widget we could find.

@fuzzybinary let me know what you think about this, but without this commit the `customGestureDetector` implemented in PR #671 doesn't add much value...


packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
```diff
@@ -195,12 +195,17 @@ class _RumUserActionDetectorState extends State<RumUserActionDetector> {
...
   _RumTreeAnnotation? _findElementInnerText(Element element, bool allowText) {
     String? elementDescription;
+    Map<String, Object?>? attributes;
 
     void visitor(Element element) {
       bool stopVisits = false;
 
       var widget = element.widget;
-      if (allowText && widget is Text) {
+      if (allowText && widget is RumUserActionAnnotation) {
```
orevial: Not sure whether the `allowText` is needed here, maybe we should consider that a `RumUserActionAnnotation` widget should always "win" the search text as we consider it the most descriptive widget no matter what ?

fuzzybinary: No, the `allowText` shouldn't be needed. It's there to prevent accidentally pulling text from something that might have sensitive information, but since `RumUserActionAnnotation` isn't an automatic capture, users should be able to prevent having sensitive information on their own.

fuzzybinary (commented):
So, essentially what this is doing is allowing `RumUserActionAnnotation` both above the detected element and below it, overriding the text in both cases....

While I get the need I'll have to think of the design.  `RumUserActionAnnotation` is meant to modify the tree below it, and having it do the opposite seems counter intuitive.  What may be better is to add a property to RumGestureDetectorInfo for the description. In which case it might make sense to pass in any *current* `RumUserActtionAnnotation` for reference.

fuzzybinary:
Having thought about this @orevial here's my suggestion. Instead of searching for a `RumUserActionAnnotation` lower down the tree (which in my opinion violates how Flutter trees should really work), we should modify `RumGestureDetectorInfo` to include a member `String? elementDescription`.

If `elementDescription` is `null` and `searchForText` is true, it follows the current code. In the case where `elementDescription` is filled in, we'll use that and stop searching.

In your example, the user code would look like this:

```dart
RumGestureDetectorInfo? customDetector(Widget widget, RumUserActionAnnotation? annotation) {
  if (widget is CustomButton) {
    return RumGestureDetectorInfo('CustomButton', description: widget.debugActionLabel);
  }
  return null;
}

class CustomButton extends StatelessWidget {
  final String debugActionLabel;

  // ....

  @override
  Widget build(BuildContext context) {
    return ElevatedButton(
      onPressed: onPressed,
      child: text != null ? Text(text!) : null,
    );
  }
}
```

We could also include attributes in this way as well. But, I would say if you're supplying both a debug action label and attributes every time you make a `CustomButton`, a secondary wrapper might work better. E.g.:

```dart
RumGestureDetectorInfo? customDetector(Widget widget, RumUserActionAnnotation? annotation) {
  if (widget is _InnerCustomButton) {
    return RumGestureDetectorInfo('CustomButton');
  }
  return null;
}

class CustomButton extends StatelessWidget {
  final String debugActionLabel;
  final Map<String, Object?>? attributes;

  @override
  Widget build(BuildContext context) {
    return RumUserActionAnnotation(
      description: debugActionLabel,
      attributes: attributes,
      child: _InnerCustomButton(
        // etc..
      )
    );
  }  
}

class _InnerCustomButton extends StatelessWidget {
  // Etc.
}
```

 Let me know what you think.

fuzzybinary:
Another approach I just thought of. 

In `RumUserActionDetector` we have this code:

```dart
if (widget is RumUserActionAnnotation) {
    rumTreeAnnotation =
        _RumTreeAnnotation(widget.description, widget.attributes);
  } else {
    // get the detecting element
  }
}
```

By removing the else and making it read like this:

```dart
if (widget is RumUserActionAnnotation) {
    rumTreeAnnotation =
        _RumTreeAnnotation(widget.description, widget.attributes);
}

// Get the detecting element
```

You could have a custom `RumUserActionAnnotation` that is also sent as a custom widget to CustomGestureElementDetector.

orevial:
Thanks for the answers @fuzzybinary.

I like your suggestion about adding the debug label to the `RumGestureDetectorInfo`. I didn't do it in the first place because I thought that we wouldn't be able to add these labels at runtime when using the widgets, but I forgot that we could access the widgets properties doing `widget.debugActionLabels` 🙄 

However, I'm not sure I understand your proposition about a second, inner wrapper to add attributes. It seems to me that it could work easily using the same principle: 

```dart
RumGestureDetectorInfo? customDetector(Widget widget) {
  if (widget is CustomButton) {
    return RumGestureDetectorInfo('CustomButton', description: widget.debugActionLabel, attributes: widget.debugActionAttributes);
  }
  return null;
}

class CustomButton extends StatelessWidget {
  final String debugActionLabel;
  final Map<String, Object?>? attributes;

  // ....

  @override
  Widget build(BuildContext context) {
    return ElevatedButton(
      onPressed: onPressed,
      child: text != null ? Text(text!) : null,
    );
  }
}
```

What do you think ? I don't really see how a wrapper widget make our life easier here...

About your last proposition of removing the else, I'm not sure I understand what that would give 🤔

fuzzybinary:
> What do you think ? I don't really see how a wrapper widget make our life easier here...

It just depends on where those attributes come from, that's all. They're determined by the parent when the `CustomButton` is created, there's no large benefit, but if portions are created by `CustomButton` itself, then wrapping makes the use of CustomButton cleaner and hides the attribute creation from the user.

The second proposal doesn't work. The idea was to be able to have something like this:

```dart
class CustomButton extends StatelessWidget implements RumUserActionAnnotation {
  // Code
}
```

But now that I've tried it, it's not a great approach.  Although, it does make me think having an interface or mixin for annotation of custom widgets would be a good alternative.

fuzzybinary:
@orevial Rereading this PR, this actually is what I was suggesting for #687, which is just allowing RumUserActionAnnotation to override the text as a child, which I'm fine with now that I'm re-thinking about it.

I'll put this through our CI but I'm going to push a few changes to add tests and some logic to make sure it actually fixes #687 before merging.

plousada (approved):

Issue #708, opened 2025-02-13
iOS compilation error with Flutter 3.29.0


Laennart:
### Describe the issue

With Flutter's latest release 3.29.0, the compilation fails for iOS with the following error message:

```
Building com.example.datadogTest for device (ios)...
Running pod install...                                           1,547ms
Running Xcode build...
Xcode build done.                                           13.2s
Failed to build iOS app
Swift Compiler Error (Xcode): Cannot use optional chaining on non-optional value of type 'FlutterEngine'
/Users/X/.pub-cache/hosted/pub.dev/datadog_flutter_plugin-2.10.0/ios/Classes/DatadogSdkPlugin.swift:346:25
```

### Reproduction steps

```bash
flutter create datadog_test
cd datadog_test
flutter pub add datadog_flutter_plugin
flutter build ios --debug --no-codesign
```

### Device Information

Apple MacBook Pro, M1 Pro
macOS 15.3.1

### SDK version

2.10.0

### Flutter Version

3.29.0

### Setup Type

Flutter Application

milodude:
You can manually edit this file in order to continue working until the issue is fixed. 
Open the terminal and run
1- xed "/Users/X/.pub-cache/hosted/pub.dev/datadog_flutter_plugin-2.10.0/ios/Classes/DatadogSdkPlugin.swift"
This will open the file using xcode. After that edit the line it points out. 

This will help you compile the app. If you do not want to do this, you will need to download the previous version of flutter.

vixez:
Have this issue as well.
I build using Codemagic with the latest Flutter version, so I can't manually change pub-cache files.

fuzzybinary:
Thanks for reporting @Laennart and thanks for the PR. I'll take a look and try to get this into a point release.

fuzzybinary:
This is now available in 2.10.1

PR #709, opened 2025-02-13
fix: iOS compilation error

Laennart:
### What and why?

This pull request fixes the compilation issue described in #708. 

### How?

It seems that in the latest Flutter release 3.29.0, the `engine` parameter on the `FlutterViewController` returns a non-optional type instead of an option type which causes compilation issues when upgrading the Flutter version. By casting the engine to an optional type, the compilation issue is fixed with Flutter 3.29.0 but also remains compatible with older versions of Flutter.

### Review checklist


fuzzybinary:
I've reopened this as #710 because it appears the Flutter update also included a stricter analyzer, which requires we make some changes in unrelated files.

Issue #716, opened 2025-02-24
[datadog_tracking_http_client] rumKey null exception when shouldTrack is false


4rthurmonteiro:
### Describe the bug

When shouldTrack is false a null exception is thrown because rumKey is null.

![Image](https://github.com/user-attachments/assets/0c3a110c-b33d-4cf5-93d9-de2dc234ae60)

### Reproduction steps

1. Add some pattern in `ignoreUrlPatterns` when is enabling the http tracking;
2. Go to the content with the url added before.

### SDK logs

![Image](https://github.com/user-attachments/assets/8b72456b-1bca-4296-8fe2-ae025b7245c4)

### Expected behavior

The SDK should not thrown an exception.

### Affected SDK versions

2.2.0

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Flutter Version

3.22.2

### Setup Type

Hybrid App

fuzzybinary:
Hi @4rthurmonteiro ,

Thanks for reporting. I see the issue and also why others may not have experienced it. From what I can tell, it requires you have both and `ignoreUrlPattern` and a `clientListener` set up.

I'll try to get a patch fix out for you ASAP.

4rthurmonteiro:
> Hi [@4rthurmonteiro](https://github.com/4rthurmonteiro) ,
> 
> Thanks for reporting. I see the issue and also why others may not have experienced it. From what I can tell, it requires you have both and `ignoreUrlPattern` and a `clientListener` set up.
> 
> I'll try to get a patch fix out for you ASAP.

thanks @fuzzybinary !!

fuzzybinary:
Fixed and available now in `datadog_tracking_http_client` 2.3.0

arthurmonteiroalvesmelo:
Great @fuzzybinary ! I tested and now is working perfectly! Congrats

PR #717, opened 2025-02-24
fix: Untracked URLs could throw an exception with a clientListener

fuzzybinary:
### What and why?

Calls into the `clientListener` expected all calls to be tracked, and didn't correctly account for untracked URLs not having a RUM Key.  These ignored urls shouldn't call into the client listener at all.

Fix the null reference exception and add a test to verify.

refs: #716

### Review checklist


plousada (approved):

4rthurmonteiro:
@fuzzybinary when is the next release?

fuzzybinary:
@4rthurmonteiro Branches for 2.3.0 have already been cut and baring any unforeseen we should have it out later today.

PR #737, opened 2025-04-04
chore: Automatically update NATIVE_SDK_VERSIONS.md on package release 

sbarrio:
### What and why?

This PR introduces a new function to version_updater.dart that automatically creates/updates NATIVE_SDK_VERSIONS.md with the current triad of versions (Flutter SDK, iOS SDK and Android SDK) whenever we release a new version of the Flutter plugin.

We need a simple and public way to see which versions of the native SDKs are being used with each version of the Flutter plugin. This list should also be automatically generated and maintained.

refs: RUM-6122

### How?

This PR leverages the existing version_updater.dart Command and checks if a NATIVE_SDK_VERSIONS.md already exists and updates it with the latest version triad or creates it and populates it if it didn't.

### Review checklist


tools/releaser/lib/version_updater.dart
```diff
@@ -157,3 +161,33 @@ Future<bool> _updateReadmeVersions(CommandArguments args, Logger logger) async {
...
+  await transformFile(nativeSDKVersionsFile, logger, args.dryRun, (line) {
+    if (line.startsWith('|-')) {
+      return'''$separator\n$newVersionEntry''';
+    }
+
+    return line;
+  });
+
+  return true;
+}
```
fuzzybinary: Missing the extra EoL here likely means you also didn't use `dart-format`.  If you can, I would set up your IDE to format on save. Dart is incredibly opinionated about formatting, but thankfully the formatter handles all of it for you.

tools/releaser/lib/version_updater.dart
```diff
@@ -157,3 +161,33 @@ Future<bool> _updateReadmeVersions(CommandArguments args, Logger logger) async {
...
+
+Future<bool> _updateNativeSDKVersions(CommandArguments args, Logger logger) async {
+  final nativeSDKVersionsFile = File(path.join(args.packageRoot, 'NATIVE_SDK_VERSIONS.md'));
+  final newVersionEntry ='| ${args.version} | ${args.iOSRelease} | ${args.androidRelease} |';
+  final header = '| Flutter | iOS SDK | Android SDK |';
+  final separator = '|---------|---------|-------------|';
+
+  if (!nativeSDKVersionsFile.existsSync()) {
+    logger.warning('⚠️ NATIVE_SDK_VERSIONS.md does not exist, creating it now.');
+    await nativeSDKVersionsFile.writeAsString('''$header\n$separator\n$newVersionEntry''');
```
fuzzybinary: This will be handled by dart-format but the triple quotes here aren't needed.

tools/releaser/lib/version_updater.dart
```diff
@@ -157,3 +161,33 @@ Future<bool> _updateReadmeVersions(CommandArguments args, Logger logger) async {
...
+    return true;
+  }
+
+  final content = await nativeSDKVersionsFile.readAsString();
+
+  if (content.contains(newVersionEntry)) {
```
fuzzybinary: I don't think this is right. It will search for any version matching, not just the Flutter versions. It would fail when we go to ship Flutter SDK 2.11.0 for example because iOS SDK 2.11.0 is in the table.

fuzzybinary (changes requested):
Looks okay, but need to fix the existing version search and make sure everything is formatted with the dart formatter.

tools/releaser/lib/version_updater.dart
sbarrio: My bad 😞 

Fixed at [0fa8f1c](https://github.com/DataDog/dd-sdk-flutter/pull/737/commits/0fa8f1c2dcc0659dcc1bd0c4a54d31d240d5c598)

tools/releaser/lib/version_updater.dart
sbarrio: Thanks for raising, changes at  [0fa8f1c](https://github.com/DataDog/dd-sdk-flutter/pull/737/commits/0fa8f1c2dcc0659dcc1bd0c4a54d31d240d5c598)

tools/releaser/lib/version_updater.dart
sbarrio: I agree, it is too specific and it only looks for a line containing all three versions together (as it uses  `newVersionEntry = '| ${args.version} | ${args.iOSRelease} | ${args.androidRelease} |';`). 

I have modified it so instead it looks for the Flutter SDK version across the whole file being present on the first column of the .md file.

changes at [0fa8f1c](https://github.com/DataDog/dd-sdk-flutter/pull/737/commits/0fa8f1c2dcc0659dcc1bd0c4a54d31d240d5c598)

fuzzybinary (approved):

Issue #738, opened 2025-04-07
Reason behind `betterThan` function


a-h-mzd:
### Question

I was trying to attach different information to my nested GestureDetectors but it is currently not possible due to `betterThan` returning false for gesture detectors. The comment for this function mentions "Literally anything is better than GestureDetector" but what about nested gesture detectors? Or when I have a gesture detector as a child of an InkWell?

This is what I'm trying to do:
```dart
final class TestWidget extends StatelessWidget {
  const TestWidget({super.key});

  @override
  Widget build(BuildContext context) {
    return RumUserActionAnnotation(
      description: 'TestWidget',
      attributes: {'placement': 'parent'},
      child: GestureDetector(
        onTap: () => 'parent tapped',
        child: Container(
          width: 100,
          height: 100,
          color: Colors.white,
          alignment: Alignment.center,
          child: Column(
            mainAxisSize: MainAxisSize.min,
            children: [
              Text('parent'),
              RumUserActionAnnotation(
                description: 'TestWidget - Child',
                attributes: {'placement': 'child'},
                child: GestureDetector(
                  onTap: () => 'child tapped',
                  child: Text('child'),
                ),
              ),
            ],
          ),
        ),
      ),
    );
  }
}
```

fuzzybinary:
Hi @a-h-mzd ,

The purpose of `betterThan` is to try to see if another detector lower down the tree is a more specific target than one we've already found and would likely win in the gesture arena.

But, you've found a bug, which is that if both detectors are GestureDetectors, the one further down the tree should win, provided it has an `onTap` handler.

I'll try to get it solved in the next version. Sorry for the inconvenience!

PR #739, opened 2025-04-08
fix: Fix an issue where nested gesture detectors would always use the topmost.

fuzzybinary:
### What and why?

`betterThan` always assumed that the currently held element was better if it was comparing to a `GestureDetector`, but this is incorrect if it was comparing itself to another `GestureDetector`. In that case, the detector lower down the tree should take precedence.

refs: #738

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
@@ -34,9 +34,15 @@ class _ElementDescription {
   }
 
   bool betterThan(_ElementDescription? other) {
-    // Literally anything is better than GestureDetector
-    if (element.widget is GestureDetector && other != null) {
-      return false;
+    if (other == null) {
+      // Something is always better than nothing
+      return true;
+    }
+
+    // Literally anything is better than GestureDetector...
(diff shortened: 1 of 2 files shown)


0xnm (approved):

a-h-mzd:
If the Gesture detector is placed inside an InkWell it should also take precedence. Does this PR address this case as well?

fuzzybinary:
@a-h-mzd No it doesn't change that. As currently specified, InkWell takes precedence over GestureDetector wherever it appears. Changing that would be a separate issue.

a-h-mzd:
> @a-h-mzd No it doesn't change that. As currently specified, InkWell takes precedence over GestureDetector wherever it appears. Changing that would be a separate issue.

@fuzzybinary Thank you for the clarification. My original issue already mentions the InkWell issue as well. Do I need to create a separate issue for this?

fuzzybinary:
@a-h-mzd Ah, sorry I missed that in the original ticket. Yes, create a separate issue with an example. I just want to make sure we have a more detailed discussion about expected behavior, and how it relates to other gesture detectors, before I implement it. The GestureDetector override was obvious, but I'm not sure the InkWell one is.

PR #749, opened 2025-04-30
feat: Add iOS overhead for SessionReplayFeature

fuzzybinary:
### What and why?

This adds a Feature that will pass through Flutter generated Session Replay Records into SRSegments. The iOS Feature and RequestBuilder take care of serializing data to disk, assembling and compressing the SRSegments to send to intake.

The feature also takes care of broadcasting RUM context changes back to Flutter.

refs: RUM-9551

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/ios/Classes/DatadogSessionReplayPlugin.swift
+++ b/packages/datadog_session_replay/ios/Classes/DatadogSessionReplayPlugin.swift
@@ -5,13 +5,108 @@ import Flutter
 import UIKit
 
 public class DatadogSessionReplayPlugin: NSObject, FlutterPlugin {
-  public static func register(with registrar: FlutterPluginRegistrar) {
-    let channel = FlutterMethodChannel(name: "datadog_session_replay", binaryMessenger: registrar.messenger())
-    let instance = DatadogSessionReplayPlugin()
-    registrar.addMethodCallDelegate(instance, channel: channel)
-  }
+    public static func register(with registrar: FlutterPluginRegistrar) {
+        let channel = FlutterMethodChannel(name: "datadog_session_replay", binaryMessenger: registrar.messenger())
+        let instance = DatadogSessionReplayPlugin(channel: channel)
+        registrar.addMethodCallDelegate(instance, channel: channel)
+    }
 
-  public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
-    
-  }
+    let channel: FlutterMethodChannel
+    var feature: FlutterSessionReplayFeature?
+
+    init(channel: FlutterMethodChannel) {
+        self.channel = channel
+    }
+
+    public func handle(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
+        guard let arguments = call.arguments as? [String: Any?] else {
+            result(
+                FlutterError.invalidOperation(message: "No arguments in call to \(call.method)")
+            )
+            return
+        }
+
+        switch call.method {
+        case "enable":
+            enableSessionReplay(argumnets: arguments, result: result)
+        case "setHasReplay":
+            setHasReplay(arguments: arguments, result: result)
+        case "setRecordCount":
+            setRecordCount(arguments: arguments, result: result)
+        default:
+            result(FlutterMethodNotImplemented)
+        }
+    }
+
+    private func enableSessionReplay(argumnets: [String: Any?], result: @escaping FlutterResult) {
+        guard let configurationJson = argumnets["configuration"] as? [String: Any?],
+              let configuration = FlutterSessionReplay.Configuration(fromEncoded: configurationJson) else {
+            result(FlutterError.missingParameter(methodName: "enable"))
+            return
+        }
+
+        enable(configuration: configuration)
+
+        result(nil)
+    }
+
+    // Visible for testing
+    internal func enable(configuration: FlutterSessionReplay.Configuration) {
+        var configuration = configuration
+        configuration.onContextChanged = { context in
+            self.onContextChanged(rumContext: context)
+        }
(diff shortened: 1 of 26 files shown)


packages/datadog_session_replay/ios/Classes/DatadogSessionReplayPlugin.swift
```diff
@@ -5,13 +5,108 @@ import Flutter
...
+    private func setHasReplay(arguments: [String: Any?], result: @escaping FlutterResult) {
+        guard let hasReplay = arguments["hasReplay"] as? Bool else {
+            result(FlutterError.missingParameter(methodName: "setHasReplay"))
+            return
+        }
+
+        feature?.setHasReplay(hasReplay)
+
+        result(nil)
+
```
ncreated:
```suggestion
```

packages/datadog_session_replay/ios/Classes/Feature/Baggages.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    /// The key referencing a `Bool` value that indicates if replay is being recorded.
+    static let hasReplay = "sr_has_replay"
+
+    /// They key referencing a `[String: Int64]` dictionary of viewIDs and associated records count.
+    static let recordsCountByViewID = "sr_records_count_by_view_id"
+}
+
+/// The RUM context received from `DatadogCore`.
+internal struct RUMContext: Codable, Equatable {
```
ncreated, line 20: 💡 There's parallel work being done by @maxep around type-sharing through `DatadogInternal`. It covers RUM types (https://github.com/DataDog/dd-sdk-ios/pull/2276). You both may want to sync on how to land both changes seamlessly.

packages/datadog_session_replay/ios/Classes/Feature/FlutterSeasionReplayFeature.swift
```diff
@@ -0,0 +1,54 @@
...
+import Foundation
+import DatadogInternal
+
+class FlutterSessionReplayFeature: DatadogRemoteFeature {
+    static var name: String = "session-replay"
+
+    let requestBuilder: DatadogInternal.FeatureRequestBuilder
+    let messageReceiver: DatadogInternal.FeatureMessageReceiver
+
+    private weak var core: DatadogCoreProtocol?
```
ncreated: convention/ It is discouraged to retain reference to `core` even if `weak`. It is best to retain `FeatureScope`:
```swift
let scope = core.scope(for: FlutterSessionReplayFeature.self)
```
The `FeatureScope` is already thread safe, non-optional and will be loaded lazily if it's requested before the feature is actually registered.

packages/datadog_session_replay/ios/Classes/Feature/RequestBuilder/RequestBuilder.swift
```diff
@@ -0,0 +1,163 @@
...
+            }
+            let segments = try wrappers.map { try $0.extractEnrichedRecord() }
+                .map { try SegmentJSON($0, source: source) }
+                .merge()
+
+            return try createRequest(segments: segments, context: context)
+        } catch let error {
+            print("Error \(error)")
+            throw error
+        }
```
ncreated: nit/ Is this `print()` intentional? Otherwise this block doesn't seem to do much.

ncreated (dismissed):
Looks good 👍. Please coordinate with @maxep to ensure there are no conflicts with the concurrent changes to DatadogInternal in dd-sdk-ios.

The code imported from `dd-sdk-ios` is appropriate for now 👍. However, for long-term maintainability, it might be worth exploring whether some of this logic can be reused via a separate library or through `SessionReplay._internal.*` extension helpers in native SR. Alternatively, we could use `DatadogInternal`, but it will lead to unnecessary coupling by introducing session replay-specific internals there.

packages/datadog_session_replay/ios/Classes/Feature/Baggages.swift
maxep: Yes, the `featureBaggages` in the core context are not yet touched, but they will be replaced by strongly typed definitions very soon. It will be a breaking change for this integration but an easy one. @fuzzybinary, I will ping you once done 👍

packages/datadog_session_replay/ios/Classes/Feature/FlutterSeasionReplayFeature.swift
fuzzybinary: I'm mostly using core to set baggages. Is there a way to do that only retaining FeatureScope? (other examples in dd-sdk-ios appear to hold the Core weak to do this)

packages/datadog_session_replay/ios/Classes/Feature/RequestBuilder/RequestBuilder.swift
fuzzybinary: No, left over from debugging I believe.

packages/datadog_session_replay/ios/Classes/Feature/FlutterSeasionReplayFeature.swift
ncreated: Yes, `set(baggage:)` should be available on feature scope as it extends [`BaggageSharing`](https://github.com/DataDog/dd-sdk-ios/blob/923de8d2c69f1548284fa161f92608d0c994bcfe/DatadogInternal/Sources/DatadogCoreProtocol.swift#L92-L121). RUM is the main example of using it:

https://github.com/DataDog/dd-sdk-ios/blob/923de8d2c69f1548284fa161f92608d0c994bcfe/DatadogRUM/Sources/RUMMonitor/Monitor.swift#L140-L142

other parts of the native SDK might be not yet migrated to this model.

ncreated (approved):

PR #754, opened 2025-05-27
feat: Support Flutter overhead for Session Replay

fuzzybinary:
### What and why?

Initial capture and processor overhead, which traverses the Element tree looking for widgets to record, then provides those to the processor for diffing and serialization.

refs: RUM-9553

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/lib/src/capture/capture_node.dart
+++ b/packages/datadog_session_replay/lib/src/capture/capture_node.dart
@@ -1,0 +1,38 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+import 'package:flutter/widgets.dart';
+
+import '../sr_data_models.dart';
+
+abstract class WireframeBuilder {
+  List<SRWireframe> buildWireframes(CaptureNode node);
+}
+
+@immutable
+class CapturedViewAttributes {
+  final Rect paintBounds;
+  final double scaleX;
+  final double scaleY;
+
+  int get x => paintBounds.left.round();
+  int get y => paintBounds.top.round();
+  int get width => paintBounds.width.round();
+  int get height => paintBounds.height.round();
+
+  const CapturedViewAttributes({
+    required this.paintBounds,
+    required this.scaleX,
+    required this.scaleY,
+  });
+}
+
+@immutable
+abstract class CaptureNode {
+  final CapturedViewAttributes attributes;
+
+  const CaptureNode(this.attributes);
+
+  List<SRWireframe> buildWireframes();
+}
--- a/packages/datadog_session_replay/lib/src/capture/element_recorders/container_recorder.dart
+++ b/packages/datadog_session_replay/lib/src/capture/element_recorders/container_recorder.dart
@@ -1,0 +1,115 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+import 'package:flutter/material.dart';
+
+import '../../extensions.dart';
+import '../../sr_data_models.dart';
+import '../capture_node.dart';
+import '../recorder.dart';
+import '../view_tree_snapshot.dart';
+
+class ContainerRecorder implements ElementRecorder {
+  final KeyGenerator keyGenerator;
+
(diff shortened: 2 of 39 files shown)


fuzzybinary:
This needs #755 to be merged before it can build properly.

packages/datadog_session_replay/lib/src/capture/capture_node.dart
```diff
@@ -0,0 +1,38 @@
...
+
+import '../sr_data_models.dart';
+
+abstract class WireframeBuilder {
+  List<SRWireframe> buildWireframes(CaptureNode node);
+}
+
+@immutable
+class CapturedViewAttributes {
+  final Rect paintBounds;
```
maciejburda, line 15: Reminds me of clipping.

Not sure how it works in Flutter regarding clips to bounds, but on iOS/Android it had to be manually propagated from parent to children because of flattening step.

https://github.com/DataDog/dd-sdk-ios/pull/2083

Just mentioning for a reference.

packages/datadog_session_replay/lib/src/capture/element_recorders/container_recorder.dart
```diff
@@ -0,0 +1,115 @@
...
+      final decoration = widget.decoration;
+      if (decoration is BoxDecoration) {
+        // TODO: TextDirection
+        cornerRadius = decoration.borderRadius?.resolve(null).topLeft.x;
+
+        // It is illegal to supply both Container.color and Decoration.color,
+        // so overwriting the background color here is okay.
+        backgroundColor = decoration.color;
+        if (decoration.border case final border?) {
+          // TODO: Look into non-uniform borders for SR
```
maciejburda, line 42: It's not supported on the player level. IMO we should ignore this for the time-being.

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -0,0 +1,196 @@
...
+abstract interface class ElementRecorder {
+  CaptureNodeSemantics? captureSemantics(
+    Element element,
+    CapturedViewAttributes attributes,
+  );
+}
+
+class KeyGenerator {
+  // This is close to JavaScript's MAX_SAFE_INT (53-bit)
+  static const int maxKey = 0x20000000000000;
```
maciejburda, line 25: I recall that we had issues when going beyond 32-bit

On iOS we have limited the max value of IDs to Int32
https://github.com/DataDog/dd-sdk-ios/blob/9a80e2eccab6f69999bf75ffb77b0ec52d8b0392/DatadogSessionReplay/Sources/Recorder/ViewTreeSnapshotProducer/ViewTreeSnapshot/NodeIDGenerator.swift#L30

On iOS, it's incremental ID generation, but I think it's fine to randomize as long as we can keep it consistent for the same view.

 On Android it's taking the memory address and trims it down to 32-bits.

packages/datadog_session_replay/lib/src/datadog_session_replay.dart
```diff
@@ -0,0 +1,72 @@
...
+    bool success = false;
+    await wrapAsync('enable', internalLogger, {}, () async {
+      success = await platform.enable(_configuration, _onContextChanged);
+    });
+
+    if (success) {
+      await _processor.start();
+
+      const timerDuration = Duration(milliseconds: 100);
+      // TODO(RUM-10155): See if we can be smarter about how often we perform tree captures
```
maciejburda, line 63: I'd just add - that's quite important to look for view update hook. Performance impact gets significantly improved.

Probably worth bumping priority of this exploration.

maciejburda (approved):
LGTM

Left some extra context that I recalled while reviewing.

packages/datadog_session_replay/lib/src/capture/element_recorders/container_recorder.dart
fuzzybinary: 👍 Still going to leave in the TODO though :)

packages/datadog_session_replay/lib/src/capture/recorder.dart
fuzzybinary: 👍 I'll pair it back down to 32-bits.

fuzzybinary: Actually, is it possible that it wasn't going beyond 32-bits that was the problem, and actually going beyond the 53-bits, as that is the maximum safe integer in JavaScript?

packages/datadog_session_replay/lib/src/datadog_session_replay.dart
fuzzybinary: 👍 I'll groom it higher.

PR #764, opened 2025-06-23
chore: Add golden / snapshot testing

fuzzybinary:
### What and why?

In order to ensure we're capturing Session Replay content correctly, we're capturing then rendering the resulting output to a canvas, which we can then compare to previous captures using Flutter's "golden" test framework.

Currently only capture and check the "simple" containers and text fields we currently capture, and text is rendered as blocks (mostly because this default way Flutter does golden testing).

An example golden file:
image

refs: RUM-9556

### How?

### Review checklist


packages/datadog_session_replay/golden_test/.gitignore
```diff
@@ -0,0 +1 @@
+failures/
```
0xnm:
```suggestion
failures/

```

0xnm (dismissed):

0xnm (approved):

PR #768, opened 2025-06-30
feat: Add support for ap2 datacenter

fuzzybinary:
### What and why?

Add support for the AP2 datacenter.

refs: RUM-10394

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
+++ b/packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
@@ -91,9 +91,13 @@ enum DatadogSite {
   /// [app.ddog-gov.com](https://app.ddog-gov.com/).
   us1Fed,
 
-  /// Asia based servers. Sends data to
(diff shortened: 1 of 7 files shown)


packages/datadog_flutter_plugin/lib/src/datadog_configuration.dart
```diff
@@ -94,6 +94,10 @@ enum DatadogSite {
   /// Asia based servers. Sends data to
   /// [ap1.datadoghq.com](https://ap1.datadoghq.com).
   ap1,
+
+  /// Asia based servers. Sends data to
```
0xnm: AP stands for Asia-Pacific and AP2 is Australia-based, so probably saying that it is Asia-based it not quite correct.

fuzzybinary: 👍 I'll update the comment on both.

0xnm (approved):

jonathanmos (approved):

PR #779, opened 2025-07-25
feat: Support spm over cocoapods

fuzzybinary:
### What and why?

As support for Cocoapods winds down, Flutter has added support for Cocoapods.  This requires a few modifications to our project, build process, and release process to support properly.

For now, the default build step on CI will use cocoapods, but I've added a step that enables SPM to test that build chain.

### Review checklist


packages/datadog_flutter_plugin/ios/.gitignore
```diff
@@ -32,6 +32,8 @@ xcuserdata
 *sync/
 Icon?
 .tags*
+.swiftpm/
+.build/
```
simaoseica-dd, line 37: You might also want to add `Package.resolved`.

melos.yaml
```diff
@@ -205,6 +205,18 @@ scripts:
...
       scope: 'datadog_flutter_plugin'
 
+  remove_cocoapods:
+    exec: |
+      cd example/ios
+      pod deintegrate
+      rm Podfile || true
+      rm Podfile.lock || true
+    description: |
+      Remove Cocoapods support from examples.  This prevents some build failures when testing
```
simaoseica-dd, line 215: question: Why not clean examples to avoid this dance? One example using cocoapods and another one using SPM.

simaoseica-dd: Also, this updates the `project.pbxproj`. And the next step might want to be in a point without git changes.

packages/datadog_flutter_plugin/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -1,10 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
    LastUpgradeVersion = "1510"
```
simaoseica-dd, line 3: Is it still using the Xcode 15?

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Package.swift
```diff
@@ -0,0 +1,31 @@
...
+    products: [
+        .library(name: "datadog-flutter-plugin", targets: ["datadog_flutter_plugin"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/Datadog/dd-sdk-ios.git", branch: "develop"),
+        .package(url: "https://github.com/almazrafi/DictionaryCoder.git", from: "1.2.0")
+    ],
+    targets: [
+        .target(
+            name: "datadog_flutter_plugin",
```
simaoseica-dd, line 20: Usually we have the library name equal to the target name. Did you found any issue to have this `-` `_` distinction?

packages/datadog_inappwebview_tracking/ios/datadog_inappwebview_tracking/Package.swift
```diff
@@ -0,0 +1,26 @@
...
+let package = Package(
+    name: "datadog_inappwebview_tracking",
+    platforms: [
+        .iOS("12.0")
+    ],
+    products: [
+        .library(name: "datadog-inappwebview-tracking", targets: ["datadog_inappwebview_tracking"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/Datadog/dd-sdk-ios.git", from: "2.20.0"),
```
simaoseica-dd, line 15: Why not the branch `develop`? `2.20.0` is too old

simaoseica-dd (commented):
Looks good but left some questions to discuss improvements

melos.yaml
fuzzybinary: Because there are some tests that rely on these examples, and I'm hoping that this tests that very few changes are needed to run using SPM over Cocoapods.

packages/datadog_flutter_plugin/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
fuzzybinary: I rarely open it so even though I'm using 16, it's still set to 15.

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Package.swift
fuzzybinary: No, but this is a recommendation from the [Flutter migration steps](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors). I think there's an incompatibility here where Flutter / Dart expects package names with underscores and Swift expects dashes.

packages/datadog_inappwebview_tracking/ios/datadog_inappwebview_tracking/Package.swift
fuzzybinary: This is a lose constraint on the lowest version that this package supports. From what I can tell, this will pull in anything above 2.20 but not anything 3+.

Pulling in the main package will pull in it's supported version, which will almost always be higher (in development, this is `develop` but when we ship we lock to a specific version of the package).

packages/datadog_flutter_plugin/ios/.gitignore
fuzzybinary: Doesn't that always go under `.swiftpm`? That's the only place I've seen it.

simaoseica-dd: It is under `*/swiftpm/`. They are very prone to change the hash, which will add a continuous noise on the PRs

simaoseica-dd: Also, there are Packages under `.spm.pods` that can be ignored.

simaoseica-dd (approved):
LGTM!

PR #787, opened 2025-07-31
feat: Support manual resource tracking on web

fuzzybinary:
### What and why?

Manual resource tracking uses the new `addEvent` functionality provided by the Browser SDK. This PR utilizes this to improve the functionality of several features, but mostly resource tracking. 

This also adds initial work on addAction, which correctly supports non-`custom` action types. However sustained actions that use `start/stopAction` and tracking of events that occur close to specific actions is still not supported.

refs: RUM-10042

### How?

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/web/ddrum_web.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/web/ddrum_web.dart
@@ -65,7 +65,8 @@ class DdRumWeb extends DdRumPlatform {
       trackFrustrations: rumConfiguration.trackFrustrations,
       trackLongTasks: rumConfiguration.detectLongTasks,
       enableExperimentalFeatures: ['feature_flags'.toJS].toJS,
-      compressIntakeRequests: true,
+      // TODO(RUM-11211): Support and document web configuration options.
+      compressIntakeRequests: false,
       plugins: plugins,
     ));
   }
--- a/packages/datadog_flutter_plugin/lib/src/rum/web/ddrum_web.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/web/ddrum_web.dart
@@ -6,14 +6,21 @@
 import 'dart:js_interop';
 
 import 'package:meta/meta.dart';
+import 'package:uuid/uuid.dart';
 
-import '../../datadog_flutter_plugin.dart';
-import '../../datadog_internal.dart';
-import '../logs/ddweb_helpers.dart';
-import '../web_helpers.dart';
-import 'ddrum_platform_interface.dart';
+import '../../../datadog_flutter_plugin.dart';
+import '../../../datadog_internal.dart';
+import '../../web_helpers.dart';
+import '../ddrum_platform_interface.dart';
+import 'raw_events.dart';
+import 'resource_tracker.dart';
+import 'rum_web_plugin.dart';
 
 class DdRumWeb extends DdRumPlatform {
+  final Uuid _uuid = Uuid();
+  RumWebPlugin? _webPlugin;
+  ResourceTracker? _resourceTracker;
+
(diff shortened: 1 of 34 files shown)


packages/datadog_flutter_plugin/lib/src/rum/web/ddrum_web.dart
```diff
@@ -48,10 +60,13 @@ class DdRumWeb extends DdRumPlatform {
       traceContextInjection:
           _contextInjectionString(rumConfiguration.traceContextInjection),
       trackViewsManually: true,
+      trackUserInteractions: false,
       trackResources: trackResources,
       trackFrustrations: rumConfiguration.trackFrustrations,
       trackLongTasks: rumConfiguration.detectLongTasks,
       enableExperimentalFeatures: ['feature_flags'.toJS].toJS,
+      compressIntakeRequests: true,
```
thomas-lebeau: **🔨 warning:** ‏ I don't know how these pages are served in the context of flutter, but this option needs some CORS headers to be configured properly to work. Otherwise, the SDK will stop completely.
see: https://docs.datadoghq.com/integrations/content_security_policy_logs/?tab=firefox#web-worker

thomas-lebeau (commented):
From what I can understand about the addEvent usage, this looks good 👍

packages/datadog_flutter_plugin/lib/src/rum/web/ddrum_web.dart
fuzzybinary: Interesting, thanks, this might explain some issues we've been having. I'll set this to false until we can add it as an option with proper documentation.

joepeeples (dismissed):

packages/datadog_flutter_plugin/lib/src/rum/web/rum_web_plugin.dart
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    return (time.millisecondsSinceEpoch - _navigationStart!).toJS;
+  }
+}
+
+@anonymous
+extension type OnRumStartOptions._(JSObject _) implements JSObject {
+  external JSFunction addEvent;
+}
```
marco-saia-datadog: nit: maybe it's redundant, but I would introduce an abstract class to reflect the [browser SDK interface](https://github.com/DataDog/browser-sdk/blob/bd6074ff1f33cb0b94acf7b6b7eae95180271475/packages/rum-core/src/domain/plugins.ts#L30), and perhaps reference it in a comment; mainly as a way to not lose track of `onInit` and `getConfigurationTelemetry()` existence, and to highlight that this class structure is defined externally

```dart
/// Mirrors the `RumPlugin` interface from the Datadog Browser SDK:
/// https://github.com/DataDog/browser-sdk/blob/bd6074ff1f33cb0b94acf7b6b7eae95180271475/packages/rum-core/src/domain/plugins.ts#L30
///
/// Keep this class updated to match the original interface.
abstract class RumPlugin {
  String get name;

  Map<String, dynamic>? getConfigurationTelemetry();

  void onInit({
    required RumInitConfiguration initConfiguration,
    required RumPublicApi publicApi,
  });

  void onRumStart(OnRumStartOptions options);
}
```

If this is overkill since we are only going to use it once, then I would at least add the comment for reference.

marco-saia-datadog (dismissed):
Looks good!

packages/datadog_flutter_plugin/lib/src/rum/web/rum_web_plugin.dart
fuzzybinary: Yeah, that's not a bad idea. I'll have to think about how to implement it in Dart though. I think I'll still end up with a Proxy object anyway, but I'll see what I can do.

marco-saia-datadog (approved):

Issue #781, opened 2025-07-29
Web + Identify not working


danielgomezriconeon:
### Describe the bug

Hey

I have a web build and it is reporting logs.

The problem is that identify does not set any `@usr.id` or the custom parameters.

I add the values like this and it work fine on ios and android, but on web nothing:

```dart
class DataDogAnalytics  {
  @override
  void identify(String id, {Map<String, Object> extraInfo = const {}}) {
    DatadogSdk.instance.setUserInfo(id: id, extraInfo: extraInfo);
  }

  @override
  void addAttribute(SessionAttribute attribute) {
    DatadogSdk.instance.rum?.addAttribute(attribute.key, attribute.value);

    DatadogSdk.instance.setUserInfo(
      extraInfo: {attribute.key: attribute.value},
    );
  }

  @override
  void track(String name, Map<String, Object?> attributes) {
    DatadogSdk.instance.rum?.addAction(RumActionType.tap, name, attributes);
  }
}
```

Version:
```yaml
datadog_flutter_plugin: 2.11.0
datadog_tracking_http_client: 2.3.0
```

index.html:
```html
  <script type="text/javascript" src="https://www.datadoghq-browser-agent.com/us5/v5/datadog-logs.js"></script>
  <script type="text/javascript" src="https://www.datadoghq-browser-agent.com/us5/v5/datadog-rum-slim.js"></script>
```

What can I do so that the logs are identified from web? it feels like the SDK from the web is detchaed from the SDK from the flutter side 🤔 

### Reproduction steps

1. Add a web project
2. Identify with any value
3. Add logs

### SDK logs

Even if Is set this I dont see any logs from DD:
```
 DatadogSdk.instance.sdkVerbosity = CoreLoggerLevel.debug;
```

### Expected behavior

All logs from web should have `@usr.id` attribute set with the identified value

### Affected SDK versions

2.11.0

### Latest working SDK version

-

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Flutter Version

3.29.2

### Device Information

Web build
Chrome

fuzzybinary:
Hi @danielgomezriconeon ,

Sorry, this is an oversight. The Flutter plugin sets this information in RUM, but not in Logs.

I'll look into getting a fix.

ishkawa:
Hi, thanks a lot for looking into and fixing this issue so quickly!

Any plans for when this fix might be released?
I’m asking since we’re planning to adopt Datadog Logs for our product.

[...]

PR #792, opened 2025-08-05
fix(web): Support user info in logs on web.

fuzzybinary:
### What and why?

`setUserInfo` only set user info on RUM, not Logs This was an unintentional oversight.

Web still does not support extra user info. This will be tracked as a separate issue and may be part of some changes coming in v3.

refs: #781

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/datadog_flutter_plugin_web.dart
+++ b/packages/datadog_flutter_plugin/lib/datadog_flutter_plugin_web.dart
@@ -5,6 +5,7 @@
 // ignore_for_file: unused_element
 
 import 'dart:async';
+import 'dart:js_interop';
 
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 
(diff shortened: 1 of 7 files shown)


packages/datadog_flutter_plugin/lib/datadog_flutter_plugin_web.dart
```diff
@@ -16,6 +17,19 @@ import 'src/logs/ddlogs_web.dart';
 import 'src/rum/ddrum_platform_interface.dart';
 import 'src/rum/ddrum_web.dart';
 
+@anonymous
+extension type JsUser._(JSObject _) implements JSObject {
```
marco-saia-datadog, line 21: nit: I would reference the original interface, as you did in https://github.com/DataDog/dd-sdk-flutter/pull/787

```suggestion
/// Mirrors the `User` interface from the Datadog Browser SDK: 
/// https://github.com/DataDog/browser-sdk/blob/562214cd1457e5d8445c1cb52424d9615f892158/packages/core/src/domain/contexts/userContext.ts#L10
///
/// Keep this class updated to match the original interface.
@anonymous
extension type JsUser._(JSObject _) implements JSObject {
```

marco-saia-datadog (approved):
Looks good! I have left a minor suggestion

packages/datadog_flutter_plugin/lib/datadog_flutter_plugin_web.dart
fuzzybinary: I may have to go back and do this with all structures, as they all have Browser SDK equivalents. I'll have to think about it...

PR #808, opened 2025-08-20
fix(web): Fix conversion of List to JSArray for Flutter Web

fuzzybinary:
### What and why?

Creation of JSArray was using JSArray.add, which translated to a Javascript `+` operator, which is incorrect. `.add` on JSArray in Dart 3.10 correctly translates to `Array.push`, but we are targeting Dart 3.3+, so we need to call `push` manually.

This was discovered because Dart 3.10's analyzer (currently in beta) complained that using `add` was only supported in Dart 3.10+, and we're targeting Dart 3.3. So this will also fix our analyzer issue in current Flutter beta checks.

To prevent further issues, I've add tests for `valueToJs` that run on browser (though `develop` does not currently run browser checks).

Fix multiple other analyzer errors.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/example/lib/rum_user_actions_screen.dart
+++ b/packages/datadog_flutter_plugin/example/lib/rum_user_actions_screen.dart
@@ -112,24 +112,22 @@ class _RumUserActionsScreenState extends State<RumUserActionsScreen> {
             });
           }),
         ),
-        Row(
-          children: [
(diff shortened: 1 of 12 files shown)


0xnm (dismissed):

sbarrio (dismissed):
👏

0xnm (approved):

Issue #809, opened 2025-08-21
`RumUserActionDetector` causes `RenderBox` error when using Page API


hmendez-figure:
### Describe the bug

Popping page/screen after using page API to push multiple pages/screens onto stack, the `RumUserActionDetector` causes `RenderBox` error. Probably because pages that are pushed onto stack this way are not actually rendered and do not have a `size` yet.

### Reproduction steps

Minimal reproducible example below when popping `Screen 2`

```dart
import 'package:datadog_flutter_plugin/datadog_flutter_plugin.dart';
import 'package:flutter/material.dart';

void main() {
  runApp(
    RumUserActionDetector(
      rum: null,
      child: MaterialApp(
        home: Navigator(pages: [page1, page2], onDidRemovePage: (page) {}),
      ),
    ),
  );
}

final MaterialPage<void> page1 = MaterialPage<void>(
  child: Scaffold(appBar: AppBar(title: const Text('Screen 1'))),
);

final MaterialPage<void> page2 = MaterialPage<void>(
  child: Scaffold(appBar: AppBar(title: const Text('Screen 2'))),
);
```

### SDK logs

<details>
<summary> Stacktrace </summary>

Reloaded 2 of 734 libraries in 101ms (compile: 4 ms, reload: 49 ms, reassemble: 40 ms).
Restarted application in 223ms.

════════ Exception caught by gesture library ═══════════════════════════════════
The following assertion was thrown while dispatching a pointer event:
RenderBox was not laid out: RenderSemanticsAnnotations#904ad NEEDS-LAYOUT NEEDS-PAINT
'package:flutter/src/rendering/box.dart':
Failed assertion: line 2251 pos 12: 'hasSize'

Either the assertion indicates an error in the framework itself, or we should provide substantially more information in this error message to help you determine and fix the underlying cause.
In either case, please report this assertion by filing a bug on GitHub:
  https://github.com/flutter/flutter/issues/new?template=02_bug.yml

When the exception was thrown, this was the stack:
#2      RenderBox.size (package:flutter/src/rendering/box.dart:2251:12)
#3      RenderBox.paintBounds (package:flutter/src/rendering/box.dart:3108:41)
#4      _RumUserActionDetectorState._getDetectingElementAtPosition.elementVisitor (package:datadog_flutter_plugin/src/rum/rum_user_action_detector.dart:307:53)
[...]

PR #810, opened 2025-08-22
fix: Don't attempt to get size of RenderBoxes that don't have size

fuzzybinary:
### What and why?

There are certain situations where scanning the element tree asks for the size of the element, both in RumUserActionDetector and in SessionReplay. However, there are situations where RenderBox will assert if you attempt to get its size. In both cases, check RenderBox's `hasSize` element before proceeding to get the element's size.

I added a screen to the Session Replay example that uses Slivers, as this is another situation in which using `size` might assert.

refs: #809

### Review checklist


packages/datadog_session_replay/example/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>12.0</string>
+  <string>13.0</string>
```
0xnm, line 24: just checking: is it expected change for this PR?

packages/datadog_session_replay/example/ios/Runner.xcodeproj/project.pbxproj
0xnm, line 1: same questions for this file

packages/datadog_session_replay/example/lib/screens/slivers_screen.dart
```diff
@@ -0,0 +1,30 @@
...
+                  (int index) => ListTile(title: Text('Item #$index')),
+                ),
+              ),
+            ),
+          ],
+        )
+      ),
+    );
+  }
+}
```
0xnm:
```suggestion
}

```

0xnm (dismissed):

packages/datadog_session_replay/example/ios/Flutter/AppFrameworkInfo.plist
fuzzybinary: It's not necessarily part of the PR specifically, but it's fine. Updated Flutter requires iOS 13 so these changes will slowly make their way into the generated project files.

Same with the project change.

0xnm (approved):

PR #811, opened 2025-08-22
chore: Add stale issue workflow.

fuzzybinary:
### What and why?

Adding a workflow that will automatically tag issues as `stale` if they are tagged as `awaiting response` for more than 14 days.

After 3 days of being marked `stale`, the workflow will close these issues automatically.

This PR sets this workflow as a debug workflow with no schedule as an initial test.

### Review checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,41 @@
+name: 'Automatically close stale issues'
+
+on:
+  # schedule:
+  # Runs every day at 7:00 AM GMT
+  #  - cron: '0 7 * * *'
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-issue-message: |
+            This issue has been automatically marked as stale because it has not had recent activity and has the `awaiting response` label.
+            
+            It will be closed if no further activity occurs within 3 days.
+        
+          # The label that will be added to the issues when automatically marked as stale
+          stale-issue-label: 'stale'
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+          # Close issues after 3 days of being marked stale
+          days-before-issue-close: 3
+          # Automatically remove the stale label when the issues or the pull requests are updated
+          remove-stale-when-updated: true
+          # Specify the reason used when closing issues: `completed` or `not_planned`
+          close-issue-reason: completed
+          # Run the stale workflow as dry-run.
+          # No GitHub API calls that can alter the issues and pull requests will happen.
+          # Useful to debug or when you want to configure the stale workflow safely. 
+          debug-only: true
```
0xnm:
```suggestion
          debug-only: true
 
```

.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
```
0xnm: minor: it is better pin to the particular hash, for the security https://github.com/DataDog/dd-sdk-android/blob/207c1990be4a1961b2d72e36530f88a1a13ffd63/.github/workflows/stale.yml#L16

0xnm (dismissed):

.github/workflows/stale.yml
fuzzybinary: @mariedm FYI, we might want to change other actions as well.

mariedm: Good point, I usually do it but forgot for some reason; I'll update on iOS

mariedm (dismissed):

0xnm (approved):

mariedm (approved):

PR #813, opened 2025-08-25
feat(sr): Support image capture in Session Replay

fuzzybinary:
### What and why?

Allow Session Replay tree traversal to capture images from Session Replay.

This works by using `RawImage`'s `toByteData(format: ImageByteFormat.rawRgba)` method, which gives a raw, RGBA image with premultiplied alpha. This is then copied directly to the native layer for processing.

On Android, we transfer the data using a `ByteBuffer`, and then use that buffer to create a `Bitmap` for compression and hashing. Because Android's raw bitmap format is also RGBA premultiplied (despite what the constant says), we can simply pass in the wrapped buffer to create the required `Bitmap`

On iOS, we create `Data` / `NSData` which copies the data directly from the buffer pointer. Doing this relatively safely requires we bypass the `objective_c` package, as using `address` in Dart requires calling an FFI leaf function (one that can be guaranteed not to call back into Dart) directly.

Because of the memory constraints of the copy, we are limiting copies to 640,000 pixels (an 800x800 image), which would be a raw size of 2.4 MB.

Any images larger than that are captured as "Large Image", shown below.
image

This still needs to support caching which images we've sent between sessions.                                                                                                     

### Review checklist


packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
```diff
@@ -0,0 +1,169 @@
...
+ * This uses synchronous, on-demand resource processing. When the processor requests
+ * the ID of a resource, it will compress the resource bytes, generate the MD5 hash, then
+ * write the resource to ResourcesFeature, which will upload the resources asynchronously.
+ */
+internal class DefaultResourceResolver(
+    val internalLogger: InternalLogger,
+    val resourceWriter: ResourceWriter,
+    val bitmapHandler: BitmapHandler = DefaultBitmapHandler(internalLogger)
+) : ResourceResolver {
+    private val resourceKeyMap: MutableMap<Int, ResourceResolver.ResourceEntry> = mutableMapOf()
```
jonathanmos, line 64: It looks like items are added to `resourceKeyMap`, but I don't see where they are cleared from it. Also, should we limit the size of the map? Same questions about `knownResources`.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
```diff
@@ -0,0 +1,169 @@
...
+            val messageDigest = MessageDigest.getInstance("MD5")
+            messageDigest.update(input)
+
+            val hashBytes = messageDigest.digest()
+
+            hashBytes.toHexString()
+        } catch (e: NoSuchAlgorithmException) {
+            internalLogger.log(
+                InternalLogger.Level.ERROR,
+                listOf(InternalLogger.Target.MAINTAINER, InternalLogger.Target.TELEMETRY),
```
jonathanmos, line 124: If we report this to telemetry, is it actionable?

packages/datadog_session_replay/android/src/test/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridgeTest.kt
```diff
@@ -0,0 +1,148 @@
...
+
+internal fun FlutterSessionReplayBridge.enableWithMock(
+    mockFeature: FlutterSessionReplayFeature
+) {
+    this.feature = mockFeature
+}
+
+@ExtendWith(ForgeExtension::class)
+class FlutterSessionReplayBridgeTest {
+    var mockCore: FeatureSdkCore = mockk(relaxed = true)
```
jonathanmos: suggestion: mocking the feature and creating the bridge happens in almost every test. What about putting them into a beforeEach?

packages/datadog_session_replay/android/src/test/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridgeTest.kt
```diff
@@ -0,0 +1,148 @@
...
+        // Given
+        val mockFeature = mockk<FlutterSessionReplayFeature>(relaxed = true)
+        val mockResourceResolver = mockk<ResourceResolver>(relaxed = true)
+        every { mockFeature.resourceResolver } returns mockResourceResolver
+
+        val bridge = FlutterSessionReplayBridge()
+        bridge.enableWithMock(mockFeature)
+
+        // When
+        val data = ByteBuffer.allocate(10)
```
jonathanmos: nit: we could use a random upper bounded number here

packages/datadog_session_replay/android/src/test/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolverTest.kt
```diff
@@ -0,0 +1,263 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+
+@ExtendWith(ForgeExtension::class)
+internal class ResourceResolverTest {
+    val mockInternalLogger = mockk<InternalLogger>()
+    val mockResourcesWriter = mockk<ResourceWriter>()
+    val mockBitmapHandler = mockk<BitmapHandler>()
+
+    fun createFakeImage(width: Int = 25, height: Int = 25, filledWith: Byte = 0): ByteBuffer {
```
jonathanmos: minor: should we randomize the image dimensions? Same question regarding all the usages of 25 here

packages/datadog_session_replay/lib/src/ios/datadog_session_replay_platform_ios.dart
```diff
@@ -101,4 +102,58 @@ class DatadogSessionReplayPlatformIos extends DatadogSessionReplayPlatform {
...
+      data: data,
+    );
+  }
+
+  @override
+  String? resourceIdForKey(int resourceKey) {
+    return _iosBridge.resourceIdForKey(resourceKey)?.toDartString();
+  }
+
+  // Bypass `package:objective_c` here to allow a direct leaf call. The allows us
```
mariedm: nit
```suggestion
  // Bypass `package:objective_c` here to allow a direct leaf call. This allows us
```

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Resource/ResouceResolver.swift
```diff
@@ -0,0 +1,88 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+import Foundation
+import DatadogInternal
+import CryptoKit
+import UIKit
+import CoreGraphics
+
+public protocol ResourceResolver {
```
mariedm: nit: file name is missing an `r` --> `ResourceResolver`.

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Resource/ResourcesWriter.swift
```diff
@@ -0,0 +1,143 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+import Foundation
+import DatadogInternal
+
+internal protocol ResourcesWrting {
```
mariedm: nit
```suggestion
internal protocol ResourcesWriting {
```

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
fuzzybinary: Probably not, but it will be good to know if it happens. This was pulled directly from the Android SDK so it's already being reported there.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
fuzzybinary: Right now we're not clearing them, except at the end of the session. I'd have to think about under what circumstances we wanted to clear them. What criteria does the Android SDK use? I didn't see it.

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Resource/ResourceRequestBuilder.swift
```diff
@@ -0,0 +1,93 @@
...
+internal struct ResourceRequestBuilder: FeatureRequestBuilder {
+    /// Custom URL for uploading data to.
+    let customUploadURL: URL?
+    /// Sends telemetry through sdk core.
+    let telemetry: Telemetry
+    /// Builds multipart form for request's body.
+    let multipartBuilder: MultipartFormDataBuilder
+
+    init(
+        customUploadURL: URL?,
```
mariedm, line 17: nit: I guess it could default to `nil`?
```suggestion
        customUploadURL: URL? = nil,
```

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Resource/EnrichedResource.swift
```diff
@@ -0,0 +1,49 @@
...
+            self.type = "resource"
+            self.application = .init(id: applicationId)
+        }
+    }
+    internal var identifier: String
+    internal var data: Data
+    internal var mimeType: String
+    internal var context: Context
+
+    internal init(
```
mariedm, line 26: nit: I don't think this init is needed; since `EnrichedResource` is a `struct`, you get memberwise initializer for free.

mariedm (dismissed):
Only a few nits, feel free to disregard them. LGMT! 🚀

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
jonathanmos: For the resources map (which I think is equivalent to resourceKeyMap) we're using an [lrucache](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/recorder/resources/ResourcesLRUCache.kt) to keep it implicitly bounded. For the DS knownResources I think it's unbounded but [deleted](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-session-replay/src/main/kotlin/com/datadog/android/sessionreplay/internal/resources/ResourceDataStoreManager.kt#L46) every 30 days when the data is considered to have expired.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
```diff
@@ -0,0 +1,169 @@
...
+        val height: Int,
+        // The resource Id which is the MD5 hash of the actual resource
+        var resourceId: String? = null,
+        // The actual byte array of the resource, which is valid only until
+        // the resource's hash is generated, at which point it is set to null
+        var resourceBytes: ByteBuffer?
+    )
+
+    /**
+     * Adds a resource to with the given Flutter Key to be resolved later.
```
jonathanmos:
```suggestion
     * Adds a resource with the given Flutter Key to be resolved later.
```

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -13,6 +17,11 @@ import '../view_tree_snapshot.dart';
 // overlappping other content in the replay.
```
jonathanmos:
```suggestion
// overlapping other content in the replay.
```

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Feature/FlutterSeasionReplayFeature.swift
```diff
@@ -5,7 +5,17 @@
 import Foundation
```
jonathanmos, line 5: nit: spelling mistake in the name of this file. Should be SessionReplay and says SeasionReplay

jonathanmos (dismissed):

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Resource/ResourceRequestBuilder.swift
fuzzybinary: It could, but since this is internal I'm okay with it not having a default.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceResolver.kt
fuzzybinary: I'll probably do that as a separate task. Created RUM-11500 to track

packages/datadog_session_replay/android/src/test/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridgeTest.kt
fuzzybinary: I feel like it's a style thing. I'm actually not sure why this mockCore is outside of the one test that uses it.

I tend to want the thing under test created in each test method, instead of relying on a `setUp` or `beforeEach` to create it for me unless the setup is significantly complicated, which here  it's not.

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/Resource/EnrichedResource.swift
fuzzybinary: I don't know why this is needed, but it is apparently?

jonathanmos (approved):

PR #829, opened 2025-09-06
feat(sr): Support privacy options for text an images 

fuzzybinary:
### What and why?

This implements TextAndInputPrivacy and ImagePrivacy options both at a root capture level and through `SessionReplayPrivacy` override widgets.

Moved golden tests to `example` so that we could add and test asset privacy options in golden tests without polluting the package with unnecessary assets.

Touch input privacy is currently not implemented and will be a follow-up PR.

refs: RUM-11127 RUM-11126

### Review checklist


packages/datadog_session_replay/example/golden_test/image_masking_golden_test.dart
```diff
@@ -0,0 +1,179 @@
...
+const String networkImageUrl =
+    'https://datadog-docs.imgix.net/img/dd_logo_n_70x75.png';
+
+void main() {
+  late SessionReplayRecorder recorder;
+  late RUMContext context;
+  late MockDatadogSessionReplayPlatform platform;
+
+  setUp(() {
+    // Needed to fetch newtork images
```
0xnm:
```suggestion
    // Needed to fetch network images
```

packages/datadog_session_replay/example/golden_test/snapshot_renderer.dart
```diff
@@ -129,7 +135,49 @@ extension TextWireframeRendering on SRTextWireframe {
...
+    final imageData = mockPlatform.imageCache[resourceId!];
+    if (imageData != null && imageData.image != null) {
+      canvas.save();
+      Rect srcRect = Rect.fromLTWH(
+        0,
+        0,
+        imageData.width.toDouble(),
+        imageData.height.toDouble(),
+      );
+      canvas.drawImageRect(imageData.image!, srcRect, toRect(), Paint());
```
0xnm, line 177: minor: since `Paint()` object doesn't depend on anything, in order to reduce the number of object allocations on the hot paths maybe it is possible to store it is a property instead of creating a new instance at each `render` call?

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -21,14 +21,68 @@ enum TextAndInputPrivacyLevel {
...
+  /// No images will be recorded.
+  maskAll,
+
+  // All images will be recorded, including ones downloaded from the internet.
+  maskNone,
+}
+
+/// Available privacy levels for touch masking in Session Replay.
+///
+/// Unlike other privacy levels, the TouchPrivacyLevel is global and cannot
```
0xnm:
```suggestion
/// Unlike other privacy levels, the [TouchPrivacyLevel] is global and cannot
```

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -21,14 +21,68 @@ enum TextAndInputPrivacyLevel {
...
+/// currently be overriden in subtrees using the [SessionReplayPrivacy] widget.
+enum TouchPrivacyLevel {
+  /// Show all user touches.
+  show,
+
+  // Hide all user touches.
+  hide,
+}
+
+/// Configuration options for Flutter Session replay, including
```
0xnm:
```suggestion
/// Configuration options for Session Replay, including
```

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -104,38 +110,52 @@ class SessionReplayRecorder {
   final Map<Key, Element> _elements = {};
   RUMContext? _currentContext;
   bool _captureInProgress = false;
-  CapturePrivacy _defaultCapturePrivacy;
+  TreeCapturePrivacy _defaultTreeCapturePrivacy;
+  // Ignore for now. Will be tackled shortly
```
0xnm: it is better to add `TODO` here

packages/datadog_session_replay/pubspec.yaml
```diff
@@ -34,7 +34,7 @@ dev_dependencies:
   datadog_common_test:
     path: ../datadog_common_test
 
-flutter:
+flutter:
```
0xnm: probably this change needs to be reverted

0xnm (dismissed):

packages/datadog_session_replay/example/golden_test/text_masking_golden_test.dart
```diff
@@ -0,0 +1,412 @@
...
+    );
+
+    context = RUMContext(
+      applicationId: randomString(),
+      sessionId: randomString(),
+    );
+    recorder.updateContext(context);
+  });
+
+  testWidgets('global unmasked text', (tester) async {
```
mariedm, line 60: Do you have anything similar to secure textfields (e.g., for password) on Flutter you could test the `maskSensitiveInputs` privacy on?

packages/datadog_session_replay/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 77;
+	objectVersion = 54;
```
mariedm, line 6: Is this change intended?

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -7,7 +7,7 @@ import 'package:datadog_flutter_plugin/datadog_flutter_plugin.dart';
 import 'src/datadog_session_replay_plugin.dart';
 
 export 'src/datadog_session_replay.dart' show DatadogSessionReplay;
-export 'src/widgets.dart' show SessionReplayCapture;
+export 'src/widgets.dart' show SessionReplayCapture, SessionReplayPrivacy;
 
 /// Privacy levels for text and input masking in Session Replay
```
mariedm: tiny nit: align with others
```suggestion
/// Available privacy levels for text and input masking in Session Replay.
```

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -21,14 +21,68 @@ enum TextAndInputPrivacyLevel {
...
+}
+
+/// Configuration options for Flutter Session replay, including
+/// default privacy levels.
 class DatadogSessionReplayConfiguration {
+  /// The sampling rate for Session Replay. This is applied in addition to the
+  /// RUM session sample rate.
+  ///
+  /// It must be a number between 0.0 and 100.0, where 0 means no replays will
+  /// be recorded and 100 means all RUM sessions will contain replay.
```
mariedm:
```suggestion
  /// be recorded and 100 means all sampled RUM sessions will contain replay.
```

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -21,14 +21,68 @@ enum TextAndInputPrivacyLevel {
...
+  /// in RUM, and within those sessions, only 20% will have replays.
   double replaySampleRate;
+
+  /// Defines the way text and input (e.g. TextFields and CheckBoxes) should be
+  /// masked.
+  ///
+  /// Defaults to [TextAndInputPrivacyLevel.maskAll]
   TextAndInputPrivacyLevel textAndInputPrivacyLevel;
+
+  /// Defines image privacy level.
```
mariedm:
```suggestion
  /// Defines the way images should be captured.
```

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -21,14 +21,68 @@ enum TextAndInputPrivacyLevel {
...
+  /// be recorded and 100 means all RUM sessions will contain replay.
+  ///
+  /// Note: This sample rate is applied in addition to the RUM sample rate. For
+  /// example, if RUM uses a sample rate of 80% and Session Replay uses a sample
+  /// rate of 20%, it means that out of all user sessions, 80% will be included
+  /// in RUM, and within those sessions, only 20% will have replays.
   double replaySampleRate;
+
+  /// Defines the way text and input (e.g. TextFields and CheckBoxes) should be
+  /// masked.
```
mariedm:
```suggestion
  /// Defines the way text and input (e.g. TextFields and CheckBoxes) should be
  /// captured.
```

packages/datadog_session_replay/lib/src/capture/capture_node.dart
```diff
@@ -32,3 +32,38 @@ abstract class CaptureNode {
...
 }
+
+/// Common Attribute Nodes
+
+@immutable
+class PlaceholderNode extends CaptureNode {
+  final int wireframeId;
+
+  /// The text to display if the width of the element is above [minWidth].
+  final String caption;
```
mariedm, line 43: I don't think we do this check on iOS, but that's a good idea. Out of curiosity, do you know what happens in the player when the text's width is too large? Does it simply cuts it? I wonder how "ugly" it is and if we should do the same on the native side (even though I consider it a nit).

packages/datadog_session_replay/example/golden_test/snapshot_renderer.dart
fuzzybinary: Probably better, but this is only for golden tests so I'm not too concerned about it.

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -46,15 +65,35 @@ class ImageRecorder implements ElementRecorder {
     }
 
     final elementId = keyGenerator.keyForElement(element);
+    // AssetImages loosen their masking to [ImagePrivacyLevel.maskNone] when
+    // they. need to, so if [ImagePrivacyLevel.maskNonAssetsOnly] is still set, then
+    // we shouldn't capture this image.
```
mariedm:
```suggestion
    // AssetImages loosen their masking to [ImagePrivacyLevel.maskNone] when
    // they need to, so if [ImagePrivacyLevel.maskNonAssetsOnly] is still set, then
    // we shouldn't capture this image.
```

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -46,15 +65,35 @@ class ImageRecorder implements ElementRecorder {
     }
 
     final elementId = keyGenerator.keyForElement(element);
+    // AssetImages loosen their masking to [ImagePrivacyLevel.maskNone] when
```
mariedm, line 68: Would there be a cleaner approach where we don't change the image privacy level and rather use an "asset flag" in the context to track the type, and then check the real privacy level and if it's an asset?

Also, could there be any other element(s) down an asset's subtree that wouldn't be an asset and could be wrongfully captured when it shouldn't?

packages/datadog_session_replay/lib/src/widgets.dart
```diff
@@ -54,6 +54,49 @@ class _SessionReplayCaptureState extends State<SessionReplayCapture> {
...
+/// [TextAndInputPrivacyLevel], [ImagePrivacyLevel], and whether a tree should
+/// be hidden from Session Replay
+///
+/// Privacy overrides can be modified multiple times in a widget tree, however,
+/// when a widget is hidden, it is replaced by a placeholder labeled as "Hidden"
+/// in the replay, and its subviews are not processed or recorded. Therefore, it
+/// is not possible to "unhide" a widget that is deeper in the tree of a hidden
+/// widget.
+@immutable
+class SessionReplayPrivacy extends StatelessWidget {
```
mariedm, line 72: Suggestion: I feel like `SessionReplayPrivacy` is too generic and doesn't clearly indicate it's for an override. Some alternatives could be `SessionReplayPrivacyOverride` or `SessionReplayOverride`. I'm saying this, but SwiftUI, which uses a similar approach to the Flutter one with widgets, uses `SessionReplayPrivacyView`.

mariedm (commented):
Good job on the overrides! I have a few questions and nits, but otherwise LGMT! 👍

packages/datadog_session_replay/example/golden_test/text_masking_golden_test.dart
fuzzybinary: No, I'll add one.

packages/datadog_session_replay/example/ios/Runner.xcodeproj/project.pbxproj
fuzzybinary: Flutter and Xcode change this occasionally. I honestly have no idea why, but they seem to automatically adjust it to the right value when needed 😅

packages/datadog_session_replay/lib/src/capture/capture_node.dart
fuzzybinary: It overflows and looks ugly. especially on very small images. Its definitely a nit but I hated it when I saw it in replays so....

packages/datadog_session_replay/lib/src/widgets.dart
fuzzybinary: Yeah, I'll have to think about it. I did base the name on iOS 😅. For use actions we use `RumUserActionAnnotation` so maybe `SessionReplayAnnotation` makes sense.

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
fuzzybinary: > Would there be a cleaner approach where we don't change the image privacy level and rather use an "asset flag" in the context to track the type, and then check the real privacy level and if it's an asset?

Possibly, but I'd actually have to refactor this to accept a "context", something it doesn't currently do. It's possibly a nice refactor.

> Also, could there be any other element(s) down an asset's subtree that wouldn't be an asset and could be wrongfully captured when it shouldn't?

It is theoretically possible, but unlikely, since `Image` doesn't expose a 'child' member. It does expose a `loadingBuilder`, `frameBuilder`, and `errorBuilder`, which are within its widget tree but could contain anything, but it would be very weird for someone to use those to display a second image that wasn't an assset.

packages/datadog_session_replay/lib/src/widgets.dart
mariedm: Actually we use `sessionReplayOverrides` on UIKit

packages/datadog_session_replay/example/golden_test/text_masking_golden_test.dart
fuzzybinary: Oh, yes, I do. `material text fields mask user input` (same exists for cupertino widgets).

mariedm (approved):

packages/datadog_session_replay/lib/src/widgets.dart
mariedm: I missed the `SessionReplayAnnotation` suggestion: don't you think we should have at least `privacy` or `override` in the name?

mariedm: Just a suggestion, in any case I'll leave it to you

PR #831, opened 2025-09-10
feat: Improve capture performance

fuzzybinary:
### What and why?

Instead of asking for semantics from every recorder, bind recorders to types so only the relevant ones are asked for capture semantics. This allows us to avoid capturing paint bounds and doing matrix translations for every element in the tree, only the ones that we need to capture semantics for.

Additionally, use `addPostFrameCallback` to make it so we only capture the tree if there was a change to the widget tree.

refs: RUM-11123

### More detail

One of the first frames after boot is usually the worst. Prior to optimization in Shopist, which could also include a GC pass.
image

A more representative sample without a GC pass registers at 11ms or so:
image

Post optimization, since we create less garbage (fewer Semantics objects and Matrix objects) we see fewer GC passes and faster results:
image

And this is the worst case with a GC trigger:
image

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/lib/src/capture/recorder.dart
+++ b/packages/datadog_session_replay/lib/src/capture/recorder.dart
@@ -108,7 +108,6 @@ class CaptureResult {
 
 class SessionReplayRecorder {
   final DatadogTimeProvider _timeProvider;
-  // final List<ElementRecorder> _elementRecorders;
   final Map<Type, ElementRecorder> _elementRecordersByType = {};
 
   final Map<Key, Element> _elements = {};
@@ -142,7 +141,7 @@ class SessionReplayRecorder {
     this._defaultTreeCapturePrivacy,
     this._touchPrivacyLevel,
   ) {
-    _sortElementRecorders([
+    _populateElementRecorderMap([
       ContainerRecorder(keyGenerator),
       TextElementRecorder(keyGenerator),
       EditableTextRecorder(keyGenerator),
@@ -162,7 +161,7 @@ class SessionReplayRecorder {
   }) : _timeProvider = timeProvider,
        _defaultTreeCapturePrivacy = defaultCapturePrivacy,
        _touchPrivacyLevel = touchPrivacyLevel {
-    _sortElementRecorders(elementRecorders);
+    _populateElementRecorderMap(elementRecorders);
   }
 
   void updateContext(RUMContext? context) {
@@ -269,7 +268,7 @@ class SessionReplayRecorder {
     }
   }
 
-  void _sortElementRecorders(List<ElementRecorder> recorders) {
+  void _populateElementRecorderMap(List<ElementRecorder> recorders) {
     for (final recorder in recorders) {
       for (final type in recorder.handlesTypes) {
         _elementRecordersByType[type] = recorder;
--- a/packages/datadog_session_replay/lib/src/capture/recorder.dart
+++ b/packages/datadog_session_replay/lib/src/capture/recorder.dart
@@ -1,6 +1,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2025-Present Datadog, Inc.
 
+import 'dart:developer';
 import 'dart:ui' as ui;
 
 import 'package:datadog_flutter_plugin/datadog_internal.dart';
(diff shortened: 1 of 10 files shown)


0xnm:
seems like https://github.com/DataDog/dd-sdk-flutter/pull/829 should be a base target for this PR?

fuzzybinary:
I was hoping #829 would be in by now :)

fuzzybinary:
@0xnm This should now be ready for review.

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -105,7 +108,8 @@ class CaptureResult {
 
 class SessionReplayRecorder {
   final DatadogTimeProvider _timeProvider;
-  final List<ElementRecorder> _elementRecorders;
+  // final List<ElementRecorder> _elementRecorders;
```
awforsythe: nit: commented-out code

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -137,25 +141,29 @@ class SessionReplayRecorder {
...
-        ContainerRecorder(keyGenerator),
-        TextElementRecorder(keyGenerator),
-        EditableTextRecorder(keyGenerator),
-        InputDecoratorRecorder(keyGenerator),
-        ImageRecorder(keyGenerator),
-        CustomPaintRecorder(keyGenerator),
-        PrivacyRecorder(keyGenerator),
-      ];
+  ) {
+    _sortElementRecorders([
```
awforsythe: nit: if `_sortElementRecorders` is non-pure for a good reason, then something like `_populateElementRecorders` might do a better job of communicating that it initializes this object's state as a side effect

packages/datadog_session_replay/lib/src/datadog_session_replay.dart
```diff
@@ -17,6 +17,13 @@ import 'processor/processor.dart';
 import 'rum_context.dart';
 
 class DatadogSessionReplay {
+  // The minimum amount of time that needs to pass before we perform another
+  // tree capture.
+  static const minCaptureTiming = Duration(milliseconds: 100);
+  // The number of times in quick succession thar SR capture can throw before
+  // we shut it down completely.
+  static const errorTollerance = 10;
```
awforsythe, line 25: sp: `errorTolerance` (one L)

packages/datadog_session_replay/lib/src/datadog_session_replay.dart
```diff
@@ -66,22 +77,25 @@ class DatadogSessionReplay {
...
+
+  void _startPeriodicCapture() async {
+    Timer.periodic(minCaptureTiming, (timer) async {
+      bool shouldScheduleNextCapture = true;
+      if (_needCapture) {
         try {
           final captureResult = await _recorder.performCapture();
           if (captureResult != null) {
             _processor.process(captureResult);
           }
```
awforsythe: OK, so `_startPeriodicCapture()` kicks off a periodic timer that will perform a perform a capture when it procs, but only if the UI has rendered since the last capture. And if we exceed an error threshold, we cancel the timer.

The logic is a little obscure here (given that there are two independent timing mechanisms, the periodic timer vs. the post-render callback), so it might benefit from a comment or two, but it looks pretty sound once I get my head around it.

To be very nitpicky: `_needCapture` might be more accurately named `_hasRenderedSinceLastCapture`, and `shouldScheduleNextCapture` feels a bit misleading given that calling `addPostFrameCallback` is not technically, in and of itself, scheduling the capture... but take my semantic gripes with a grain of salt 🙂

awforsythe (dismissed):
Ship it!

packages/datadog_session_replay/lib/src/capture/recorder.dart
fuzzybinary: Yeah that should be removed. Nice catch.

packages/datadog_session_replay/lib/src/datadog_session_replay.dart
fuzzybinary: Yeup that's all correct. I'll add some comments to make it clearer.

awforsythe (approved):

PR #837, opened 2025-09-17
feat(sr): Implement touch privacy

fuzzybinary:
### What and why?

The `SessionReplayPrivacy` widget can now override touch privacy for a tree. It does this by canceling touch events from a given pointer until the next capture. This can mean we might miss some taps if they are moving in and out of a private area within a capture period.

Add tests for both pointer recorder and associated widgets.

refs: RUM-11681

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/example/lib/screens/touch_privacy_screen.dart
+++ b/packages/datadog_session_replay/example/lib/screens/touch_privacy_screen.dart
@@ -1,0 +1,88 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+import 'package:datadog_session_replay/datadog_session_replay.dart';
+import 'package:flutter/material.dart';
+
+class TouchPrivacyScreen extends StatefulWidget {
+  const TouchPrivacyScreen({super.key});
+
+  @override
+  State<TouchPrivacyScreen> createState() => _ImagesScreenState();
+}
+
+class _ImagesScreenState extends State<TouchPrivacyScreen> {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(title: Text('Touch Privacy')),
+      body: SingleChildScrollView(
+        child: Column(
+          children: [
+            Text('Here touch recording should work'),
+            SizedBox(height: 20.0),
+            Text('The below pin pad should not show touches'),
+            _pinPad(),
+          ],
+        ),
+      ),
+    );
+  }
+
+  Widget _pinPad() {
+    return SessionReplayPrivacy(
+      touchPrivacyLevel: TouchPrivacyLevel.hide,
+      child: Padding(
+        padding: EdgeInsetsGeometry.all(10),
+        child: Column(
+          spacing: 20.0,
+          children: [
+            Row(
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                _PinButton(text: '1'),
+                _PinButton(text: '2'),
+                _PinButton(text: '3'),
+              ],
+            ),
+            Row(
+              mainAxisAlignment: MainAxisAlignment.center,
+              children: [
+                _PinButton(text: '4'),
+                _PinButton(text: '5'),
+                _PinButton(text: '6'),
+              ],
+            ),
+            Row(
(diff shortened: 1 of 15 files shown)


packages/datadog_session_replay/lib/src/capture/pointer_capture.dart
```diff
@@ -32,10 +32,34 @@ class PointerCapture {
   });
 }
 
+/// Provides a PointerSnapshotRecorder to other classes that need it.
+/// This is used primarily by [SessionReplayPrivacy] to cancel pointer
+/// tracking that occurs in its subtree if [TouchPrivacyLevel ]
```
mariedm:
```suggestion
/// tracking that occurs in its subtree if [TouchPrivacyLevel] is set
```

packages/datadog_session_replay/lib/src/capture/pointer_capture.dart
```diff
@@ -57,6 +85,20 @@ class PointerSnapshotRecorder {
       ),
     );
   }
+  
+  // Uncapturing a pointer removes any previous events from the pointer, and
+  // flags it as hidden until the next snapshot. This is partially to support
+  // 'hover' events that don't have up / down that are trackable, and because
+  // "up" triggers occur in reverse tree order, which means we can't 'uncapture'
+  // the tracking that happens at the root of the tree
```
mariedm:
```suggestion
  // the tracking that happens at the root of the tree.
```

packages/datadog_session_replay/example/lib/screens/touch_privacy_screen.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:flutter/material.dart';
+
+class TouchPrivacyScreen extends StatefulWidget {
+  const TouchPrivacyScreen({super.key});
+
+  @override
+  State<TouchPrivacyScreen> createState() => _ImagesScreenState();
+}
+
+class _ImagesScreenState extends State<TouchPrivacyScreen> {
```
mariedm: Should it be `_TouchPrivacyScreenState`?
```suggestion
class _TouchPrivacyScreenState extends State<TouchPrivacyScreen> {
```

packages/datadog_session_replay/lib/src/capture/element_recorders/container_recorder.dart
```diff
@@ -80,4 +82,36 @@ class ContainerRecorder implements ElementRecorder {
       cornerRadius: borderStyle?.cornerRadius ?? 0.0,
     );
   }
+
+  CapturedViewAttributes _adjustAttributesForShape(
```
mariedm, line 86: This is not related to touch privacy, right? Do I understand correctly that it "corrects" rounded shapes to improve its representation in replay?

packages/datadog_session_replay/lib/src/capture/pointer_capture.dart
```diff
@@ -65,6 +107,7 @@ class PointerSnapshotRecorder {
 
     final copy = _pointerBuffer;
     _pointerBuffer = [];
+    _hiddenPointers.clear();
```
mariedm, line 110: If I understand correctly, `_hiddenPointers` is cleared at every snapshot taken? Wouldn't that cause to remove some pointers too soon? 

An alternative could be to clear a pointer from the set when interaction is "up". And/or add a max size and remove a value when that happes.

ambushwork (dismissed):

mariedm (dismissed):

packages/datadog_session_replay/lib/src/capture/element_recorders/container_recorder.dart
fuzzybinary: It's not related to touch privacy, it's more of a visual bug I discovered while creating the pin pad. It created pills instead of circles if I don't adjust the shape properly.

packages/datadog_session_replay/lib/src/capture/pointer_capture.dart
fuzzybinary: Yes, it does. But, in testing this doesn't actually cause any problems because of how down / up events happen in the tree (and how SR interprets them).

The reason it's like this is because of potential Desktop support. `hover` events shouldn't be tracked in hidden views, so up / down isn't enough, and unfortunately Flutter doesn't offer `hoverEntered` / `hoverExited` events to do something similar.

awforsythe (dismissed):
Commits made since last approval are trivial and LGTM.

PR #853, opened 2025-09-25
fix: improve test and add locations.toList

tyandl:
fix: GqlErrors causing stopResource ArgumentError

When handling gql response with errors, errors are serialized and passed to rum.stopResource as an attribute however, attributes must be of type Map<String, Object?> but when gqllink has an error an attribute of MappedListIterable<GraphQLError, Map<String, Object?>> is add. By calling toList() on the serialized errors stopResources will no longer throw an exception.

### What and why?

What changes does this pull request introduce and why is it necessary?
resolve MappedListIterable to list in serializeErrors function by calling `.toList()` avoiding a downstream exception in RUM.stopresource()

### How?

### Review checklist

### DIFF
--- a/packages/datadog_gql_link/test/datadog_gql_link_test.dart
+++ b/packages/datadog_gql_link/test/datadog_gql_link_test.dart
@@ -666,6 +666,8 @@ query UserInfo($id: ID!) {
       final captured = verify(() => mockRum.stopResource(
           any(), any(), RumResourceType.native, any(), captureAny()));
       final capturedAttrs = captured.captured[0] as Map<String, dynamic>;
+      expect(capturedAttrs['_dd']['graphql']['errors'].runtimeType,
+          List<Map<String, Object?>>);
       expect(capturedAttrs['_dd']['graphql']['errors'], [
         {
           'message': 'GraphQL Error Message',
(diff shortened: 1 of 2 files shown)


packages/datadog_gql_link/test/datadog_gql_link_test.dart
```diff
@@ -666,6 +666,8 @@ query UserInfo($id: ID!) {
       final captured = verify(() => mockRum.stopResource(
           any(), any(), RumResourceType.native, any(), captureAny()));
       final capturedAttrs = captured.captured[0] as Map<String, dynamic>;
+      expect(capturedAttrs['_dd']['graphql']['errors'].runtimeType,
+          List<Map<String, Object?>>);
       expect(capturedAttrs['_dd']['graphql']['errors'], [
```
fuzzybinary: This is fine but better might be if we use `findInvalidAttribute` and ensure it returns `null` (this will ensure all the keys / values are properly serializable as well).

fuzzybinary (changes requested):
Thanks for the contribution!

One minor thought on how this could be improved.  Also, if you could modify your commit to use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) that will help some of our automated tooling.

fuzzybinary:
We also need you to [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) if you can.

Just let me know if you want me to reopen this myself, and I'll try to get it in this week.

tyandl:
@fuzzybinary I would appreciate the help across the finish line. 
I think I have corrected the commit message and have signed the commit.

packages/datadog_gql_link/test/datadog_gql_link_test.dart
tyandl: good call. and it found an additional issue. Both test and additioanl fix added.

fuzzybinary (approved):

PR #868, opened 2025-10-16
feat: Support mandatory user and account info.

fuzzybinary:
### What and why?

Also support clearing user info, and check that account info and anonymous ids are coming in through Logs.

refs: RUM-8707 RUM-8675

### Review checklist


packages/datadog_flutter_plugin/MIGRATING.md
```diff
@@ -17,6 +17,11 @@ Clients using Flutter Web should update to using the Datadog Browser SDK v6.  Ch
 +  <script type="text/javascript" src="https://www.datadoghq-browser-agent.com/us1/v6/datadog-rum-slim.js"></script> 
 ```
 
+# RUM Changes
+
+An ID is no longer optional in `setUserInfo`. If you need to clear user info, use `clearUserInfo` instead.
+
+
 # Migration from 1.x to 2.0
```
0xnm, line 24:
```suggestion
An ID is no longer optional in `setUserInfo`. If you need to clear user info, use `clearUserInfo` instead.

# Migration from 1.x to 2.0
```

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogSdkPluginTest.kt
```diff
@@ -693,6 +704,211 @@ class DatadogSdkPluginTest {
...
+        plugin.onMethodCall(methodCall, mockResult)
+
+        // THEN
+        val core = Datadog.getInstance()
+        val coreFeature: Any = core.getPrivate("coreFeature")!!
+        // Ensure safe write before attempting to read
+        Thread.sleep(10)
+        val userInfo: UserInfo? = coreFeature
+            .getFieldValue<Any, Any>("userInfoProvider")
+            .getFieldValue("internalUserInfo")
```
0xnm, line 749: alternatively you can cast core to `InternalSdkCore` and call `getDatadogContext()` to avoid reflection and flaky `Thread.sleep`

packages/datadog_flutter_plugin/lib/datadog_flutter_plugin.dart
```diff
@@ -241,6 +260,51 @@ class DatadogSdk {
...
+    });
+  }
+
+  /// Add custom attributes to the current user information
+  ///
+  /// This extra info will be added to already existing extra info that is added
+  /// to logs, traces, and RUM events automatically.
+  ///
+  /// Setting an existing attribute to `null` will remove that attribute from
+  /// the user's extra info
```
0xnm: this should be about account, not user

0xnm (dismissed):

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/DatadogSdkPluginTest.kt
fuzzybinary: When attempting to do that I get an exception trying to call `getSafe`

> java.lang.NullPointerException: Cannot invoke "android.content.res.Resources.getConfiguration()" because the return value of "android.content.Context.getResources()" is null

Is there a way around that?

fuzzybinary: Sorry, this appears to be happening when attempting to get the `deviceType`. Specifically `isTablet`

0xnm: Probably then you need to mock `Context` you pass to the SDK initialization. But if it is too complex, you can use reflection (although it imposes using `Thread.sleep`).

0xnm (dismissed):

0xnm (approved):

PR #877, opened 2025-10-31
feat: Enable JNI mappers for Android

fuzzybinary:
### What and why?

Switch from using Method Channel communication for event mappers to using JNI mappers on Android only.

The old method of using Method Channels forced multiple thread transitions and used a timeout lock to make sure we didn't stall the Android SDK's upload worker thread for too long. The mapper would have to transition from the  Android SDK's worker thread, to the main thread, to Flutter's main Isolate , and back before successfully completing the mapping.

Using JNI mappers, we're avoid some of the complexity in our code, and allow Flutter's JNI library to take over some of it. We also potentially avoid a transition to the main thread, since Flutter can transition directly from the Android SDK's worker thread to Flutter's main isolate.

iOS, unfortunately, does not yet have a way in their ffi / objective-c libraries to allow this type of blocking callback, and in fact does not allow callbacks to return data at all.  This means using FFI would still necessitate us managing all of the thread transitions, and add the complexity of FFI on top of that.  For that reason, we are sticking with MethodChannels for iOS over FFI callbacks.

refs: RUM-12052

### Review checklist


estherk15 (dismissed):

packages/datadog_flutter_plugin/jnigen.yaml
```diff
@@ -0,0 +1,20 @@
...
+  dart:
+    path: lib/src/android/datadog_android_bridge.dart
+    structure: single_file
+source_path:
+  - 'android/src/main/kotlin'
+classes:
+  - 'com.datadoghq.flutter.DatadogRumEventMapper'
+  - 'com.datadoghq.flutter.DatadogRumPlugin'
+  - 'com.datadoghq.flutter.DatadogLogsPlugin'
+  - 'com.datadoghq.flutter.DatadogLogEventMapper'
```
0xnm:
```suggestion
  - 'com.datadoghq.flutter.DatadogLogEventMapper'

```

0xnm (dismissed):

packages/datadog_flutter_plugin/jnigen.yaml
```diff
@@ -0,0 +1,21 @@
...
+
```
0xnm, line 21:
```suggestion
  - 'com.datadoghq.flutter.DatadogLogEventMapper'

```

0xnm (approved):

PR #880, opened 2025-11-06
fix: Support properly stopping Dio's `badResponse` exceptions

fuzzybinary:
### What and why?

Dio can throw an exception when it encounters certain user defined error codes in `validateStatus`. However, we need to stop these resources as non-errors in RUM for them to show properly.

Detect `badResponse` exceptions and properly stop them before forwarding the error to other Dio handlers.

refs RUM-12688

### Review checklist

### DIFF
--- a/packages/datadog_dio/example/lib/screens/instrumentation_second_screen.dart
+++ b/packages/datadog_dio/example/lib/screens/instrumentation_second_screen.dart
@@ -64,6 +64,15 @@ class _InstrumentationSecondScreenState
     });
     await widget.dio.postUri(Uri.parse(_config.thirdPartyPostUrl));
 
+    setState(() {
+      _buttonText = 'Third party missing url';
+    });
+    try {
+      await widget.dio.getUri(Uri.parse(_config.thirdPartyMissingUrl));
+    } catch (_) {
+      // Do nothing, this will throw from Dio
+    }
+
     await Future.delayed(const Duration(milliseconds: 100));
     setState(() {
       _buttonText = 'All Done';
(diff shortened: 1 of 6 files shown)


packages/datadog_dio/example/lib/screens/instrumentation_second_screen.dart
```diff
@@ -64,6 +64,15 @@ class _InstrumentationSecondScreenState
...
     await widget.dio.postUri(Uri.parse(_config.thirdPartyPostUrl));
 
+    setState(() {
+      _buttonText = 'Third party missing url';
+    });
+    try {
+      await widget.dio.getUri(Uri.parse(_config.thirdPartyMissingUrl));
+    } catch (_) {
+      // Do nothing, this will throw from Dio
+    }
```
awforsythe, line 74: nit: At first glance, it's not entirely clear whether the intention of this `try` block is to catch errors from `dio.getUri` or from `Uri.parse`. It might be cleaner if we construct the `Uri` up-front, so the `try`/`catch` covers only the case we care about.

awforsythe (approved):

packages/datadog_dio/example/lib/screens/instrumentation_second_screen.dart
fuzzybinary: 👍 I agree, but if Uri parsing throws integration test test would fail, which is all that would happen in the up-front approach as well.

If this were anything other than an integration test I would change it.

0xnm (approved):

PR #881, opened 2025-11-07
feat: Support RUM context in trace headers.

fuzzybinary:
### What and why?

Add user id, account id, and session id to the `baggage` header in network requests.

The method for getting a user id and account id is different between iOS and Android, as iOS does not have FFI setup and Android does not broadcast context changes the way iOS does, so different implementation approaches needed to be used on each platform.

Web does not support these baggage headers at the moment, though web handles most of its own networking through XHR requests, not through our interceptors.

refs: RUM-8411

### Review checklist


packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
```diff
@@ -38,6 +40,11 @@ class DatadogSdkPlugin : FlutterPlugin, MethodCallHandler {
         // but the application will still be running. We keep the configuration separate
         // from the plugin to warn about reinitialization.
         var previousConfiguration: Map<String, Any?>? = null
+
+        fun getCoreContext(): DatadogContext? {
+            val sdkCore = Datadog.getInstance() as? InternalSdkCore
+            return sdkCore?.getDatadogContext()
```
0xnm, line 46: Beware that this call is blocking. It will add a request to the context queue and wait until it is processed.

packages/datadog_flutter_plugin/jnigen.yaml
```diff
@@ -18,4 +18,7 @@ classes:
   - 'com.datadoghq.flutter.DatadogRumPlugin'
   - 'com.datadoghq.flutter.DatadogLogsPlugin'
   - 'com.datadoghq.flutter.DatadogLogEventMapper'
-  
\ No newline at end of file
+  - 'com.datadoghq.flutter.DatadogSdkPlugin'
+  - 'com.datadog.android.api.context.DatadogContext'
+  - 'com.datadog.android.api.context.UserInfo'
+  - 'com.datadog.android.api.context.AccountInfo'
```
0xnm:
```suggestion
  - 'com.datadog.android.api.context.AccountInfo'

```

packages/datadog_flutter_plugin/lib/src/datadog_sdk_method_channel.dart
```diff
@@ -8,13 +8,29 @@ import 'package:flutter/foundation.dart';
...
+  LogCallback? _logCallback;
+
+  // Used by iOS, not by Andoid at the moment
+  DatadogContext? _cachedContext;
+  @override
+  DatadogContext? get cachedContext {
+    if (Platform.isIOS) {
+      return _cachedContext;
+    } else if (Platform.isAndroid) {
+      return AndroidDatadogFlutterPlugin.getContext();
```
0xnm, line 29: we can probably also cache and invalidate it when `setUser`/`setAccount`/`addExtraInfo` calls are made

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
```diff
@@ -473,7 +486,12 @@ void main() {
     final server = Server.create(services: [loggingService]);
     await server.serve(port: port);
 
+    mockPlatform = DatadogPlatformMock();
+    mockPlatform = DatadogPlatformMock();
     mockDatadog = DatadogSdkMock();
+    when(() => mockDatadog.platform).thenReturn(mockPlatform);
+
+    when(() => mockDatadog.platform).thenReturn(mockPlatform);
```
0xnm: things are repeated twice here

packages/datadog_flutter_plugin/test/tracing/tracing_headers_test.dart
```diff
@@ -371,16 +603,129 @@ void main() {
...
+      // Then
+      final baggageValues = newBaggage.split(',');
+      expect(baggageValues, contains('session.id=$sessionId'));
+      expect(baggageValues, contains('user.id=${datadogContext.userId}'));
+      expect(baggageValues, contains('account.id=${datadogContext.accountId}'));
+    });
+
+    test('existing baggage header adds new values', () {
+      // Given
+      String oldBaggage = 'test_value_1=1,test_value_2=string';
```
0xnm, line 648: FYI @marco-saia-datadog recently also added few tests https://github.com/DataDog/dd-sdk-reactnative/pull/1016/files#diff-395eac9b3950188084fb673149f381453a4f7fee342e91ed57f36ae94f0c3718, so maybe it is worth checking them as well.

0xnm (dismissed):

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
fuzzybinary: Do I have another option?

packages/datadog_flutter_plugin/lib/src/datadog_sdk_method_channel.dart
fuzzybinary: That assumes those are the only ways for someone to set the user / account / extraInfo. It's unfortunately a bad assumption in hybrid apps.

packages/datadog_flutter_plugin/android/src/main/kotlin/com/datadoghq/flutter/DatadogSdkPlugin.kt
0xnm: I don't think so. That is why implementing caching is very beneficial.

packages/datadog_flutter_plugin/lib/src/datadog_sdk_method_channel.dart
fuzzybinary: Unless you're saying Android should cache them?

0xnm: No, I'm talking about the Flutter methods. Android won't cache anything here, because `DatadogContext` includes many properties.

packages/datadog_flutter_plugin/test/tracing/tracing_headers_test.dart
fuzzybinary: I've updated this to do the same amount of validation, but I have pending decision. I'll hold off on merging this until I get some clarification.

fuzzybinary:
@0xnm this is ready for re-review. I do think we want to take the listener approach in the future but we'll stick with the context stall for now.

packages/datadog_grpc_interceptor/test/datadog_grpc_interceptor_test.dart
```diff
@@ -473,7 +486,11 @@ void main() {
     final server = Server.create(services: [loggingService]);
     await server.serve(port: port);
 
+    mockPlatform = DatadogPlatformMock();
+    mockPlatform = DatadogPlatformMock();
```
0xnm: minor: the same line is repeated twice

packages/datadog_flutter_plugin/lib/src/tracing/baggage_helpers.dart
```diff
@@ -0,0 +1,172 @@
...
+      sb.write(str);
+    }
+  }
+  return sb.toString();
+}
+
+String _percentDecode(String value) {
+  // This looks for percent encoded values and replaces them with their decoded
+  // values, leaving stand alone percents.
+  final escapeRE = RegExp(r'(?:%[\da-fA-F]{2})+');
```
0xnm: should this be compiled and stored in a class/static property? seems there is no need to create a new instance each time given it is stateless.

packages/datadog_flutter_plugin/lib/src/tracing/baggage_helpers.dart
```diff
@@ -0,0 +1,172 @@
...
+  return value.replaceAllMapped(escapeRE, (m) => Uri.decodeComponent(m[0]!));
```
0xnm: Is it safe to do `!` here? Should a check for non-empty `m` be made before?

0xnm (dismissed):

packages/datadog_flutter_plugin/lib/src/tracing/baggage_helpers.dart
fuzzybinary: I'm pretty sure the `!` is safe here. The mapping function shouldn't be called unless there's a match, which guarantees that there's an item in the first portion of the match. `m[0]` should only return `null` if there was no match when calling `RegExp.match`.

packages/datadog_flutter_plugin/lib/src/tracing/baggage_helpers.dart
fuzzybinary: Yeah, I can move it up to be a global of the file.

0xnm (approved):

PR #891, opened 2025-11-13
fix(sr): Fix to issues around clipping

fuzzybinary:
### What and why?

There were two issues in recent replays around proper capture of widgets. The first was we were not capturing `BoxShape.circle` if it was set for BoxDecoration. 

The second is that text was overflowing in the player past the bounds of its containing wireframes. Since Flutter does a lot of the clipping, setting `paintBounds` appropriately as it goes, we only need to make sure that the clipping of the paint bounds of text elements is respected by setting a clip of zero (because SR uses `insets` for clip-path).

However, the SR player will also ignore clip paths set to all zeros, so at least one of the elements needs to be set to 1 for it to take. We set `top` to one as I believe it is the least likely to result in rendering artifacts.

### Review checklist


0xnm (approved):

fuzzybinary:
Closing as this is going to be fixed on the player side. I'll reopen with just the `BoxShape.circle` fix.

PR #903, opened 2025-11-24
fix: Fix issue re-initializing platform channel platform

fuzzybinary:
### What and why?

Normally, the DatadogSdk should not be reinitialized, but the e2e tests are an exception. Doing this reinitialization was causing an exception when calling `listen` multiple times on the isolate communication stream.

Make a minor change where the communication stream is initialized and listened to once as a member of the Datadog platform.

Also make a minor fix to have session broadcast pushed to the main thread on iOS.

### Review checklist


packages/datadog_flutter_plugin/lib/src/datadog_sdk_method_channel.dart
```diff
@@ -54,10 +54,16 @@ class DatadogSdkMethodChannel extends DatadogSdkPlatform {
     return null;
   }
 
+  // These are used to communicate the current configuraiton between isolates.
```
0xnm:
```suggestion
  // These are used to communicate the current configuration between isolates.
```

0xnm (dismissed):

0xnm (approved):

PR #907, opened 2025-12-09
feat: Support context from background isolates for http tracing

fuzzybinary:
### What and why?

Datadog baggage headers include certain context which was cached in the Dart layer.  Unfortunately, this makes using this context in background isolates impossible, which means if users are tracking network requests from background isolates, this context would not be included in the baggage headers.

So, Instead of relying on Dart held cached context, we utilize native platforms shared memory instead.

On Android, this just means asking for more context during the request.

On iOS, I've moved the Context listener to being a singleton which will cache it. Fetching it requires a specialized C method to bridge Flutter and Swift.

We have tests to ensure that this tracking works, but the default method of performing Http tracking (`HttpOverrides`) is also isolate specific. The current test creates a new client on the background isolate, because of this.

Being able to use HttpOverrides in background isolates is tracked as RUM-13278

refs: RUM-12874

### How?

### Review checklist


packages/datadog_flutter_plugin/lib/src/ios/ios_platform_bridge.dart
```diff
@@ -0,0 +1,53 @@
...
+final class _DatadogCContext extends ffi.Struct {
+  external ffi.Pointer<Utf8> sessionId;
+  external ffi.Pointer<Utf8> accountId;
+  external ffi.Pointer<Utf8> userId;
+}
+
+class IosPlatformBridge {
+  static DatadogContext? getContext() {
+    // Instead of using Objective-C interop, we're going to use straight C / Swift interop
+    // to avoid pulling in the Objective-C package as part of Core.
```
0xnm, line 20: SDK v3 has ObjC bindings in the same module as Swift API, there is no dedicated `DatadogObjC` anymore.

0xnm (dismissed):

packages/datadog_flutter_plugin/lib/src/ios/ios_platform_bridge.dart
fuzzybinary: This is referring to the Dart `objective_c` package.

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
```diff
@@ -9,6 +9,29 @@ import DatadogCrashReporting
...
+@_cdecl("flutterGetDatadogContext")
+func flutterGetDatadogContext() -> DatadogCContext {
+    let context = ContextMessageReceiver.shared.cachedContext
+    guard let context = context else {
+        return DatadogCContext(sessionId: nil, accountId: nil, userId: nil)
+    }
+
+    let cSessionID = context.sessionId?.withCString { strdup($0) }
+    let cAccountId = context.accountId?.withCString { strdup($0) }
+    let cUserId = context.userId?.withCString { strdup($0) }
```
gonzalezreal, line 29: These will leak if `flutterGetDatadogContext()` is called but the resulting `DatadogCContext` is not consumed.

Should we add documentation about caller responsibility for memory?

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
```diff
@@ -46,26 +69,33 @@ extension Datadog.Configuration {
...
-        self.channel = channel
     }
 
-    public func receive(message: DatadogInternal.FeatureMessage, from core: any DatadogInternal.DatadogCoreProtocol) -> Bool {
+    static let shared: ContextMessageReceiver = {
+        let instance = ContextMessageReceiver()
+        return instance
+    }()
+
+    public var cachedContext: Context?
```
gonzalezreal, line 87: As far as I understand, this is written from the message-bus serial queue, but read from arbitrary threads (background isolate) via `flutterGetDatadogContext()`, right?

Then we should make it thread-safe:

```swift
@ReadWriteLock
public var cachedContext: Context?
```

gonzalezreal (commented):
Looks good!
My only concerns are the thread-safety of the cached context and the memory management of the strings. Let me know your thoughts.

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
fuzzybinary: Yes, that's correct so that's a good call. Thanks!

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
fuzzybinary: Yes that's correct, I'll add the note.

gonzalezreal (approved):
🚀

PR #914, opened 2025-12-30
Fix: Pin DictionaryCoder to 1.2.0 for iOS 12 compatibility

### LINKED ISSUE #913: iOS build fails: DictionaryCoder version conflict with iOS 12 target
**Describe the issue**

Getting a build error on iOS with the DictionaryCoder dependency:

```
Target Integrity (Xcode): The package product 'DictionaryCoder' requires minimum platform version 13.0 for the iOS platform, but this target supports 12.0
```

The Package.swift file is using \`from: "1.2.0"\` for DictionaryCoder, which lets SPM pull in version 1.3.0. That version bumped the minimum iOS to 13, but we're targeting iOS 12.

**Reproduction steps**

1. Pull latest from develop
2. Try building for iOS
3. Build fails with the error above

**Device Information**

iOS build environment, Xcode 15+

**SDK version**

2.16.0

**Flutter Version**

3.35.0

**Setup Type**

Standard Flutter iOS build

**Other relevant information**

Few ways to fix this:
- Pin to exact 1.2.0 (quick, keeps iOS 12 support)
- Update min iOS to 13 (breaking change)

I'll submit a PR with the first option since it's the least disruptive, but wanted to get maintainers' input on the right approach here.

rcjuancarlosuwu:
### What and why?

Fixes #913

iOS builds are failing because DictionaryCoder 1.3.0 requires iOS 13+, but we're targeting iOS 12. The Package.swift was using \`from: "1.2.0"\` which lets SPM grab 1.3.0.

This pins it to exactly 1.2.0 to prevent the version bump.

### How?

Changed Package.swift dependency from:
```swift
.package(url: "https://github.com/almazrafi/DictionaryCoder.git", from: "1.2.0")
```

to:
```swift
.package(url: "https://github.com/almazrafi/DictionaryCoder.git", exact: "1.2.0")
```

This is the least invasive option - keeps iOS 12 support without requiring code changes. Other options would be bumping to iOS 13 (breaking) or updating to DictionaryCoder 1.3.0 (has breaking API changes).

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Package.swift
+++ b/packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/Datadog/dd-sdk-ios.git", branch: "develop"),
-        .package(url: "https://github.com/almazrafi/DictionaryCoder.git", from: "1.2.0")
+        .package(url: "https://github.com/almazrafi/DictionaryCoder.git", exact: "1.2.0")
     ],
     targets: [
         .systemLibrary(name: "datadog_flutter_plugin_c"),


fuzzybinary:
Hi @rcjuancarlosuwu,

Thanks for the fix. This is in line with what we do with the Podfile, I just used the wrong keyword when pinning it in the SPM version, so this change looks good to me. I'll try to get it up into a patch release of 2.0 ASAP.

Just a heads up, this might be a ways off for you but starting with Flutter 3.35, the minimum version of iOS for Flutter will be 13, and our example applications [now use 13+](https://github.com/DataDog/dd-sdk-flutter/commit/afe9bd7df2acc081add0cf9c2045139d6d6a0ab2).

fuzzybinary (approved):

fuzzybinary:
Bypassing to get around an unrelated Android build error.

PR #926, opened 2026-01-12
fix: Fix crash with ios webview tracking in Flutter 3.38+

fuzzybinary:
### What and why?

Flutter 3.38+ changes where plugins are registered, using the ViewController over the AppDelegate as the registrar / registry.

FWFWebViewFlutterWKWebViewExternalAPI also performs a force cast (`as!`) to get the plugin from the registry, which crashes on iOS.

While Flutter 3.38 adds a `viewController` property to FlutterPluginRegistrar to more easily access the `PluginRegistry`, this property is not available for our min versions (3.27). So to fix the crash we're taking two steps:

First, we ask for the `viewController` from the `FlutterAppDelegate`, which should be valid for both 3.27+ and 3.28. Just in case as a final fallback, we use the AppDelegate as the plugin registry.

Second, we added a check that prevents the force cast `as!`. We look to see if the plugin is registered *before* we ask the external API for it. This way, if the `PluginRegistry` we're using isn't the one that the WebView plugin used, we won't crash on the force cast. We'll just miss out on tracking the given webview.

### Review checklist

### DIFF
--- a/packages/datadog_webview_tracking/ios/datadog_webview_tracking/Sources/DatadogWebviewTrackingPlugin.swift
+++ b/packages/datadog_webview_tracking/ios/datadog_webview_tracking/Sources/DatadogWebviewTrackingPlugin.swift
@@ -4,6 +4,8 @@
 
 import Flutter
 import UIKit
+import DatadogCore
+import DatadogInternal
 import DatadogWebViewTracking
 import webview_flutter_wkwebview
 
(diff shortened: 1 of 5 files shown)


packages/datadog_webview_tracking/ios/datadog_webview_tracking/Sources/DatadogWebviewTrackingPlugin.swift
```diff
@@ -36,17 +38,28 @@ public class DatadogWebViewTrackingPlugin: NSObject, FlutterPlugin {
...
+                        id: "webview_tracking_init_failed",
+                        message: "Failed to initialie WebViewTracking because a WebViewFlutterPlugin instance was not found",
+                        kind: "DependencyFailure",
+                        stack: nil
+                    )
                 }
                 result(nil)
             } else {
+                consolePrint(
+                    "⚠️ Could not find WebViewFlutterPlugin to enable Datadog tracking. This may be because of a change in Flutter." +
```
0xnm:
```suggestion
                    "⚠️ Could not find WebViewFlutterPlugin to enable Datadog tracking. This may be because of a change in Flutter. " +
```

0xnm (dismissed):

awforsythe (approved):

Issue #932, opened 2026-01-15
[Android] Hot restart crash: "Callback invoked after it has been deleted" in PortProxyBuilder (v3.0.0)


dfdgsdfg:
## Description

After upgrading to `datadog_flutter_plugin` v3.0.0, hot restart on Android causes a fatal crash with the error:

```
Callback invoked after it has been deleted.
```

The crash originates from `Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke`.

## Environment

- **Flutter version**: 3.38.6 (stable)
- **datadog_flutter_plugin version**: 3.0.0
- **jni version**: 0.14.2 (transitive dependency)
- **Platform**: Android (tested on Samsung device, Android 16 API 36)
- **datadog_session_replay**: 1.0.0-preview.8

## Steps to Reproduce

1. Add `datadog_flutter_plugin: ^3.0.0` to pubspec.yaml
2. Initialize DatadogSdk in your app
3. Run the app on an Android device in debug mode
4. Perform a hot restart (press `r` in terminal)
5. App crashes immediately

## Expected Behavior

Hot restart should work without crashing.

## Actual Behavior

App crashes with a fatal signal (SIGABRT) and the following stack trace:

```
E/Dart    (  800): ../../../flutter/third_party/dart/runtime/vm/runtime_entry.cc: 5113: error: Callback invoked after it has been deleted.
E/DartVM  (  800): version=3.10.7 (stable)
E/DartVM  (  800): pc 0x0000006d9f621fd4 fp 0x0000007ffda15a00 libflutter.so+0x1ff9fd4
E/DartVM  (  800): pc 0x0000006d9f50fec0 fp 0x0000007ffda15b98 libflutter.so+0x1ee7ec0
E/DartVM  (  800): pc 0x0000006d9ae6b590 fp 0x0000007ffda15c80 Java_com_github_dart_1lang_jni_PortProxyBuilder__1invoke+0x2fc
F/libc    (  800): Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE)
Abort message: '../../../flutter/third_party/dart/runtime/vm/runtime_entry.cc: 5113: error: Callback invoked after it has been deleted.'
```

## Root Cause Analysis

Version 3.0.0 introduced "Support event mappers through JNI on Android" which brings in `jni 0.14.2`. The JNI `PortProxyBuilder` is used for callbacks from Java to Dart.

During hot restart:
1. Dart objects are recreated
[...]

PR #934, opened 2026-01-15
fix(sr): Utilize a singleton bridge for SessionReplay on Android.

fuzzybinary:
### What and why?

During Hot Restart, Flutter will re-run main and recreate all of the SessionReplay objects, which results in the Context callback we're using in Session Replay to invalid, and resulting in a crash when RUM tries to tell SR about the new context.

To prevent the crash, and to fix the fact that during Hot Restart, we are improperly re-registering the SR feature on Android, we've switched the FlutterSessionReplayBridge to a single instance.

Additionally, when `enable` is called, we replace the context listener with the newly configured one.

refs: #932

### Review checklist


packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
```diff
@@ -9,10 +9,18 @@ package com.datadoghq.flutter.sessionreplay
...
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadoghq.flutter.sessionreplay.feature.DefaultFlutterSessionReplayFeature
-import com.datadoghq.flutter.sessionreplay.feature.FlutterSessionReplayFeature
 import java.nio.ByteBuffer
 
 internal class FlutterSessionReplayBridge {
+    companion object {
+        private var instance: FlutterSessionReplayBridge = FlutterSessionReplayBridge()
+
+        @JvmStatic
```
0xnm: This is not needed if only callers of this are from Kotlin.

upd: it seems it is needed for generator?

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
```diff
@@ -9,10 +9,18 @@ package com.datadoghq.flutter.sessionreplay
 import com.datadog.android.Datadog
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadoghq.flutter.sessionreplay.feature.DefaultFlutterSessionReplayFeature
-import com.datadoghq.flutter.sessionreplay.feature.FlutterSessionReplayFeature
 import java.nio.ByteBuffer
 
 internal class FlutterSessionReplayBridge {
+    companion object {
+        private var instance: FlutterSessionReplayBridge = FlutterSessionReplayBridge()
```
0xnm: You can simply do `internal object FlutterSessionReplayBridge` is you want to have a singleton, no need for the companion object. And then from Java it is visible as `FlutterSessionReplayBridge.INSTANCE`.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
fuzzybinary: Yes. The main caller is Dart using JNI as a bridge, so having it be JvmStatic is easier.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
fuzzybinary: 👍 I'll make that change.

0xnm (approved):

Issue #940, opened 2026-02-03
RUM resource size reports as 0b when responses are chunked/streamed (no Content-Length)


vincevargadev:
### Feature description

Some of the HTTP requests that we track with the Flutter Datadog SDK report `Resource size 0b`. This causes visibility issues in our dashboards: We would like to see all requests' size so that we can spot large responses, and we can take action and improve our app.

I checked the implementation and it looks like the `tracking_http_client` only reports `Content-Length` and if that is not available, the request size will show up as 0b.

We would like to see the response size in the Datadog web UI for all requests are app makes.

### Proposed solution

Use `Content-Length` when available, manually count response bytes otherwise, and report that value as resource size.

Rough draft to clarify idea:

```dart
int _bytesReceived = 0;

return innerResponse.listen(
  (data) {
    _bytesReceived += data.length;
    if (onData != null) onData(data);
  },
  // ...
);

final size = innerResponse.contentLength > 0
    ? innerResponse.contentLength
    : (_bytesReceived > 0 ? _bytesReceived : null);

rum.stopResource(rumKey!, statusCode, resourceType, size, attributes);
```

fuzzybinary:
This seems reasonable to me. I'll see if we can add it. I'd also be open to a PR if you're so inclined.

vincevargadev:
@fuzzybinary I opened a PR, let me know there what you think about the changes. Thank you. https://github.com/DataDog/dd-sdk-flutter/pull/945/changes

vincevargadev:
The PR addressing this issue was merged into `develop` https://github.com/DataDog/dd-sdk-flutter/pull/945

@fuzzybinary Do you know when you will have a new release? https://pub.dev/packages/datadog_flutter_plugin/changelog

fuzzybinary:
`datadog_tracking_http_client 3.1.0` and `datadog_dio 2.1.0` went out today with these changes.

Let me know if you see any issues!

PR #945, opened 2026-02-11
RUM Add "manual" resource size tracking when content-length is not available (Issue #940)

vincevargadev:
### What and why?

Track resource size "manually" (tracking incoming stream total byte size) when content-length is not available.

Updates `datadog_dio` and `datadog_tracking_http_client`.

### How?

When the content length is not available, track the size of the response manually.

### Review checklist

- [x] This pull request references a Github or JIRA issue #940

### DIFF
--- a/packages/datadog_dio/lib/src/datadog_dio_interceptor.dart
+++ b/packages/datadog_dio/lib/src/datadog_dio_interceptor.dart
@@ -4,6 +4,7 @@
 
 // ignore_for_file: invalid_use_of_internal_member
 
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:datadog_flutter_plugin/datadog_flutter_plugin.dart';
@@ -92,6 +93,20 @@ class DatadogDioInterceptor extends Interceptor {
     super.onError(err, handler);
   }
 
+  /// Returns byte length of response body when already in memory (e.g. chunked response).
+  static int? _responseDataByteLength(dynamic data) {
+    if (data is List<int>) {
+      return data.length;
+    }
+    if (data is Uint8List) {
+      return data.length;
+    }
+    if (data is String) {
+      return utf8.encode(data).length;
+    }
+    return null;
+  }
+
   void _stopWithResponse(
       DatadogRum rum, String rumKey, Response<dynamic> response) {
     final contentTypeHeader =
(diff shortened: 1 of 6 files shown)


packages/datadog_dio/lib/src/datadog_dio_interceptor.dart
```diff
@@ -92,6 +93,21 @@ class DatadogDioInterceptor extends Interceptor {
...
+    if (data is List<int>) {
+      return data.length;
+    }
+    if (data is Uint8List) {
+      return data.length;
+    }
+    if (data is String) {
+      // This could be expensive at such a low level library?
+      return utf8.encode(data).length;
+    }
```
fuzzybinary, line 106: It also makes an assumption that the data was encoded as UTF-8 when it came in. This should be okay since Dio's default transformer is the Utf8Decoder.

fuzzybinary (commented):
This LGTM -- can you [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and amend your commit message to follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) ?

packages/datadog_dio/lib/src/datadog_dio_interceptor.dart
vincevargadev: Should we wrap it in a try catch, and fallback back to `null` if it errs? Don't want to cause any unexpected issues to the users of this package. (though probably my entire career I only used utf8, but who knows)

fuzzybinary: Yes, that's probably a good idea.

It is a wonder that Dart Strings are held UTF-16 huh? 😅

vincevargadev: @fuzzybinary I just thought about it a bit more, I am not sure wrappin it in a try-catch makes sense...

> Any unpaired surrogate character (U+D800-U+DFFF) in the input string is encoded as a Unicode Replacement character U+FFFD (�).  [Source](https://api.flutter.dev/flutter/dart-convert/Utf8Encoder/convert.html)

So with unpaired surrogates not throwing, I don't know of any valid UTF16 Dart string that would throw.

Can you come up with a scenario where a valid Dart string (which is guaranteed, because we check if it's a string) can cause an exception to be thrown?

vincevargadev:
@fuzzybinary My changes should be now in a single, signed, conventional commit: https://github.com/DataDog/dd-sdk-flutter/pull/945/changes/a166e8da892e497ad8647bd8bab978cf01893b63

Let me know if there is anything else I need to make this PR mergeable. Thank you!

fuzzybinary (approved):

fuzzybinary:
Bypassing rules as the build failures are not related and fixed in #949

PR #948, opened 2026-02-16
fix: Nested GestureDetector should take precedence over parent InkWell

### LINKED ISSUE #740: Nested GestureDetector should take precedence over parent InkWell
### Describe the bug

When theres a GestureDetector down the widget tree which has its onTap set, it takes precedence over the InkWell higher up the tree. But this is not the case when it comes to RumUserActionAnnotation 👇🏼 

### Reproduction steps

```dart
final class TestWidget extends StatelessWidget {
  const TestWidget({super.key});

  @override
  Widget build(BuildContext context) {
    return RumUserActionAnnotation(
      description: 'TestWidget',
      attributes: {'placement': 'parent'},
      child: InkWell(
        onTap: () => 'parent tapped',
        child: Container(
          width: 100,
          height: 100,
          color: Colors.white,
          alignment: Alignment.center,
          child: Column(
            mainAxisSize: MainAxisSize.min,
            children: [
              Text('parent'),
              RumUserActionAnnotation(
                description: 'TestWidget - Child',
                attributes: {'placement': 'child'},
                child: GestureDetector(
                  onTap: () => 'child tapped',
                  child: Text('child'),
                ),
              ),
            ],
          ),
        ),
      ),
    );
  }
}
```

### SDK logs

_No response_

### Expected behavior

_No response_

### Affected SDK versions

2.10.2

### Latest working SDK version

_No response_

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Flutter Version

_No response_

### Setup Type

_No response_

### Device Information

_No response_

### Other relevant information

_No response_

brunovsiqueira:
### What and why?

Fixes #740

When a `GestureDetector` with its own `RumUserActionAnnotation` is nested inside an `InkWell`, it should take precedence over the parent `InkWell` when reporting user actions to RUM. Previously, the `InkWell` would always be reported because the `betterThan()` method treated `GestureDetector` as the "worst" option that could never replace other widgets.

This is a common pattern when users want separate tappable areas within an `InkWell`, each with their own tracking annotations.

### How?

Modified the `betterThan()` method in `_ElementDescription` class to allow a `GestureDetector` to take precedence over an `InkWell` **only when they have different annotations**. This distinguishes between:

1. **User-created nested GestureDetector** (has its own annotation → different description) - should take precedence
2. **InkWell's internal GestureDetector** (inherits parent annotation → same description) - should NOT take precedence

The comparison uses the `elementDescription` field which contains the annotation text, ensuring that only intentionally annotated nested `GestureDetector`s override their parent `InkWell`.

### Review checklist

### Tests Added

| Test | Scenario | Expected Result |
|------|----------|-----------------|
| `nested GestureDetector inside InkWell uses inner GestureDetector` | InkWell + child annotation + GestureDetector | Reports GestureDetector ✓ |
| `nested GestureDetector inside InkWell without annotation uses InkWell` | InkWell + unannotated GestureDetector | Reports InkWell ✓ |
| `nested GestureDetector inside InkWell reports correct attributes` | Verifies attributes pass through | Correctly reported ✓ |

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/rum_user_action_detector.dart
@@ -39,9 +39,18 @@ class _ElementDescription {
       return true;
     }
 
-    // Literally anything is better than GestureDetector...
+    // GestureDetector and InkWell are both generic tappable containers.
+    // A GestureDetector with its own annotation (different from the parent
+    // InkWell's annotation) should take precedence, indicating the user
+    // intentionally created a separate tappable area. However, if they share
+    // the same annotation, prefer InkWell as the GestureDetector is likely
+    // internal or unintentional.
     if (element.widget is GestureDetector) {
-      // ... except another GestureDetector
+      if (other.element.widget is InkWell &&
+          elementDescription != other.elementDescription) {
+        return true;
+      }
+      // Otherwise, GestureDetector only replaces another GestureDetector
       return other.element.widget is GestureDetector;
     }
 
--- a/packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
+++ b/packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
@@ -259,6 +259,125 @@ void main() {
     verifyNoMoreInteractions(mockRum);
   });
 
+  testWidgets(
+      'nested GestureDetector inside InkWell uses inner GestureDetector',
+      (tester) async {
+    final mockRum = MockDdRum();
+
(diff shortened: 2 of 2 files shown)


fuzzybinary (commented):
Looks good! Thanks for tackling this @brunovsiqueira !

Can you make sure you've [signed your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and I'll get this onto CI.

brunovsiqueira:
Hi @fuzzybinary 
It's done :)

fuzzybinary:
Bypassing rules as build failures are unrelated and fixed in #949

PR #977, opened 2026-03-25
Reproduce RUMS-5633: images not appearing in Session Replay

mariusc83:
## Reproduction for RUMS-5633

**Jira:** [RUMS-5633](https://datadoghq.atlassian.net/browse/RUMS-5633)

### Issue Summary
Customer reports PNG images from `Image.asset()` are not appearing in Session Replay despite `imagePrivacyLevel` being set to `ImagePrivacyLevel.maskNone`.

### Reproduction Tests
- Unit tests: 3 (in `packages/datadog_session_replay/test/capture/image_asset_reproduction_test.dart`)
- Integration tests: 0

### What the Tests Prove

1. **Default config masks all images** (passes, confirming expected behavior): `DatadogSessionReplayConfiguration` defaults `imagePrivacyLevel` to `maskAll`. If the customer's `maskNone` override doesn't propagate correctly, all images appear as placeholders.

2. **Null `resourceId` causes missing images** (fails): When the native platform's `resourceIdForKey()` returns `null` (due to processing failure or timing issues), `SRImageWireframe` is created with `resourceId: null`. The Session Replay player cannot render an image without a valid `resourceId`, causing images to "not appear."

3. **`_extractAssetImage` doesn't handle all asset providers** (fails): With `maskNonAssetsOnly`, image providers that are not exactly `AssetImage` or `ResizeImage(AssetImage)` — such as `ExactAssetImage` (used by `Image.asset(scale: ...)`) — are incorrectly treated as non-asset images and masked with placeholders.

### Failure Output
```
Test 2: Expected resourceId to be non-null, but SRImageWireframe.resourceId is null
         when platform.resourceIdForKey returns null.

Test 3: Expected no placeholder wireframes, but _extractAssetImage fails to recognize
         non-AssetImage providers, causing RawImage to be masked with placeholder.
```

---
*Generated by rum:tee-triage-insights*

[RUMS-5633]: https://datadoghq.atlassian.net/browse/RUMS-5633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/datadog_session_replay/test/capture/image_asset_reproduction_test.dart
+++ b/packages/datadog_session_replay/test/capture/image_asset_reproduction_test.dart
@@ -36,6 +36,7 @@ import 'package:datadog_session_replay/src/capture/recorder.dart';
 import 'package:datadog_session_replay/src/datadog_session_replay_platform_interface.dart';
 import 'package:datadog_session_replay/src/rum_context.dart';
 import 'package:datadog_session_replay/src/sr_data_models.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_test/flutter_test.dart';
--- a/packages/datadog_session_replay/test/capture/image_asset_reproduction_test.dart
+++ b/packages/datadog_session_replay/test/capture/image_asset_reproduction_test.dart
@@ -1,0 +1,389 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+// Reproduction tests for RUMS-5633: Images not appearing in Session Replay.
+//
+// Customer reports that png images from Image.asset() on regular Flutter screens
+// are not appearing in Session Replay, even with imagePrivacyLevel set to
+// ImagePrivacyLevel.maskNone.
+//
+// These tests target two plausible failure modes:
+//
+// 1. The default imagePrivacyLevel is maskAll (not maskNone). If the customer's
+//    configuration isn't applied correctly, images are masked by default.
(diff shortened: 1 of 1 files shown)


fuzzybinary:
This one's pretty good! 100% the `ExactAssetImage` is a miss. If the customer is complaining about assets missing, than this is the cause (and it looks like that's the case here).

The second issue it brings up -- it doesn't really address what the root cause of the null resource id is, right? If the platform returns a null resource id, something went wrong is processing, so *what* went wrong is the root cause, which it didn't analyze. It's not enough here to say "guard against null resourceIds". The Player does guards against null resource ids by supplying a placeholder.

mariusc83:
> This one's pretty good! 100% the `ExactAssetImage` is a miss. If the customer is complaining about assets missing, than this is the cause (and it looks like that's the case here).
> 
> The second issue it brings up -- it doesn't really address what the root cause of the null resource id is, right? If the platform returns a null resource id, something went wrong is processing, so _what_ went wrong is the root cause, which it didn't analyze. It's not enough here to say "guard against null resourceIds". The Player does guards against null resource ids by supplying a placeholder.

On the second part, I would argue that an image with null resource id will still create a placeholder and that might still be a cause of the problem right ?

fuzzybinary:
Yes, but the problem isn't that a null resource id creates a placeholder (that's expected), the problem is that we got a null resource id in the first place.

mariusc83:
> Yes, but the problem isn't that a null resource id creates a placeholder (that's expected), the problem is that we got a null resource id in the first place.

That's correct. So the approach should have been to check the problem at the bottom in the native layer. The problem though is that this could happen in the client's app randomly. Depending on what they're using for creating images/assets. So it would have been really hard to handle that in a test. So I would say the test it added here is still valuable as it is showing you a case where this could happen.

PR #978, opened 2026-03-25
fix(sr): Fix a lifecycle issue with Android calling a dead callback on activity destruction.

fuzzybinary:
### What and why?

Fixing a crash in Session Replay caused when a context changed callback is received after the main Flutter Engine has detached. This happens when a user backs out of the first screen of an application, which closes the MainActivity, killing the Flutter Engine, but keeping the Application (and all FFI / JNI singletons) alive.

Because we've seen that multiple Flutter engines are pretty wide spread, we've gone through lengths to make sure we don't accidentally attach to the lifecycle of the wrong engine.

Flutter doesn't give FFI plugins a way to access the lifecycle of the current engine, or a way to figure out the currently executing context or isolate, so in order to make sure we connect the correct "main" engine with Session Replay, and only dereference the context callback when that engine is detached / removed, we utilize a method channel call to make the connection.

The method channel call is creates a very slight chance of a race condition if the engine were to quickly detach after the `enable` call but before the method channel can process the `claimOwnership` method channel call, but this is incredibly unlikely.

We perform this attachment for both iOS and Android, although the likelyhood of iOS performing the detach is much lower than Android (which will detach if the MainActivity is closed, but not the application). The hope is that this potentially solves a "crash on application termination" that might occur if an app is force killed while a context update is being broadcast.

refs: RUMS-5684 RUMS-5712 RUMS-5705

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
+++ b/packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
@@ -37,7 +37,9 @@ internal object FlutterSessionReplayBridge {
         val onContextChanged: ContextListener
     )
 
+    @Volatile
     var contextListener: ContextListener? = null
+
     var feature: DefaultFlutterSessionReplayFeature? = null
     internal var listenerOwner: BinaryMessenger? = null
 
--- a/packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
+++ b/packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
@@ -9,6 +9,7 @@ package com.datadoghq.flutter.sessionreplay
 import com.datadog.android.Datadog
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadoghq.flutter.sessionreplay.feature.DefaultFlutterSessionReplayFeature
+import io.flutter.plugin.common.BinaryMessenger
 import java.nio.ByteBuffer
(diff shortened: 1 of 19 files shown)


fuzzybinary:
@codex review

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/FlutterSessionReplayBridge.kt
```diff
@@ -37,6 +39,11 @@ internal object FlutterSessionReplayBridge {
 
     var contextListener: ContextListener? = null
     var feature: DefaultFlutterSessionReplayFeature? = null
+    internal var listenerOwner: BinaryMessenger? = null
```
mariusc83, line 44: shouldn't this be declared as @Volatile ? also maybe make it an `AtomicReference` instead to make sure of atomicity ? Again I am not certain how this system works so this is only needed in case there's a multi-threading access here.

fuzzybinary: It's a good callout, but no we should be fine. Dart is essentially single threaded, and this is only needed during initialization and shutdown, which should both happen on the Dart mutator thread (which as of Dart 3.29 is now the main thread).

The contextListener however.... 🤔

mariusc83: yeah that's what I was worried about...

fuzzybinary: Switched `contextListener` to Volatile. That should be enough to deal with thread contention.

aleksandr-gringauz (approved):

PR #989, opened 2026-04-03
feat(sr): Add CheckboxRecorder to capture Checkbox widgets properly (cupertino checkbox included)

JuanNaranjoDD:
### What and why?

Adds CheckboxRecorder, a new ElementRecorder implementation that captures Flutter Checkbox widgets in Session Replay. Without this recorder, checkboxes were matched by CustomPaintRecorder and rendered as a SRPlaceholderWireframe, showing a generic box with an × inside regardless of their actual state. This ensures checked, unchecked, and tristate states are faithfully represented in replays.

### How?

CheckboxRecorder resolves the widget's visual properties (fill color, checkmark color, border) by walking the widget properties and theme fallback chain, matching Flutter's own rendering logic. It produces a single SRTextWireframe per checkbox: the box shape is rendered via shapeStyle/border, and the checkmark symbol (✓, —, or empty) is centered as its text. The subtree is ignored to prevent CustomPaintRecorder from double-capturing the inner CustomPaint.

### Review checklist

- [x] This pull request references a Github or JIRA issue — RUM-10251


packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -149,6 +150,7 @@ class SessionReplayRecorder {
       ImageRecorder(keyGenerator),
       CustomPaintRecorder(keyGenerator),
       PrivacyRecorder(keyGenerator),
+      CheckboxRecorder(keyGenerator),                     // Support for rendering properly checkbox widget on SR
```
fuzzybinary: This comment isn't particularly useful....

packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart
```diff
@@ -0,0 +1,688 @@
...
+      // Then
+      expect(capture, isNotNull);
+      final treeCapture = capture!.viewTreeSnapshot;
+      expect(treeCapture.nodes.length, 1);        // It should be just one node, the checkbox node
+      final checkboxNode = treeCapture.nodes.first;
+
+      final wireframes = checkboxNode.buildWireframes();
+      expect(wireframes.length, 1);
+      final checkboxWireframe = wireframes.first as SRTextWireframe;
+      expect(checkboxWireframe.text, '\u2713');
```
fuzzybinary: Since Flutter uses UTF-16 encoded strings, and we can add unicode characters to source files, it might be worth replacing this with the actual checkmark character. It would make this more readable.

packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart
```diff
@@ -0,0 +1,688 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+
+// Note: to properly test recorders, we need to supply a full widget tree, as Element
+// is too difficult to mock effectively.
```
fuzzybinary: We should remove this comment and instead add it as a comment to `SimpleTestCapture`. I know I put this here in other files but we should take this opportunity to fix it.

packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart
```diff
@@ -0,0 +1,688 @@
...
+      final capture = await recorder.performCapture();
+
+      // Then
+      expect(capture, isNotNull);
+      final treeCapture = capture!.viewTreeSnapshot;
+      final checkboxNode = treeCapture.nodes.first;
+      final wireframes = checkboxNode.buildWireframes();
+      final checkboxWireframe = wireframes.first as SRTextWireframe;
+      // Colors.blue (#ff2196f3)
+      expect(checkboxWireframe.shapeStyle!.backgroundColor, '#2196f3ff');
```
fuzzybinary: The comment and the string here don't match?

packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart
```diff
@@ -0,0 +1,688 @@
...
+      final treeCapture = capture!.viewTreeSnapshot;
+      final checkboxNode = treeCapture.nodes.first;
+      final wireframes = checkboxNode.buildWireframes();
+      final checkboxWireframe = wireframes.first as SRTextWireframe;
+      expect(checkboxWireframe.shapeStyle!.backgroundColor, '#00000000');
+
+    });
+
+    testWidgets('tristate checkbox has non-transparent fill', (tester) async {
+      // Given
```
fuzzybinary: For all of these non-transparent fill tests, it's likely better to check the expected value instead of it being non-transparent.

Technically, also, testing non-transparency would required parsing  color and checking the alpha value, as you can have non-zero values for the color components and zero on the alpha.

packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart
```diff
@@ -0,0 +1,688 @@
...
+      final wireframes = checkboxNode.buildWireframes();
+      final checkboxWireframe = wireframes.first as SRTextWireframe;
+      // Colors.blue (#ff2196f3) with alpha 0.38 --> round(0.38 * 255) = 97 = 0x61
+      expect(checkboxWireframe.shapeStyle!.backgroundColor, '#2196f361');
+    });
+  });
+
+  group('border', () {
+
+    testWidgets('unchecked checkbox has a border', (tester) async {
```
fuzzybinary: This is fine, but I also would have been okay checking the border with the checkmark in the same test.

packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart
```diff
@@ -0,0 +1,688 @@
...
+      final capture = await recorder.performCapture();
+
+      // Then
+      expect(capture, isNotNull);
+      final treeCapture = capture!.viewTreeSnapshot;
+      final checkboxNode = treeCapture.nodes.first;
+      final wireframes = checkboxNode.buildWireframes();
+      final checkboxWireframe = wireframes.first as SRTextWireframe;
+      // Colors.red (#fff44336) with full alpha --> #f44336ff
+      expect(checkboxWireframe.border!.color, '#f44336ff');
```
fuzzybinary: For readability and to avoid these comment disconnects, it might make sense to instead use `Colors.red.toHexString`. We have `toHexString` tested elsewhere, so that might make things cleaner in tests.

packages/datadog_session_replay/example/golden_test/simple_widget_golden_test.dart
fuzzybinary, line 1: I would also add masked tests to this.  Additionally, since you are heavily relying on actual text here, see if you can load the test fonts and see if that show's the actual test symbols:  In `setupAll()` add:

```dart
await TestFonts.loadAppFonts();
```

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+import '../../../../datadog_session_replay.dart';
+import '../../../extensions.dart';
+import '../../../sr_data_models.dart';
+import '../../capture_node.dart';
+import '../../recorder.dart';
+import '../../view_tree_snapshot.dart';
+
+// Characters to represent the checkbox states
+const String checkmark = '\u2713';
+const String dash = '\u2014';
```
fuzzybinary: See if you can paste the actual unicode symbols to the file. It will make it way easier to read / understand.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+const String maskedSymbol = 'x';
+
+// Scale for the text size within the box
+const double textScale = 0.7;
+
+// Corner radius for the box.
+const double checkboxRadius = 2.0;
```
fuzzybinary: Should these be private?

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+      checkColor = widget.checkColor;
+      activeColor = widget.activeColor;
+      fillColor = widget.fillColor;
+      widgetSide = widget.side;
+      visualDensity = null;                         // CupertinoCheckbox does not has visualDensity property
+
+    } else {
+      // It must be never be reached
+      throw UnsupportedError('Unsupported widget type: ${widget.runtimeType}');
+    }
```
fuzzybinary: Don't throw here. While you're right, this shouldn't be reached, if something accidentally changes throwing here will causes SR to register a capture error. If it gets enough of them, it will shutdown. 

If we can, we should log this to telemetry, but regardless we should return `null` instead of throwing.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+    final Color? activeColor;
+    final WidgetStateProperty<Color?>? fillColor;
+    final BorderSide? widgetSide;
+    final VisualDensity? visualDensity;
+
+    Color backgroundColor;
+    Color symbolColor;
+    BorderSide? borderSide;
+
+
```
fuzzybinary: I'm not a huge fan of some of these extra spaces, especially around the blocks.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+    final node = CheckboxNode(
+          attributes,
+          wireframeId: wireframeKey,
+          value: value,
+          fillColor: backgroundColor,
+          symbolColor: symbolColor,
+          side: borderSide,
+          isMasked: isMasked,
+        );
```
fuzzybinary: Setup `dartfmt` to run on save. I'm pretty sure this should have been kicked back an indentation level given the trailing comma, and I'm seeing a few other things `dartfmt` should have caught.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+
+    if (!isMasked) {
+
+      // Resolve fill color: checked and tristate (value != false) use the active color,
+      // unchecked uses transparent since the border conveys the unchecked state instead.
+      backgroundColor = fillColor?.resolve(states) ??
+          ((value != false)
+            ? (activeColor ?? theme.colorScheme.primary)
+            : Colors.transparent);
+      if (!isEnabled && value != false) backgroundColor = backgroundColor.withValues(alpha: 0.38);
```
fuzzybinary: This is a lot of work to parse this multi-layered null-coallessing, ternary operator operation, which is potentially overridden by the `!isEnabled && value != false` check immediately after it. Can we look for a way to simplify it?

Additionally, you did a great job putting a lot of constants at the top of the file, but the `0.38` and `0.6` got missed, as did the default border side width (`2.0`)

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+      visualDensity = null;                         // CupertinoCheckbox does not has visualDensity property
+
+    } else {
+      // It must be never be reached
+      throw UnsupportedError('Unsupported widget type: ${widget.runtimeType}');
+    }
+
+    // Resolve checkbox theme for colors and border
+    final theme = Theme.of(element);
+    final checkboxTheme = theme.checkboxTheme;
```
fuzzybinary: If you move the theme capture up above the if, you can probably do a lot of the color resolution and visual density resolution to non-null concrete values in the above `if`s, instead of waiting until further down in the code.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,245 @@
...
+  // Renders the checkbox as a single SRTextWireframe: the box shape is drawn
+  // via shapeStyle/border, and the checkmark symbol is centered as text.
+  @override
+  List<SRWireframe> buildWireframes() {
+
+    final symbol = isMasked ? maskedSymbol : switch (value) {
+      true => checkmark,
+      false => '',
+      null => dash
+    };
```
fuzzybinary: You might be able to use straight pattern matching for this since we support Dart 3.6+.

```
 final symbol = switch ((isMasked, value)) {
    (true, _) => maskedSymbol,
    (_, true) => checkMark,
    (_, false) => '',
    (_, null) => dash,
};

File: packages/datadog_session_replay/test/capture/material_widgets/checkbox_recorder_test.dart [hunk 16]
```diff
@@ -0,0 +1,688 @@
...
+      final treeCapture = capture!.viewTreeSnapshot;
+      final checkboxNode = treeCapture.nodes.first;
+      final wireframes = checkboxNode.buildWireframes();
+      final checkboxWireframe = wireframes.first as SRTextWireframe;
+      expect(checkboxWireframe.text, '\u2713');
+    });
+  });
+
+  group('cupertino checkbox', () {
+    testWidgets('checked cupertino checkbox has checkmark symbol', (tester) async {
```
fuzzybinary: This isn't necessary, but something you could have done to ensure full coverage against both `Checkbox` and `CupertinoCheckbox`, especially if the expectations are very similar, would be to write a meta-tester which could test multiple widget setups with the same assertions. Something similar to:

```dart
void metaTestWidgets(String testDescription, List<Widget Function()> setups), void Function(CaptureResult?) checks) {
  for (final setup in  widgets) {
    testWidgets(testDescription, (tester) async {
      // Given
      final tree = setup();
      await tester.pumpWidget(tree);
      
      // When
      final capture = await recorder.performCapture();
      
      // Then
      checks(capture);
    }
  }
}

metaTestWidgets('my test', [ 
  () => captureCheckbox(recorder, Checkbox(value: true, onChanged: (_) {}),
  () => captureCheckbox(recorder, CupertinoCheckbox(value: true, onChanged: (_) {}),
], (capture) {
  expect(capture, isNotNull);
  // Etc.
});
```

fuzzybinary (changes requested):

JuanNaranjoDD:
All comments addressed, ready for another review, I already finished the Radio widget support, you can either check it now or I can open an specific PR for that feature, just let me know what is more convenient for you.

fuzzybinary:
Do a separate PR for the Radio button please.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
```diff
@@ -0,0 +1,190 @@
...
+    CapturedViewAttributes attributes,
+    TreeCapturePrivacy capturePrivacy,
+  ) {
+    final widget = element.widget;
+    if (widget is! CupertinoCheckbox) return null;
+
+    // Resolves for privacy settings
+    final bool isMasked = capturePrivacy.isMasked;
+
+    final bool? value = (!isMasked) ? widget.value : false;
```
fuzzybinary: nit: You do not have to specify the type in Dart when you declare the variable unless it's ambiguous for some reason. Some style guides (including the one used at Google) require you to do that, but we don't, so it's okay to leave them off.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
```diff
@@ -0,0 +1,190 @@
...
+
+    final states = <WidgetState>{
+      if (widget.onChanged == null) WidgetState.disabled,
+      if (value != false) WidgetState.selected,
+    };
```
fuzzybinary, line 74: I know some widgets carry their state in a `Set<WidgetState>` but Checkbox and Radio don't appear to, and you're only interested in two bools here - `isDisabled` and `isSelected` -- you can simplify this to a Dart `Record` (https://dart.dev/language/records, likely using a [typedef](https://dart.dev/language/records#records-and-typedefs)) which would make the code checking those states easier to follow.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
```diff
@@ -0,0 +1,190 @@
...
+    required Set<WidgetState> states,
+  }) {
+    if (states.contains(WidgetState.disabled) &&
+        states.contains(WidgetState.selected)) {
+      return widget.checkColor ?? _kDisabledCheckColor.resolveColor(element);
+    }
+    if (states.contains(WidgetState.selected)) {
+      return widget.checkColor ?? _kDefaultCheckColor.resolveColor(element);
+    }
+    return CupertinoColors.white;
```
fuzzybinary, line 155: Default here probably shouldn't be white, since that will be white on white.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
```diff
@@ -0,0 +1,315 @@
...
+        if (states.contains(WidgetState.error)) {
+          return theme.colorScheme.onError;
+        }
+        return theme.colorScheme.onPrimary;
+      }
+      return Colors
+          .transparent; // No icons available when the checkbox is unselected.
+    }
+
+    return Color(0xFFFFFFFF);
```
fuzzybinary, line 188:
```suggestion
    return Colors.white;
```

fuzzybinary: Also, we're always going to return `white` if the theme doesn't use Material3. Is that correct? The fill color seems to have a fallback for non-material 3 apps...

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/radio_recorder.dart
```diff
@@ -0,0 +1,313 @@
...
+
+import '../../../extensions.dart';
+import '../../../sr_data_models.dart';
+import '../../capture_node.dart';
+import '../../recorder.dart';
+import '../../view_tree_snapshot.dart';
+import '../recording_extensions.dart';
+
+const double _kOuterRadius = 8.0;
+const double _kInnerRadius = 4.5;
```
fuzzybinary: Dart doesn't use the `k` Hungarian prefix for constants. And if nothing else we should be consistent about it 🙂

I've pointed out a few other places in the PR where you're using the `k` but not all of them. Do a quick scan and remove them all if you can.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
```diff
@@ -0,0 +1,190 @@
...
+import 'package:flutter/cupertino.dart';
+
+import '../material_widgets/checkbox_recorder.dart';
+import '../../capture_node.dart';
+import '../../recorder.dart';
+import '../../view_tree_snapshot.dart';
+import '../recording_extensions.dart';
+
+// Default checkbox size
+const double _kEdgeSize = CupertinoCheckbox.width; // 14 px
```
fuzzybinary: Remove `k`

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
```diff
@@ -0,0 +1,190 @@
...
+);
+const CupertinoDynamicColor _kDefaultFillColor =
+    CupertinoDynamicColor.withBrightness(
+  color: CupertinoColors.activeBlue,
+  darkColor: Color.fromARGB(255, 50, 100, 215),
+);
+const Color _kDefaultCheckColor = CupertinoDynamicColor.withBrightness(
+  color: CupertinoColors.white,
+  darkColor: Color.fromARGB(255, 222, 232, 248),
+);
```
fuzzybinary: Remove `k`

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
```diff
@@ -0,0 +1,190 @@
...
+    };
+
+    final Color backgroundColor =
+        _getBackgroundColor(element: element, widget: widget, states: states);
+    final Color symbolColor =
+        _getSymbolColor(element: element, widget: widget, states: states);
+    final BorderSide borderSide =
+        _getBorderSide(element: element, widget: widget, states: states);
+    final double cornerRadius =
+        _getCornerRadius(widget: widget) * attributes.scaleX;
```
fuzzybinary, line 83: I do like this refactoring to use a function to grab each property. 👍

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_radio_recorder.dart
```diff
@@ -0,0 +1,163 @@
...
+
+import '../material_widgets/radio_recorder.dart';
+import '../../capture_node.dart';
+import '../../recorder.dart';
+import '../../view_tree_snapshot.dart';
+import '../recording_extensions.dart';
+
+const double _kOuterRadius = 7.0;
+const double _kInnerRadius = 2.975;
+const double _kBorderOutlineStrokeWidth = 1; // Should be 0.3
```
fuzzybinary: If it should be 0.3... why isn't it 0.3?

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_radio_recorder.dart
```diff
@@ -0,0 +1,163 @@
...
+  Color _getBackgroundColor({
+    required Element element,
+    required CupertinoRadio<dynamic> widget,
+    required Set<WidgetState> states,
+  }) {
+    return states.contains(WidgetState.disabled)
+        ? _kDisabledOuterColor.resolveColor(element)
+        : states.contains(WidgetState.selected)
+            ? (widget.activeColor ?? _kDefaultOuterColor.resolveColor(element))
+            : (widget.inactiveColor ?? CupertinoColors.white);
```
fuzzybinary: These nested ternary operators get really hard to parse and edit. A simpler `if` / `else if` is preferable in these cases.

Also, having the default background and fill colors both be white should probably be changed.

packages/datadog_session_replay/lib/src/capture/element_recorders/recording_extensions.dart
```diff
@@ -27,3 +29,28 @@ extension SRTextAlignment on TextAlign {
     }
   }
 }
+
+extension TreeCapturePrivacyExtension on TreeCapturePrivacy {
+  bool get isMasked => 
+    textAndInputPrivacyLevel == TextAndInputPrivacyLevel.maskAllInputs ||
+    textAndInputPrivacyLevel == TextAndInputPrivacyLevel.maskAll;
+}
```
fuzzybinary, line 37: `isMasked` isn't a great name here, as you can still have unmasked text if `maskAllInputs` is set. `shouldMaskInputs` or something similar is better for this.

packages/datadog_session_replay/lib/src/capture/element_recorders/recording_extensions.dart
```diff
@@ -2,9 +2,11 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2025-Present Datadog, Inc.
 
-import 'package:flutter/widgets.dart';
+import 'package:flutter/cupertino.dart';
```
fuzzybinary: Can we avoid importing `cupertino.dart`? Or move the Cupertino specific stuff into it's own file?

Reason: Flutter is in the middle of a [very large project](https://github.com/flutter/flutter/issues/101479) that will decouple the visual representation of widgets, such as the Material and Cupertino libraries, out of the main Flutter package. While this will take a while means *lots* of SR will have to change as a result, I'd like to keep the "core" packages free of references to `cupertino` and `material` so we can eventually split out the design language packages cleanly.

packages/datadog_session_replay/lib/src/capture/element_recorders/recording_extensions.dart
```diff
@@ -27,3 +29,28 @@ extension SRTextAlignment on TextAlign {
...
+  BorderSide? resolveSide(Set<WidgetState> states) {
+    if (this is WidgetStateProperty) {
+      return WidgetStateProperty.resolveAs<BorderSide?>(this, states);
+    }
+    if (!states.contains(WidgetState.selected)) {
+      return this;
+    }
+    return null;
+  }
+}
```
fuzzybinary:
```suggestion
}

```

Make sure you're running `dartfmt` on save. I think it should have caught this.

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -67,20 +76,25 @@ class KeyGenerator {
...
   final Expando<int> _resourceIdExpando = Expando('sr-resource-key');
 
-  int keyForElement(Element e) {
-    var value = _nodeIdExpando[e];
-    if (value != null) return value;
-
-    value = _nextElementKey;
-    _nextElementKey = _nextElementKey + 1;
-    if (_nextElementKey >= maxKey) _nextElementKey = 0;
+  int keyForElement(Element e, {int wireFrame = 0}) {
```
fuzzybinary:
```suggestion
  int keyForElement(Element e, {int wireframe = 0}) {
```

I know, it's weird, but it's technically one word.

A more descriptive name might also be `wireframeIndex` or `wireframeId`.

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -67,20 +76,25 @@ class KeyGenerator {
   var _nextElementKey = 0;
   var _nextResourceKey = startingResourceKey;
 
-  final Expando<int> _nodeIdExpando = Expando('sr-key');
+  final Expando<List<int>> _nodeIdExpando = Expando('sr-key');
```
fuzzybinary: This might be better as a `Map<int, int>`, which would make the list expansion code below unnecessary.

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -91,11 +105,15 @@ class KeyGenerator {
 
     value = _nextResourceKey;
     _nextResourceKey = _nextResourceKey + 1;
-    if (_nextResourceKey >= maxKey) _nextResourceKey = startingResourceKey;
+    if ( _nextResourceKey >= maxKey ) _nextResourceKey = startingResourceKey;
```
fuzzybinary:
```suggestion
    if (_nextResourceKey >= maxKey) _nextResourceKey = startingResourceKey;
```

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -91,11 +105,15 @@ class KeyGenerator {
...
 
     _resourceIdExpando[e] = value;
     return value;
   }
+
+  int nextWireFrame(Element e) {
+    return _nodeIdExpando[e]?.length ?? 0;
+  }
```
fuzzybinary, line 123: This isn't used?

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -335,7 +362,8 @@ class SessionReplayRecorder {
       }
 
       final widget = e.widget;
-      final recorder = _elementRecordersByType[widget.runtimeType];
+      final recorder = _elementRecordersByType[widget.runtimeType] 
+        ?? _genericElementRecorder.firstWhereOrNull((r) => r.accepts(widget));
```
fuzzybinary, line 373: I'm curious why this was added over using the existing mechanism? It's not a bad way of doing it, but it doesn't seem to offer anything over the current mechanism, at least not in this PR.

If we want to switch to allowing more complicated checks for recording with an `accepts` style, I'd rather we refactor to do it across the board rather than have two registration / lookup mechanisms.

packages/datadog_session_replay/pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
   plugin_platform_interface: ^2.0.2
   datadog_flutter_plugin: ^3.1.0
   json_annotation: ^4.9.0
+  collection: ^1.19.1
```
fuzzybinary: While `collection is a great package and I use it pretty much everywhere in the test and example libraries, we want to be careful with the dependencies we bring in, as it adds a maintenance burden on us and a potential version constraint on our clients (who may want to update to, say 2.0 when it comes out before we're ready to).

I think if we can rethink the `GenericElementRecorder` we won't need to call `firstWhereOrNull` which will allow us to drop `collection`.

fuzzybinary (changes requested):

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
JuanNaranjoDD: Actually Flutter internally associates a Set<WidgetState> to each widget so it can resolve its layout and styling based on the current configuration. For example, you can see this in how the fillColor property of the Cupertino checkbox widget is resolved: https://api.flutter.dev/flutter/cupertino/CupertinoCheckbox/fillColor.html

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
JuanNaranjoDD: Good point, but looking directly at the Flutter codebase, here’s how checkColor is actually resolved:

WidgetStateProperty<Color> get _defaultCheckColor {
    return WidgetStateProperty.resolveWith((Set<WidgetState> states) {
      if (states.contains(WidgetState.disabled) && states.contains(WidgetState.selected)) {
        return widget.checkColor ?? CupertinoDynamicColor.resolve(_kDisabledCheckColor, context);
      }
      if (states.contains(WidgetState.selected)) {
        return widget.checkColor ?? CupertinoDynamicColor.resolve(_kDefaultCheckColor, context);
      }
      return CupertinoColors.white;
    });
  }
  
Look for:
class CupertinoCheckbox extends StatefulWidget
Declared in package:flutter/src/cupertino/checkbox.dart.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
JuanNaranjoDD: Would you like me to update all of them, or is it fine to leave them as they are for now?

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/checkbox_recorder.dart
JuanNaranjoDD: Absolutely, the original code for M2 material is: 

@override
  WidgetStateProperty<Color> get checkColor {
    return WidgetStateProperty.all<Color>(const Color(0xFFFFFFFF));
  }

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
fuzzybinary: Just an FYI, no need to resolve.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/radio_recorder.dart
JuanNaranjoDD: Actually, I copied the constant names directly from Flutter, look at package:flutter/src/material/radio.dart.

enum _RadioType { material, adaptive }

const double _kOuterRadius = 8.0;
const double _kInnerRadius = 4.5;

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
JuanNaranjoDD: That's how Flutter calls the variables in the file: package:flutter/src/cupertino/checkbox.dart.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
JuanNaranjoDD: package:flutter/src/cupertino/checkbox.dart.

// Eyeballed from a checkbox on a physical Macbook Pro running macOS version 14.5.
const Color _kDisabledCheckColor = CupertinoDynamicColor.withBrightness(
  color: Color.fromARGB(64, 0, 0, 0),
  darkColor: Color.fromARGB(64, 255, 255, 255),
);
const Color _kDisabledBorderColor = CupertinoDynamicColor.withBrightness(
  color: Color.fromARGB(13, 0, 0, 0),
  darkColor: Color.fromARGB(13, 0, 0, 0),
);
const CupertinoDynamicColor _kDefaultBorderColor = CupertinoDynamicColor.withBrightness(
  color: Color.fromARGB(255, 209, 209, 214),
  darkColor: Color.fromARGB(50, 128, 128, 128),
);
const CupertinoDynamicColor _kDefaultFillColor = CupertinoDynamicColor.withBrightness(
  color: CupertinoColors.activeBlue,
  darkColor: Color.fromARGB(255, 50, 100, 215),
);
const Color _kDefaultCheckColor = CupertinoDynamicColor.withBrightness(
  color: CupertinoColors.white,
  darkColor: Color.fromARGB(255, 222, 232, 248),
);
const double _kPressedOverlayOpacity = 0.15;
// In dark mode, the fill color of a checkbox is an opacity gradient of the
// background color.
const List<double> _kDarkGradientOpacities = <double>[0.14, 0.29];
const List<double> _kDisabledDarkGradientOpacities = <double>[0.08, 0.14];

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_checkbox_recorder.dart
fuzzybinary: It's old style that is [currently discouraged](https://dart.dev/effective-dart/style#dont-use-prefix-letters) in Dart. We can copy code bug that doesn't mean we have have to copy over their mistakes 🙂

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_radio_recorder.dart
JuanNaranjoDD: We cannot render borders with non-integer widths. This is not a Flutter issue; it is simply how wireframes work. If wireframes support fractional border widths in the future, we can add that support as well. Right now, I round the values in code, but a value like 0.3 rounds to 0, which results in no border being rendered.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_radio_recorder.dart
JuanNaranjoDD: I get this: "Also, having the default background and fill colors both be white should probably be changed." but it wasn't design choice, I'm mirroring the real implementation: 

WidgetStateProperty<Color> get _defaultOuterColor {
    return WidgetStateProperty.resolveWith((Set<WidgetState> states) {
      if (states.contains(WidgetState.disabled)) {
        return CupertinoDynamicColor.resolve(_kDisabledOuterColor, context);
      }
      if (states.contains(WidgetState.selected)) {
        return widget.activeColor ?? CupertinoDynamicColor.resolve(_kDefaultOuterColor, context);
      }
      return widget.inactiveColor ?? CupertinoColors.white;
    });
  }
  
  Declared in package:flutter/src/cupertino/radio.dart.

packages/datadog_session_replay/lib/src/capture/element_recorders/recording_extensions.dart
JuanNaranjoDD: I use resolveColor on all Cupertino widgets, then I need to import the package here, another solution could be creating a cupertino_extensions.dart file within cupertino_widgets folder and put the function there, let me know if it fits you.

packages/datadog_session_replay/lib/src/capture/recorder.dart
JuanNaranjoDD: Is not used but it could be useful in the future I think, anyway I can delete it if you want, is more like a help method.

packages/datadog_session_replay/lib/src/capture/recorder.dart
JuanNaranjoDD: The GenericElementRecorder / accepts() mechanism was specifically needed because Radio<T> and CupertinoRadio<T> are generic widgets. At runtime their runtimeType resolves to Radio<int>, Radio<String>, etc., so a single Type key in the map would never match them. The is Radio check is the only reliable way to handle this (the user can associate custom objects to it, then making a matching list like [Radio<int>, Radio<String>, etc] is not a reliable option here)

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_radio_recorder.dart
```diff
@@ -120,44 +121,48 @@ class CupertinoRadioRecorder implements GenericElementRecorder {
...
-        : states.contains(WidgetState.selected)
-            ? (widget.activeColor ?? _kDefaultOuterColor.resolveColor(element))
-            : (widget.inactiveColor ?? CupertinoColors.white);
+    if (states.contains(WidgetState.disabled)) {
+      return _disabledOuterColor.resolveColor(element);
+    }
+    if (states.contains(WidgetState.selected)) {
+      return widget.activeColor ?? _defaultOuterColor.resolveColor(element);
+    }
+    return widget.inactiveColor ?? CupertinoColors.white;
```
fuzzybinary, line 130: Yeah this is way clearer 👍

packages/datadog_session_replay/lib/src/capture/recorder.dart
fuzzybinary: Got it. Let's then shift all of the recorders to use `accepts` over the map in a future PR. Having two methods for recorder resolution creates unnecessary complexity I think we can do without.

fuzzybinary (approved):

fuzzybinary:
You'll likely need to merge from master to get this to pass `build-android` as it might still need to ignore analyzing 3rd party packages.

Issue #979, opened 2026-03-26
flutterGetDatadogContext FFI symbol stripped by linker in static framework builds — breaks distributed tracing


MelonMania:
## Description

The `flutterGetDatadogContext` FFI symbol is correctly implemented in Swift (`DatadogSdkPlugin.swift:20-36`) with `@_cdecl`, but gets dead-stripped by the linker when the plugin is built as a static framework. This causes distributed tracing to silently fail in both `datadog_dio` and `datadog_tracking_http_client`.

## Environment

- datadog_flutter_plugin: 3.1.0
- datadog_dio: 2.1.0
- datadog_tracking_http_client: 3.1.0
- datadog_session_replay: 1.0.0-preview.10
- Flutter: 3.32.5 / Dart: 3.8.1
- Platform: iOS (CocoaPods with `use_frameworks!`)

## Steps to Reproduce

1. Setup Datadog SDK with `firstPartyHostsWithTracingHeaders` configured
2. Use `datadog_dio` or `datadog_tracking_http_client` for distributed tracing
3. Make an HTTP request
4. Observe: trace headers (`x-datadog-trace-id`, `traceparent`) are NOT injected

## Error

When calling `generateTracingContext()` which calls `sdk.platform.getContext()` → `_getDatadogContext()` FFI:

```
Invalid argument(s): Couldn't resolve native function 'flutterGetDatadogContext'
```

This error is silently caught by `datadog_dio`'s `_trackRequest` try-catch, making it invisible unless explicitly debugged.

## Root Cause Analysis

The `@ffi.Native` annotation in `ios_platform_bridge.dart:51-53`:
```dart
@ffi.Native<_DatadogCContext Function()>(
    isLeaf: true, symbol: 'flutterGetDatadogContext')
external _DatadogCContext _getDatadogContext();
```

Has **no `assetId` parameter** and no `@DefaultAsset` library annotation, so Dart falls back to `dlsym(RTLD_DEFAULT, "flutterGetDatadogContext")` for process-level symbol lookup.

Meanwhile, the podspec sets `s.static_framework = true`. Combined with `use_frameworks!` in the Podfile, this produces a static framework. The linker extracts object files from the static archive and only includes those needed to resolve static references.

Since `flutterGetDatadogContext` is a **module-level free C function** (not an ObjC class member), the `-ObjC` linker flag does NOT force its inclusion. The linker dead-strips it because nothing statically references it.

Evidence:
[...]

PR #991, opened 2026-04-15
fix(gql_link): Catch errors on tracing header creation

fuzzybinary:
### What and why?

Errors in creating tracing headers shouldn't tank the whole network request. Make sure that we catch any errors in creating the tracing headers but still track the resource and perform the network request if an error occurs.

While this doesn't fix this missing symbol error that's causing issues, it will prevent it from affecting end users.

refs: #979

### Review checklist


0xnm (approved):

marco-saia-datadog (approved):
Looks good 👍  I would add a unit test to cover this scenario for completeness

Issue #990, opened 2026-04-09
`flutterGetDatadogContext` FFI symbol not resolved on iOS AOT release builds (possible regression of #979)


fabn:
## The bug

In an iOS AOT release build (distributed via TestFlight), every GraphQL request going through `DatadogGqlLink` throws with:

```
Invalid argument(s): Couldn't resolve native function 'flutterGetDatadogContext' in
'package:datadog_flutter_plugin/src/ios/ios_platform_bridge.dart' :
No asset with id 'package:datadog_flutter_plugin/src/ios/ios_platform_bridge.dart' found.
No available native assets. Attempted to fallback to process lookup.
dlsym(RTLD_DEFAULT, flutterGetDatadogContext): symbol not found.
```

The exception originates from `IosPlatformBridge._getDatadogContext()`:

```dart
@ffi.Native<_DatadogCContext Function()>(
    isLeaf: true, symbol: 'flutterGetDatadogContext')
external _DatadogCContext _getDatadogContext();
```

and is reached indirectly via `generateTracingContext(sdk, rum)` at `lib/src/tracing/tracing_headers.dart:214`, which is the path `datadog_gql_link` uses to attach trace headers to every outbound GraphQL operation.

**Both lookup strategies fail**:
1. The Dart native-assets registry reports `No asset with id 'package:datadog_flutter_plugin/src/ios/ios_platform_bridge.dart' found. No available native assets.`
2. The `dlsym(RTLD_DEFAULT, ...)` fallback reports `symbol not found`.

The Swift side declares the symbol as expected (unchanged since the fix in #979):

```swift
@_cdecl("flutterGetDatadogContext")
public func flutterGetDatadogContext() -> DatadogCContext { ... }
```

## Environment

- **datadog_flutter_plugin**: `3.2.0`
- **datadog_gql_link**: `2.0.0`
- **Flutter**: `3.35.5` stable
- **Dart**: `3.9.2`
- **Xcode**: `26.4`
- **iOS runtime** on physical device: `26.3.1` (arm64e)
- **Build mode**: `flutter build ios --release` with `--obfuscate` and `--split-debug-info`
- **Distribution**: TestFlight
- **Datadog site**: EU1

## Expected behaviour

`DatadogGqlLink.request` should be able to call `sdk.platform.getContext()` on every GraphQL request without the FFI lookup failing, exactly as #979 fixed for AOT release builds in `3.1.1`.

## Actual behaviour

[...]

PR #993, opened 2026-04-16
fix: Prevent stripping of iOS context methods

fuzzybinary:
### What and why?

Archived builds on iOS were stripping our C functions through dead symbol stripping, regardless of whether we told it to keep the symbol or not. Following SR as an example, we've switched from using a C method to using a static method on an Objective-C object instead. This prevents iOS from stripping the symbol, even in archive / ipa builds.

Other options would have been to force clients to modify their `STRIP_STYLE` setting in their Xcode project, which we decided was an unreasonable ask.

refs: #990

### Review checklist


packages/datadog_flutter_plugin/lib/src/ios/ios_platform_bridge.dart
```diff
@@ -8,46 +8,67 @@ import 'package:ffi/ffi.dart';
...
+
+      final obj = _objcMsgSend(cls,
+          _selRegisterName('current'.toNativeUtf8(allocator: arena).cast()));
+      if (obj == ffi.nullptr) return null;
+
+      final sessionId = _readNSString(obj, 'sessionId', arena);
+      final accountId = _readNSString(obj, 'accountId', arena);
+      final userId = _readNSString(obj, 'userId', arena);
+
+      if (sessionId == null && accountId == null && userId == null) return null;
```
Valpertui, line 48: Is it necessary to return `null` instead of an empty `DatadogContext` if all the fields are `null` ?  
I believe it create confusion between "I have no context" and "I have a context but it is empty".

fuzzybinary: It's not necessary, but consistent with the previous version of this function. In reality, if `sessionId` is null SessionReplay can't do much with the context anyway.

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
```diff
@@ -177,6 +172,7 @@ public class DatadogSdkPlugin: NSObject, FlutterPlugin, DatadogFeature {
                     // swiftlint:disable:next force_try
                     try! core?.register(feature: self)
 
+
```
tienquocbui, line 175: nit: redundant blank line.

tienquocbui (approved):

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Package.resolved
```diff
@@ -0,0 +1,50 @@
+{
+  "pins" : [
+    {
+      "identity" : "dd-sdk-ios",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/Datadog/dd-sdk-ios.git",
+      "state" : {
+        "branch" : "develop",
```
simaoseica-dd, line 8: Is this ok to point to `develop`?

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
```diff
@@ -9,30 +9,25 @@ import DatadogCrashReporting
...
-// owned by the caller, and must be freed to prevent a memory leak.
-@_cdecl("flutterGetDatadogContext")
-public func flutterGetDatadogContext() -> DatadogCContext {
-    let context = ContextMessageReceiver.shared.cachedContext
-    guard let context = context else {
-        return DatadogCContext(sessionId: nil, accountId: nil, userId: nil)
+// Exposes the cached Datadog context to Dart via ObjC runtime dispatch.
+// Using an @objc class rather than a @_cdecl C function ensures the symbol
+// survives Xcode archive builds.
+@objc(DatadogContextBridge)
```
tienquocbui, line 16: The PR description says "following SR", but `datadog_session_replay` also keeps a `@_silgen_name("__datadog_session_replay_keep_symbols")` function as belt‑and‑suspenders against the linker dropping `@objc` classes. Was leaving that out here intentional, or worth mirroring with e.g. `__datadog_flutter_plugin_keep_symbols`?

simaoseica-dd:
@codex review

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Package.resolved
fuzzybinary: Yes. For ease of development and to catch incompatibilities early, we point resolution to `develop`. We pin to the most recent version during release.

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
fuzzybinary: It's a good callout. Leaving it out wasn't intentional but it doesn't seem to make a difference to archive stripping, and in fact the `@_silgen_name` fix for SR probably doesn't do what we intended when it was added (it seemed to, but it was probably a coincidence with something else).

In reality, the Obj-C combination being public with a non-mangled name using `@objc` seems to keep things safe from the symbol stripper, even during Archive builds.

PR #996, opened 2026-04-20
fix(sr): Recognize `ExactAssetImage` as an asset when masking non-assets

gonzalezreal:
### What and why?

Under `ImagePrivacyLevel.maskNonAssetsOnly`, Session Replay was incorrectly masking images loaded via `Image.asset()` when an explicit `scale` parameter was provided. Flutter creates an `ExactAssetImage` internally in that case, but `_extractAssetImage` only recognized `AssetImage`, so these images were treated as non-assets and replaced with placeholders in the replay.

### How?

- `_extractAssetImage` now checks for `AssetBundleImageProvider` instead of `AssetImage`. This parent class covers both `AssetImage` and `ExactAssetImage`, as well as any future `AssetBundleImageProvider` subclasses.

- Added a regression test.

refs: PANA-7063

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
+++ b/packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
@@ -171,14 +171,14 @@ class ImageRecorder implements ElementRecorder {
     );
   }
 
-  AssetImage? _extractAssetImage(Image widget) {
-    AssetImage? assetImage;
-    if (widget.image is AssetImage) {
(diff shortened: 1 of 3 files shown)


fuzzybinary (approved):
It might be worth adding a set of golden tests that we know resolves to an ExactAssetImage as well.

gonzalezreal:
> It might be worth adding a set of golden tests that we know resolves to an ExactAssetImage as well.

I will follow up on this in a future PR. Thanks for taking a look!

Issue #995, opened 2026-04-17
[Rum & Session Replay] [IOS] How long does it take for a session to appear/transmitted on dashboard ?


Mohammad-amleh:
### Describe the bug

How long does it take for a session to be displayed or transmitted? I’m using the iOS Simulator, where sessions appear only occasionally, while on a real device no sessions appear at all (test-flight build)—even with RUM and session replay sampling set to 100%. Android works perfectly. Here is my setup:

it's called before run app 

   ```
    final configuration = DatadogConfiguration(
      clientToken: config.clientToken,
      env: config.environment,
      site: DatadogSite.us1,
      service: config.serviceName,
      version: config.version,
      uploadFrequency: UploadFrequency.frequent,
      batchSize: BatchSize.small,
      nativeCrashReportEnabled: true,
      firstPartyHosts: config.firstPartyHosts,
      rumConfiguration: DatadogRumConfiguration(
        applicationId: config.applicationId,
        sessionSamplingRate: config.sessionSamplingRate,
        detectLongTasks: true,
        trackFrustrations: true,
        errorEventMapper: (event) {
          final url = event.error.resource?.url ?? '';
          if (_isThirdPartyTelemetry(url) ||
              _isThirdPartyTelemetry(event.error.message)) {
            return null;
          }
          return event;
        },
      ),
    )
      ..enableSessionReplay(
        DatadogSessionReplayConfiguration(
          textAndInputPrivacyLevel:
              TextAndInputPrivacyLevel.maskSensitiveInputs,
          touchPrivacyLevel: TouchPrivacyLevel.show,
          imagePrivacyLevel: ImagePrivacyLevel.maskNone,
          replaySampleRate: config.sessionRecordingPercentage,
        ),
      )
      ..enableHttpTracking();

    final dd = DatadogSdk.instance;
    await dd.initialize(configuration, TrackingConsent.granted);
    dd.updateConfigurationInfo(LateConfigurationProperty.trackErrors, true);
    dd.sdkVerbosity = CoreLoggerLevel.debug;
  }
```

**Is there anything I can do on my end to fix this? Is it an iOS limitation, a simulator issue, an SDK bug, or something else? I haven’t tested it on a release (App Store) build yet.**

### Reproduction steps

follow above setup 

### SDK logs

```
flutter: [DATADOG SDK] 🐶 → 19:03:10.600 💡 (session-replay) No upload. Batch to upload: NO, System conditions: ✅
[...]

PR #1001, opened 2026-04-22
chore(sr): Update symbol stripping setting

fuzzybinary:
### What and why?

iOS Archiving uses a second pass dead code stripper that removes our Session Replay FFI trampolines in SPM builds. this doesn't seem to affect Cocoapods because Cocoapods builds itself into a dynamic library, where the public symbols are kept intact. Since SPM links everything statically, the archiving code stripper removes the statically linked (supposedly unused) trampoline functions.

The only way we've found around this is to force users to change the `STRIP_STYLE` setting in their Xcode project. 

This updates the documentation and changes the example applications to uses the correct `STRIP_STYLE`.

Note, the fix in #993 still holds and the `STRIP_STYLE` change is not needed for the core library (yet), because the archiver will not strip public Objective-C classes, as it knows they can be called dynamically through Objective-C message passing. The same fix is not available here as functions are generated by `ffigen` and must be C.

refs: #995 #990 

### Review checklist


maycmlee (dismissed):

fuzzybinary:
Bypassing checks as this is mostly a documentation change (and docs reviewed)

PR #1004, opened 2026-04-22
fix: Prevent stripping of iOS context methods

fuzzybinary:
### What and why?

Cherry pick #993 into the 3.1.x branch so we can issue a patch release to fix the isssue.

_Archived_ builds on iOS were stripping our C functions through dead symbol stripping, regardless of whether we told it to keep the symbol or not. Following SR as an example, we've switched from using a C method to using a static method on an Objective-C object instead. This prevents iOS from stripping the symbol.

Other options would have been to force clients to modify their `STRIP_STYLE` setting in their Xcode project, which we decided was an unreasonable ask.

refs: #990
(cherry picked from commit aa74b5e56fcab588780e10367d7b2dffe6177a78)

### Review checklist


0xnm (approved):

fuzzybinary:
Android build issues are unrelated. Bypassing checks.

PR #1011, opened 2026-04-27
chore: Fix android build by adding missing methods.

fuzzybinary:
### What and why?

Updates to Android API need to be implemented in the Mock to fix the build.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/MockRumMonitor.kt
+++ b/packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/MockRumMonitor.kt
@@ -8,6 +8,7 @@ import com.datadog.android.rum.RumResourceKind
 import com.datadog.android.rum.RumResourceMethod
 import com.datadog.android.rum._RumInternalProxy
 import com.datadog.android.rum.featureoperations.FailureReason
+import com.datadog.android.rum.operations.OperationOptions
 import io.mockk.mockk
 
 class MockRumMonitor : RumMonitor {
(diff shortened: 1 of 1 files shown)


packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/MockRumMonitor.kt
```diff
@@ -87,6 +88,16 @@ class MockRumMonitor : RumMonitor {
         mockMonitor.failFeatureOperation(name, operationKey, failureReason, attributes)
     }
 
+    @ExperimentalRumApi
+    override fun failOperation(
+        name: String,
+        operationKey: String?,
+        failureReason: com.datadog.android.rum.operations.FailureReason,
```
0xnm, line 95:
```suggestion
        failureReason: FailureReason,
```

0xnm (approved):

packages/datadog_flutter_plugin/android/src/test/kotlin/com/datadoghq/flutter/MockRumMonitor.kt
fuzzybinary: I wanted to do that. It collides with the deprecated namespace for `FailureReason` used in `failFeatureOperation` above.

0xnm: well, it is possible to use `as` in the import to give different names, but I guess since it is a mock, we can simply use FQCN if it doesn't trigger any lint rules.

PR #1013, opened 2026-04-27
test: Add a test for addUserExtraInfo

fuzzybinary:
### What and why?

We wanted to see what the behavior of `addUserExtraInfo` would be without having first called `setUserInfo`.

It turns out all the native SDKs allow this, and I've added a test to ensure this functionality doesn't break.

### Review checklist


gonzalezreal (approved):

JuanNaranjoDD (commented):
Got it, you just moving the user properties matching after the first setUserInfo. It's correct.

PR #1014, opened 2026-04-28
feat(sr): add Switch and CupertinoSwitch session replay recorders

JuanNaranjoDD:
### What and why?

Adds Session Replay support for 'Switch' and 'CupertinoSwitch' widgets. Without this, switch controls render in SR as a box with an 'x' inside regardless the state or the privacy settings.

### How?

Introduces 'SwitchRecorder' (Material) and CupertinoSwitchRecorder' (Cupertino), each producing two 'SRShapeWireframe's per widget: a rounded-rect track and a circular thumb.

### Review checklist

- [x] This pull request references a Github or JIRA issue — RUM-10250


packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_switch_recorder.dart
```diff
@@ -0,0 +1,173 @@
...
+
+/// Detects 'CupertinoSwitch' widgets and places a Switch icon
+/// on SessionReplay.
+class CupertinoSwitchRecorder implements ElementRecorder {
+  final KeyGenerator keyGenerator;
+
+  const CupertinoSwitchRecorder(this.keyGenerator);
+
+  @override
+  List<Type> get handlesTypes => [CupertinoSwitch];
```
fuzzybinary, line 27: Are we not going to move to the `accepts` system for all recorders?

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_switch_recorder.dart
```diff
@@ -0,0 +1,173 @@
...
+    return color;
+  }
+
+  Color _resolveActiveColor(CupertinoSwitch widget, Element element) {
+    final CupertinoThemeData theme = CupertinoTheme.of(element);
+    final Color colorToResolve = widget.activeTrackColor ??
+        ((widget.applyTheme ?? theme.applyThemeToAll)
+            ? theme.primaryColor
+            : null) ??
+        CupertinoColors.systemGreen;
```
fuzzybinary: Don't go overboard with these ternary / null-coalescing one liners. Again, they're hard to parse and modify, and `if / else` can increase the clarity.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_switch_recorder.dart
```diff
@@ -0,0 +1,173 @@
...
+
+    return colorToResolve.resolveColor(element);
```
fuzzybinary, line 126: I should have asked about this when the extension method was added, but `resolveColor` is just:
```
Color resolveColor(Element element) {
    final resolved = CupertinoDynamicColor.resolve(this, element);
    return Color.from(
        alpha: resolved.a,
        red: resolved.r,
        green: resolved.g,
        blue: resolved.b);
  }
}
```

Is there a reason we recreate the color with `Color.from` here instead of returning it directly?  And is it actually simpler than just doing:

```
return CupertinoDynamicColor.resolve(colorToResolve, element);
```

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/radio_recorder.dart
```diff
@@ -220,7 +220,7 @@ class RadioRecorder implements GenericElementRecorder {
     required CapturedViewAttributes attributes,
     required double radius,
   }) {
-    final w = attributes.width;
+    final w = attributes.height;
```
fuzzybinary: Was this intentional? If so, you should probably change the variable name to h.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
```diff
@@ -0,0 +1,377 @@
...
+    final ThemeData theme = Theme.of(element);
+
+    final bool applyCupertinoTheme = switch (theme.platform) {
+      TargetPlatform.iOS ||
+      TargetPlatform.macOS =>
+        widget.applyCupertinoTheme ??
+            theme.cupertinoOverrideTheme?.applyThemeToAll ??
+            false,
+      _ => false,
+    };
```
fuzzybinary, line 50: If we switch to the `accepts` pattern across the board, we could probably remove the special cases here and change the Cupertino capture class to accept `Swich` if the target platform is iOS or macOS.  That would keep all the Cupertino rendering isolated to one file, and remove a lot of special cases from this one.

Alternately, if you identify a Cuperitno rendered Switch, you could create a temporary CupertinoSwitchRecorder and call into it to do the capture, again isolating those cases to that file.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
```diff
@@ -0,0 +1,377 @@
...
+
+  bool _readIsCupertinoFromPainter(Element element) {
+    bool result = false;
+    void visit(Element child) {
+      if (result) return;
+      final ro = child.renderObject;
+      if (ro is RenderCustomPaint) {
+        try {
+          result = (ro.painter as dynamic).isCupertino as bool? ?? false;
+        } catch (_) {}
```
fuzzybinary: 😬 This construct has a really bad code smell to me. Not only are we casting to `dynamic`, but relying on an exception throw to avoid any issues.

But I'm curious if we we really need to do this? `isCupertino` is private inside of Flutter and seems to only be set [here](https://github.com/flutter/flutter/blob/fe517fcbb48b4ab65630a3a6cffbae3fa6408dc5/packages/flutter/lib/src/material/switch.dart#L942), which is a case we've already covered.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
```diff
@@ -0,0 +1,377 @@
...
+                  : theme.switchTheme.trackColor?.resolve(states)) ??
+              _widgetThumbColor(widget, states)
+                  ?.withValues(alpha: 0x80 / 255.0),
+          false => theme.switchTheme.trackColor?.resolve(states),
+        } ??
+        _defaultTrackColor(
+            element: element,
+            states: states,
+            isCupertinoStyle: isCupertinoStyle,
+            theme: theme);
```
fuzzybinary: If your one liner goes beyond 5 lines it's probably time to not have it be a one liner.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
```diff
@@ -0,0 +1,377 @@
...
+        (isCupertinoStyle
+            ? Colors.transparent
+            : switch (theme.useMaterial3) {
+                true => states.contains(WidgetState.selected)
+                    ? Colors.transparent
+                    : states.contains(WidgetState.disabled)
+                        ? theme.colorScheme.onSurface.withValues(alpha: 0.12)
+                        : theme.colorScheme.outline,
+                false => Colors.transparent,
+              });
```
fuzzybinary: Same here, this is complicated enough it should rely on optional chaining.

fuzzybinary (changes requested):

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_switch_recorder.dart
JuanNaranjoDD: Yes, I'm gonna do that in another branch, so don't worry about it, I just want to keep this branch just for adding the new widgets.

packages/datadog_session_replay/lib/src/capture/element_recorders/cupertino_widgets/cupertino_switch_recorder.dart
JuanNaranjoDD: The reason is that CupertinoDynamicColor.resolve can return a CupertinoColor in some cases (I know it's just a color extension but it parses it differently), and that does not always share the exact structure we need from a plain Color.
So the conversion is necessary to normalize the result before using it. Without that parsing, the replay does not work. I realized that while working on the previous PR.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/radio_recorder.dart
JuanNaranjoDD: Yesss, It was intentional, but you are right, I forgot to change the variable name, this was because I can use the height for determine the radio on both, radio and switch widgets, width only works on radio because it's symmetric, it means, width and height are the same.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
JuanNaranjoDD: applyCupertinoTheme does not mean necessarily the widget is gonna be render as a cupertino widget, it only means the color resolving for track color could be made using cupertino theme, but the shape is still material widget, that one is a border case, I followed the switch material structure.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
JuanNaranjoDD: It was a really big problem for me because I needed to know if the switch was adaptive, anyway, going deeper I just realize that I only need to know if the switch was cupertino (only cupertino is the special case), that could be really easy but this enum is private: enum _SwitchType { material, adaptive }, it was impossible for me to access to it and isCupertino property is within a private class then I cannot identify if cupertino is adaptive in any other way, but I'm gonna check again.

JuanNaranjoDD:
I’m going to address this in a new PR on a separate branch right after this one. I prefer to handle it that way to keep the repository history clear and well organized.

packages/datadog_session_replay/lib/src/capture/element_recorders/material_widgets/switch_recorder.dart
JuanNaranjoDD: I reviewed this and unfortunately dynamic is unavoidable here. Switch.adaptive does not create a CupertinoSwitch child in the subtree, it stays a Material Switch internally but configures itself with Cupertino metrics. _SwitchType is private and isCupertino is only set on _SwitchPainter, which is also private, leaving dynamic as the only escape hatch Dart provides.

I improved the catch to on NoSuchMethodError to be more specific. That said, since this method is only called from within SwitchRecorder when  we already know the widget is a Switch, it should never throw in practice.

fuzzybinary (dismissed):
You'll probably have to merge in some fixes for SR Goldens to be able to pass CI.

fuzzybinary (approved):

PR #1016, opened 2026-04-30
feat(sr): add configurable font-family transform

nottmey:
### What and why?

* Add `DatadogSessionReplayConfiguration.fontFamilyTransform` (`FontFamilyStrategy` / `FontFamilyTransformConfig`) so captured font families can be rewritten to web-compatible CSS stacks in the processor isolate before upload (default `FontFamilyStrategy.none` for backwards compatibility; opt into `FontFamilyStrategy.smart` for normalization). In smart mode, `FontFamilyTransformConfig.rules['']` sets the stack when the captured family is empty or ends up empty after sentinel stripping.

### Review checklist


packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
```diff
@@ -0,0 +1,224 @@
...
+  }
+  final lower = name.toLowerCase();
+  if (_genericFamilyLower.contains(lower)) {
+    if (lower == 'blinkmacsystemfont') {
+      return 'BlinkMacSystemFont';
+    }
+    if (lower == '-apple-system') {
+      return '-apple-system';
+    }
+    return lower;
```
fuzzybinary: nit: You could use pattern matching for this:

```dart
return switch (lower) {
  'blinkmacsystemfont' => 'BlinkMacSystemFont',
  '-apple-system' => '-apple-system',
  _ => lower,
};
```

which I find very slightly cleaner.

packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
```diff
@@ -0,0 +1,224 @@
...
+  FontFamilyTransform(this.config);
+
+  final FontFamilyTransformConfig config;
+
+  String apply(String captured) {
+    switch (config.strategy) {
+      case FontFamilyStrategy.none:
+        return captured;
+      case FontFamilyStrategy.fallback:
+        return kIosParityFontStack;
```
fuzzybinary: Is this only needed on iOS? Is there an Android fallback we should be using as well?

packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
```diff
@@ -0,0 +1,224 @@
...
+      case FontFamilyStrategy.smart:
+        return _applySmart(captured);
+    }
+  }
+
+  SRTextWireframe rewrite(SRTextWireframe wireframe) {
```
fuzzybinary: I think it makes sense for this method to be called `apply` and to potentially change the name of the String transform method. I can see a situation where we might need multiple of these `Transforms` in the future, and having them all implement a method called `apply` makes sense to me, whereas `rewrite` doesn't quite seem like the right term for generalized application of a transform.

I'll leave it to your discretion if you want to change it though.

packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
```diff
@@ -0,0 +1,224 @@
...
+        continue;
+      }
+
+      if (_flutterFontSentinels.contains(stripped)) {
+        continue;
+      }
+
+      if (stripped.isEmpty) {
+        continue;
+      }
```
fuzzybinary: Should this be above the check in _flutterFontSentinels?

packages/datadog_session_replay/lib/src/processor/processor.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'dart:isolate';
 
+import '../../datadog_session_replay.dart';
```
fuzzybinary: This should be below the package imports... I'm surprised the linter didn't complain about that....

packages/datadog_session_replay/lib/src/processor/processor_worker.dart
```diff
@@ -5,6 +5,7 @@
 import 'dart:async';
 import 'dart:convert';
 
+import '../../datadog_session_replay.dart';
```
fuzzybinary: Same here, this should move down to be with the other relative imports.

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -45,6 +45,67 @@ enum TouchPrivacyLevel {
   hide,
 }
 
+/// Controls how captured `TextStyle.fontFamily` values are rewritten before
+/// they are sent as `SRTextStyle.family` on text wireframes.
+///
+/// Custom callbacks are intentionally not supported: Session Replay builds
+/// wireframes in a background isolate, which cannot serialize Dart closures—use
+/// [FontFamilyTransformConfig.rules] instead.
```
fuzzybinary:
```suggestion
/// wireframes in a background isolate, which cannot serialize Dart closures — 
/// use [FontFamilyTransformConfig.rules] instead.
```

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -45,6 +45,67 @@ enum TouchPrivacyLevel {
...
+  /// guaranteed fallback. Yields the iOS-parity stack when the
+  /// captured family is empty or fully sentinel.
+  smart,
+
+  /// Always emits the single hardcoded CSS stack
+  /// `-apple-system, BlinkMacSystemFont, Roboto, sans-serif`,
+  /// regardless of the captured family. Matches the current iOS SDK
+  /// behavior and is the safest choice if you do not want any
+  /// Flutter font names leaving the device.
+  fallback,
```
fuzzybinary, line 71: Does it make sense to check the current Platform and supply an Android fallback as well?

packages/datadog_session_replay/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Add `DatadogSessionReplayConfiguration.fontFamilyTransform` (`FontFamilyStrategy` / `FontFamilyTransformConfig`) so captured font families can be rewritten to web-compatible CSS stacks in the processor isolate before upload (default `FontFamilyStrategy.none` for backwards compatibility; opt into `FontFamilyStrategy.smart` for normalization). In smart mode, `FontFamilyTransformConfig.rules['']` sets the stack when the captured family is empty or ends up empty after sentinel stripping.
+
```
fuzzybinary: We pull CHANGELOG items from the commit message now. I can add in the extra details during the release.

fuzzybinary (commented):
Minor suggestions if you want to take care of them.

I'm going to actually ask @gonzalezreal to review as I think he has familiarity with how iOS handles this and will be able to give me more insight on whether or not we need separate iOS / Android fallback values.

fuzzybinary:
If you want, we could actually setup a golden test for this as well. Since goldens can *only* render one specific font, you could take any of the tests that render text and setup the transformer to make the switch.

packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
nottmey: The `kIosParityFontStack` mirrors the iOS native SDK's `Fallback.fontFamily` (REPLAY-1421). The Android SDK uses "Roboto, sans-serif" — same effective result without the Apple-specific prefixes. Since the replay player is a web browser, the -apple-system, BlinkMacSystemFont prefixes are harmless CSS fallbacks that improve rendering on Safari/Chrome macOS. Roboto already serves as the Android (and general web) fallback. No separate Android stack needed — one stack covers both platforms.

-> update the naming (now `defaultReplayFontStack`)  and comment

packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
nottmey: - rewrite(SRTextWireframe) → apply(SRTextWireframe) — the wireframe-level method now called apply, which is the generalized name that future transforms can all share
- apply(String) → transformFamily(String) — the string-level helper renamed to transformFamily so it no longer collides and better describes what it does (transforms the font family string specifically)
- Updated the one call site in processor_worker.dart and all 27 references in tests

packages/datadog_session_replay/lib/src/processor/font_family_transform.dart
nottmey: yes, good call!

packages/datadog_session_replay/lib/src/processor/processor.dart
nottmey: updated, neither `flutter_lints` nor the underlying `package:lints/recommended.yaml` seem to enable `directives_ordering`

nottmey: enabled it for sr

packages/datadog_session_replay/lib/datadog_session_replay.dart
nottmey: updated the comment, we already use a android compatible fallback

packages/datadog_session_replay/CHANGELOG.md
nottmey: ok, removed

fuzzybinary (dismissed):

nottmey:
@fuzzybinary I'm not sure whether goldens make sense here, created a separate PR for it: https://github.com/DataDog/dd-sdk-flutter/pull/1028 (commit https://github.com/DataDog/dd-sdk-flutter/pull/1028/changes/11033079a74893bf38190c599da29638a88e351b)

fuzzybinary (approved):

PR #1017, opened 2026-04-30
feat(sr): image downscaling with configurable pixel budget

nottmey:
### What and why?

* Add `DatadogSessionReplayConfiguration.imageDownscaling` (`ImageDownscaling`, default `disabled` for backwards compatibility). Set to `enabled` so images larger than the on-screen rendered size (logical bounds × device pixel ratio) or larger than the pixel budget are downscaled in Dart before upload; if raster downscale fails for an oversized image, a "Failed Downscale" placeholder is shown. When `disabled`, oversized images use the "Large Image" placeholder without attempting downscale.
* Add `DatadogSessionReplayConfiguration.maxImagePixelBudget` (default 640 000, ~800×800) to configure the maximum total pixel count for captured images.

### Review checklist


packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -88,7 +96,8 @@ class ImageRecorder implements ElementRecorder {
     }
 
     final totalPixelSize = uiImage.width * uiImage.height;
-    if (totalPixelSize > maxImageSize) {
+    if (totalPixelSize > maxImageSize &&
+        imageDownscaling == ImageDownscaling.disabled) {
```
fuzzybinary: As part of this, would you be up for making `maxImageSize` configurable?

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -130,18 +139,69 @@ class ImageRecorder implements ElementRecorder {
...
+        CoreLoggerLevel.debug,
+        'Session Replay image downscale: '
+        '${srcW}x$srcH (${totalPixels}px) -> ${dstW}x$dstH '
+        '(${dstW * dstH}px), '
+        'viewBounds=${attributes.width.toStringAsFixed(1)}x'
+        '${attributes.height.toStringAsFixed(1)}, '
+        'oversized=${totalPixels > maxImageSize}, '
+        'success=${scaled != null}, '
+        'took ${stopwatch.elapsedMilliseconds}ms',
+      );
```
fuzzybinary: Probably don't want this in the final code 😅

Out of curiosity, did you measure this on lower end devices? Like any of the Samsung A series? What's the timing look like?

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -152,23 +212,146 @@ class ImageRecorder implements ElementRecorder {
...
+      1.0,
+    );
+    var w = math.max(1, (srcW * scale1).round());
+    var h = math.max(1, (srcH * scale1).round());
+
+    final pixels = w * h;
+    if (pixels > maxImageSize) {
+      final scale2 = math.sqrt(maxImageSize / pixels);
+      w = math.max(1, (w * scale2).floor());
+      h = math.max(1, (h * scale2).floor());
```
fuzzybinary: I'm a bit confused what this is doing.... why the square root here?

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
```diff
@@ -152,23 +212,146 @@ class ImageRecorder implements ElementRecorder {
...
 
-    return SpecificElement(
-      subtreeStrategy: CaptureNodeSubtreeStrategy.ignore,
-      nodes: nodes,
+  static double _devicePixelRatio(Element element) {
+    final view = View.maybeOf(element);
+    if (view != null) {
+      return view.devicePixelRatio;
+    }
+    return WidgetsBinding.instance.platformDispatcher.views.first.devicePixelRatio;
```
fuzzybinary: Does this have any possibility of throwing? For example, if views is empty for whatever reason, `first` will likely throw.

packages/datadog_session_replay/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Add `DatadogSessionReplayConfiguration.imageDownscaling` (`ImageDownscaling`, default `disabled` for backwards compatibility). Set to `enabled` so images larger than the on-screen rendered size (logical bounds × device pixel ratio) or larger than `maxImageSize` (~800×800 pixels) are downscaled in Dart before upload; if raster downscale fails for an oversized image, a "Failed Downscale" placeholder is shown. When `disabled`, oversized images use the "Large Image" placeholder without attempting downscale.
+
```
fuzzybinary: CHANGELOG items are now populated from conventional commit messages. In this case you should definitely modify the commit message to be a bit more descriptive.

fuzzybinary (changes requested):
Mostly just looking to remove the log.

I also think it would be good to go with longer variable names for things like `srcW` and `dstW` - `sourceWidth` and `destWidth` take up more space but are easier to skim / parse.

Additionally, before we add this, we may want to add something that can check for "low end" devices and prevent downscaling for those devices, as we already know that SR on low end Android devices can push the performance boundary... image scaling on top of that will likely be detrimental to user experience.

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
nottmey: will check

nottmey: Hmm, Flutter does not expose a reliable "device tier" API — we'd need heuristics (RAM, CPU cores, GPU info) which are platform-specific and fragile

Idea: We could add a maxDownscaleDurationMs timeout to _downscaleImage so if scaling takes too long on a weak device, it bails out and shows the placeholder. The stopwatch logging already there would make this easy to add.

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
nottmey: This is standard uniform scaling math. If an image is w × h pixels and we want to scale both dimensions by the same factor s to hit a target pixel count budget, then:

image

The square root is needed because total pixels scale quadratically with the linear scale factor — scaling width and height each by s multiplies total area by s². I added a comment in the code explaining this.

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
nottmey: afaik usually there is a view attached when the recorder would be running, but let's not count on that. I added a fallback 1.0

packages/datadog_session_replay/CHANGELOG.md
nottmey: removed, updated the commit

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
nottmey: done

packages/datadog_session_replay/lib/src/capture/element_recorders/image_recorder.dart
nottmey: -> building a circuit breaker that ejects after 3 consecutive timeouts (500ms, normal devices take 1-30ms)

packages/datadog_session_replay/lib/datadog_session_replay.dart
```diff
@@ -45,6 +47,20 @@ enum TouchPrivacyLevel {
   hide,
 }
 
+/// Controls Dart-side downscaling of captured images before they are sent to
+/// the native resource pipeline.
+enum ImageDownscaling {
+  /// Legacy behavior: images larger than the pixel budget are replaced by
+  /// a "Large Image" placeholder wireframe.
```
fuzzybinary: nit: I don't think we need to call this out as legacy behavior. Especially because Session Replay is in preview, we have expressed that we might make breaking changes without notice, so we should select what we think is the correct default but allow for configuration.

I do think that  disabling image downscaling is still the correct default here though.

fuzzybinary (dismissed):

fuzzybinary:
@nottmey Rerun `dart format` when you have a minute.

fuzzybinary (dismissed):

nottmey:
@fuzzybinary can't reproduce the CI error, checks are green locally (can't see logs)

fuzzybinary:
I'll take a look and try to bring this one home. I think Flutter goldens generate very, very slightly different images on different versions of macOS, which we can safely ignore.

fuzzybinary:
Actually @nottmey it looks like there's a bug that the golden tests caught.

It appears with downsampling disabled, images that fit within the upload budget are being masked as large images if the need is set to something other than `DownscalingNeed.none`. If downscaling is disabled, even if the image is larger than its on screen size, SR should send it without performing downsscaling.

nottmey:
@fuzzybinary thanks for the investigation! I'll have a look.

fuzzybinary:
Hi @nottmey,

Sorry for the delay on this. The golden tests are still failing because of some differences in how some combination of macOS and Flutter renders the resulting capture. I've created #1031 to be more tolerant of small differences in these images, which you'll have to merge in order to get this one in.

The other build issues are flakes which we'll take care of once the golden tests pass.

nottmey:
@fuzzybinary fyi, merged latest develop

fuzzybinary (approved):
Looks good! Thanks for your patience on this!

nottmey:
@fuzzybinary thanks of the support! 💜

PR #1018, opened 2026-04-30
feat(sr): capture Icon widgets as raster images

nottmey:
### What and why?

* Capture `Icon` widgets as rasterized images with a transparent background and upload them through the same image resource pipeline as `Image`/`RawImage`. Icons are rasterized at a configurable logical size (`DatadogSessionReplayConfiguration.iconRasterLogicalSize`, default `20`) × device pixel ratio, cached by glyph + font + color so the same icon is rasterized once per recorder instance regardless of on-screen `Icon` size. Concurrent raster work for the same icon is deduplicated; additional per-frame capture processing runs in parallel. `ImagePrivacyLevel.maskAll` shows a placeholder (no upload); `maskNone` and `maskNonAssetsOnly` record icons; the inner `RichText` of `Icon` is not captured as text.

### Review checklist


packages/datadog_session_replay/lib/src/capture/element_recorders/icon_recorder.dart
```diff
@@ -0,0 +1,441 @@
...
+    this.resourceKey,
+    this.msToImage = 0,
+    this.msToByteData = 0,
+    this.msSave = 0,
+  });
+
+  final bool success;
+
+  /// `rasterized`, `toImageFailed`, or `toByteDataNull` from [_performRaster].
+  final String outcome;
```
fuzzybinary: Why have these as a String over an `enum`?

packages/datadog_session_replay/lib/src/capture/element_recorders/icon_recorder.dart
```diff
@@ -0,0 +1,441 @@
...
+          ),
+        ],
+      );
+    }
+    final theme = IconTheme.of(context);
+    final displayLogicalSize = widget.size ?? theme.size ?? kDefaultFontSize;
+    final color = widget.color ?? theme.color ?? const Color(0xFF000000);
+    final fontFamily = iconData.fontFamily;
+    if (fontFamily == null || fontFamily.isEmpty) {
+      return SpecificElement(
```
fuzzybinary: We should be able to render the Material Icons font without needing to perform a raster. I think these get captured as Text elements,  so it might be worth checking for that and sending its corresponding Text element instead of performing the raster.

packages/datadog_session_replay/lib/src/capture/element_recorders/icon_recorder.dart
```diff
@@ -0,0 +1,441 @@
...
+      'toByteDataMs=$msToByteData, '
+      'saveMs=$msSave, '
+      'totalMs=$totalMs',
+    );
+  }
+
+  static SpecificElement _iconPlaceholder(
+    int elementId,
+    CapturedViewAttributes attributes,
+  ) {
```
fuzzybinary, line 256: This is a good method, and should probably be used in a few other places you returned icon placeholders.

fuzzybinary (commented):
Looks good to me with some minor nits.

But this needs a Golden test as well.

packages/datadog_session_replay/lib/src/capture/element_recorders/icon_recorder.dart
fuzzybinary: It turns out this feature has a bug on the player side, so if you weren't seeing Material lcons, that's why. We're working on a fix.

nottmey:
image
(example link: https://app.datadoghq.eu/rum/replay/sessions/6a189644-871a-420c-93b0-aa2e2c1616e3?ts=1778180629349)
Note, we are noticing that datadog is dropping uploaded images. Maybe this has to do with your backend bug, maybe it's some kind of upload limit dropping oldest cache entries. @fuzzybinary

fuzzybinary:
@nottmey The dropped uploaded images should be Android specific and fixed in preview 12. If it's not please let me know.

nottmey:
@fuzzybinary tested it out, images are still dropped with icon rasterization + preview 12
https://app.datadoghq.eu/rum/replay/sessions/5c8460d0-2f3e-4cb1-9f2e-480d5073cbfe
image

nottmey:
maybe it works for you?

fuzzybinary:
@gonzalezreal Can you take a look at ☝️ . The android resource id fix should have been pushed in preview 12.

@nottmey Can you open a support request for that? That way we can potentially look at your replays and determine what's going on.

nottmey:
@fuzzybinary will do, tested on an iOS simulator (fyi)

nottmey:
support ticket: https://help.datadoghq.com/hc/requests/2894616

gonzalezreal:
@fuzzybinary Just to confirm, the Android resource-context fix did land in `preview.12` (it's in the changelog).
@nottmey Could you share which platforms you're seeing the dropped images on through the support ticket? You mentioned testing on an iOS simulator. Is that where the screenshots are from, or are you also seeing it on Android? We'll follow up there.

nottmey:
@gonzalezreal we see dropped images on all sessions; we have the feature running in production (start/stop on a selected set of users) (feel free to check all our session recordings)

yes, the screenshot and example session is done from an ios simulator; Tomorrow, I can also reproduce it on android as well and send you the matching link.

fuzzybinary:
Hey @nottmey,

Haven't forgotten about this. I do want to figure out the dropped image issue and make sure you have the built in Material Icons working before we dive into non-material icon capture.

Did you open a support request related to the dropped images? We can't look your sessions without one. If you don't have one, I'd open one. Either way, have them ping me and @gonzalezreal linking to this PR and we can try to figure out why you have dropped images.

nottmey:
@fuzzybinary I opened one, it's [help.datadoghq.com/hc/requests/2894616](https://help.datadoghq.com/hc/requests/2894616), but I didn't reproduce it on Android yet, as I'm on sick leave. Please assume the issue is also present on Android devices. Every session I click into has images missing.

I think there issue could be one of these things:
- rasterization causes too many images to be uploaded to your api, which is then counteracted by dropping old ones, leading to sessions without images
- image id handling is different than I think and ids are not matching up

nottmey:
@fuzzybinary The MaterialIcon only solution would be to reference them by font and assume the font is present in the re-playing browser. That's nothing that would help us and is not a solution that works in general, so I'd vouch for making rasterization work with your api context.

fuzzybinary:
> The MaterialIcon only solution would be to reference them by font and assume the font is present in the re-playing browser.

We actually have the MaterialIcon font available as part of Session Replay. There was a bug where the font wasn't deployed properly that has now been fixed, so we should try to special case out the Material Icons specifically.

But I also want to make rasterization of non material icons work for sure, so I think there's still value here.

What we're waiting on here is that I want to try to figure out why you're seeing missing images before we go ahead and get this working.

Sorry this one is taking so long. We'll probably be releasing a version with your other changes soon, and this one will have to follow after.

nottmey:
Alright, that's fine, we can keep this open until the missing images riddle is solved.

I was asked to create a minimal working reproducible example and will put some effort there in the next days.

nottmey:
Hi @fuzzybinary, thanks for preparing 1.0.0-preview.13! We'll move to that version from then on. 

I don't have the capacity to create a reproducible example for this one, but I'm sure our images will still be broken with 1.0.0-preview.13 (and dropped icon rasterization). I hope this can be solved internally without us having to put more time into it.

Thank you for the support!

fuzzybinary:
👍 We'll take over icon rasterization. 

If you're still seeing dropped images (specifically, those `404` icons) please make sure to update your support ticket and we'll try to figure out what's going on.

cc: @gonzalezreal

fuzzybinary:
Hi @nottmey,

Talking with the SR team - it's possible the "image not found" icons are caused by a token expiration - meaning this is an issue on the Player side, not on the SDK side (which is why it's been hard for us to reproduce).

If you're still seeing this issue with `preview.13`, please let us know through support channels so we can take another look. If the "image not found" icons go away when you close and reopen the player, that would also be good for us to know.

Thanks again for all your help!

PR #1023, opened 2026-05-05
feat(sr): migrate all element recorders from ElementRecorder to GenericElementRecorder

JuanNaranjoDD:
### What and why?

This PR migrates all element recorders in the Session Replay capture pipeline from the old ElementRecorder interface (with List<Type>  handlesTypes) to GenericElementRecorder (with bool accepts(Widget widget)).

Previously, the system used two parallel interfaces: ElementRecorder dispatched recorders via a Map<Type, ElementRecorder> keyed on widget.runtimeType, while GenericElementRecorder extended it and added an accepts method for generic/parameterized widgets (e.g., Radio<T>, CupertinoRadio<T>) that can't be matched by exact runtime type. This duality required special-casing in _populateElementRecorderMap and maintained two separate lookup structures.

By making GenericElementRecorder the single interface and replacing handlesTypes lists with accepts predicates across all recorders, the dispatch logic collapses to a single _genericElementRecorder list, removes the type-map entirely, and gives each recorder full control over its matching logic without loss of performance for common cases.

### How?

- Merged ElementRecorder and GenericElementRecorder into a single GenericElementRecorder interface with captureSemantics + accepts.
- Updated all 12 concrete recorders to implement accepts(Widget widget) using is-checks in place of List<Type> get handlesTypes.
- Removed the _elementRecordersByType map from SessionReplayRecorder and simplified _populateElementRecorderMap to a single addAll.
- Widget dispatch now goes through a single firstWhereOrNull((r) => r.accepts(widget)) call.

### Review checklist

- [x] This pull request references a Github or JIRA issue — RUM-16162

### DIFF
--- a/packages/datadog_session_replay/lib/src/capture/recorder.dart
+++ b/packages/datadog_session_replay/lib/src/capture/recorder.dart
@@ -54,7 +54,7 @@ class TreeCapturePrivacy {
   }
 }
 
-abstract interface class GenericElementRecorder {
+abstract interface class ElementRecorder {
   CaptureNodeSemantics? captureSemantics(
     Element element,
     CapturedViewAttributes attributes,
@@ -131,7 +131,7 @@ class CaptureResult {
 
 class SessionReplayRecorder {
   final DatadogTimeProvider _timeProvider;
-  final List<GenericElementRecorder> _genericElementRecorder = [];
+  final List<ElementRecorder> _elementRecorders = [];
 
   final Map<Key, Element> _elements = {};
   RUMContext? _currentContext;
@@ -164,7 +164,7 @@ class SessionReplayRecorder {
     this._defaultTreeCapturePrivacy,
     this._touchPrivacyLevel,
   ) {
-    _populateElementRecorderMap([
+    _elementRecorders.addAll([
       ContainerRecorder(keyGenerator),
       TextElementRecorder(keyGenerator),
       EditableTextRecorder(keyGenerator),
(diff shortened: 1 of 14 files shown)


packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -54,14 +54,14 @@ class TreeCapturePrivacy {
   }
 }
 
-abstract interface class ElementRecorder {
-  List<Type> get handlesTypes;
-
+abstract interface class GenericElementRecorder {
```
fuzzybinary: I think this name should be shifted back to `ElementRecorder`. The `Generic` bit was more to signify that it was what worked with generics, but now this is just how all `ElementRecorder`s behave.

packages/datadog_session_replay/lib/src/capture/recorder.dart
```diff
@@ -306,17 +301,8 @@ class SessionReplayRecorder {
...
-      if (recorder is GenericElementRecorder) {
-        // Broad match recorder for widgets with generic parameters
-        _genericElementRecorder.add(recorder);
-        continue;
-      }
-      for (final type in recorder.handlesTypes) {
-        _elementRecordersByType[type] = recorder;
-      }
-    }
+  void _populateElementRecorderMap(List<GenericElementRecorder> recorders) {
```
fuzzybinary: This method name now doesn't make sense, and can probably be removed.

fuzzybinary (changes requested):
Very minor naming nits.

fuzzybinary (approved):

PR #1025, opened 2026-05-06
feat(rum): Capture HTTP request and response headers

kikoveiga:
### What and why?

- Add support for capturing HTTP request and response headers in RUM resources. Customers can opt in and have automatic instrumentation with this new feature.
- This aligns the Flutter SDK with the same feature that came recently in Android, iOS and Browser SDKs.

### How?

- Add new `trackResourceHeaders` public API on `DatadogRumConfiguration` that accepts a `ResourceHeadersExtractor`.
- Add new `ResourceHeadersExtractor` class in `datadog_flutter_plugin` that encapsulates all the logic regarding matching, filtering and size-limiting headers. All the internal behavior is aligned with the other SDKs, including default headers, security regex, and size limits.
- Wire this extractor into all 3 tracking clients: `DatadogClient`, `DatadogTrackingHttpClient` and `DatadogDioInterceptor`. Extraction happens in Dart at the point each client calls `stopResource()`.
- Captured headers flow as `_dd.request_headers` and `_dd.response_headers` internal cross-platform attributes, which are then read in the native `RumResourceScope` side and mapped into the resource schema.
- Unit and integrations tests coverage.

Resources with headers taken from the `simple_example` app:
- [Android](https://mobile-integration.datadoghq.com/rum/sessions?query=%40type%3Aresource%20%40application.id%3A948d890c-c895-4060-a960-ac92c51db12d&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZ4HDgwNKacI5wAAABhBWjRIRGhrX0FBQmRacVhjQnlUVGFnQUcAAAAkZjE5ZTA3MGUtNDIzZC00OWE0LWEwMDktYzAwZTFkNDdkODA3AAAAOQ&fromUser=false&viz=stream&from_ts=1777630053253&to_ts=1778234853253&live=true)
- [iOS](https://mobile-integration.datadoghq.com/rum/sessions?query=%40type%3Aresource%20%40application.id%3A948d890c-c895-4060-a960-ac92c51db12d&agg_m=count&agg_m_source=base&agg_t=count&cols=&event=AwAAAZ4G_rY3TUPdHQAAABhBWjRHX3NfeUFBQ01iR1Q5c04xYlZRQVMAAAAkZjE5ZTA2ZmUtYmI4MS00MzgyLTk1MGUtZTJiYmVlYTUwYmJlAAAADg&fromUser=false&viz=stream&from_ts=1777629027641&to_ts=1778233827641&live=true)

### Follow-up
- Add documentation in code if necessary and docs.datadoghq.com.
- If it makes sense, add telemetry to this feature in an upcoming pr, following what we did on Browser and Android.

### Review checklist

### DIFF
--- a/packages/datadog_dio/example/integration_test/dio_interceptor_headers_test.dart
+++ b/packages/datadog_dio/example/integration_test/dio_interceptor_headers_test.dart
@@ -1,0 +1,110 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2025-Present Datadog, Inc.
+import 'dart:convert';
+
+import 'package:collection/collection.dart';
+import 'package:datadog_common_test/datadog_common_test.dart';
+import 'package:datadog_common_test/widget_tester_extensions.dart';
+import 'package:datadog_dio_example/main.dart' as app;
+import 'package:datadog_dio_example/scenario_config.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:integration_test/integration_test.dart';
+
+import 'common.dart';
+
+Future<void> performRumUserFlow(WidgetTester tester) async {
+  await tester.pump(const Duration(seconds: 5));
+
+  var topItem = find.text('Item 0');
+  await tester.tap(topItem);
+  await tester.pumpAndSettle();
+
+  var fetchButton = find.text('Fetch Resources');
+  await tester.tap(fetchButton);
+  await tester.pumpAndSettle();
+
+  var nextButton = find.widgetWithText(ElevatedButton, 'Next Page');
+  await tester.waitFor(nextButton, const Duration(seconds: 100),
+      (e) => (e.widget as ElevatedButton).enabled);
+  await tester.tap(nextButton);
+  await tester.pumpAndSettle();
+}
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  testWidgets('captures request and response headers on RUM resource events',
+      (WidgetTester tester) async {
+    final sessionRecorder = await startMockServer();
+
+    const clientToken = bool.hasEnvironment('DD_CLIENT_TOKEN')
+        ? String.fromEnvironment('DD_CLIENT_TOKEN')
+        : null;
+    const applicationId = bool.hasEnvironment('DD_APPLICATION_ID')
+        ? String.fromEnvironment('DD_APPLICATION_ID')
+        : null;
+
+    final scenarioConfig = RumAutoInstrumentationScenarioConfig(
+      firstPartyHosts: [(sessionRecorder.sessionEndpoint)],
+      firstPartyGetUrl: '${sessionRecorder.sessionEndpoint}/integration_get',
+      firstPartyPostUrl: '${sessionRecorder.sessionEndpoint}/integration_post',
+      firstPartyBadUrl: 'https://foo.bar',
+      thirdPartyGetUrl: 'https://httpbingo.org/get',
+      thirdPartyPostUrl: 'https://httpbingo.org/post',
+      thirdPartyMissingUrl: 'https://httpbingo.org/status/404',
+    );
+    RumAutoInstrumentationScenarioConfig.instance = scenarioConfig;
+
+    app.testingConfiguration = TestingConfiguration(
+        customEndpoint: sessionRecorder.sessionEndpoint,
+        clientToken: clientToken,
+        applicationId: applicationId,
+        firstPartyHosts: ['localhost']);
+    await app.main();
+    await tester.pumpAndSettle();
+
+    await performRumUserFlow(tester);
+
+    final rumLog = <RumEventDecoder>[];
+    await sessionRecorder.pollSessionRequests(
+      const Duration(seconds: 50),
+      (requests) {
+        for (var request in requests) {
+          if (!request.requestedUrl.contains('integration')) {
+            request.data.split('\n').forEach((e) {
+              var jsonValue = json.decode(e);
+              if (jsonValue is Map<String, dynamic>) {
+                rumLog.add(RumEventDecoder(jsonValue));
+              }
+            });
+          }
+        }
+        return RumSessionDecoder.fromEvents(rumLog).visits.length >= 3;
+      },
+    );
+
+    final session = RumSessionDecoder.fromEvents(rumLog);
+    expect(session.visits.length, greaterThanOrEqualTo(3));
+
+    final view2 = session.visits[1];
+
+    final getEvent = view2.resourceEvents
+        .firstWhereOrNull((e) => e.url == scenarioConfig.firstPartyGetUrl);
+    expect(getEvent, isNotNull);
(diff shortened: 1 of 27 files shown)


packages/datadog_dio/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  datadog_flutter_plugin: ^3.0.0
+  datadog_flutter_plugin: ^3.3.0
```
fuzzybinary, line 14: 👍 Glad you caught this.

packages/datadog_flutter_plugin/lib/src/rum/resource_headers_extractor.dart
```diff
@@ -0,0 +1,208 @@
...
+    return result;
+  }
+
+  static Map<String, String> _extractHeaders(
+    Map<String, List<String>> headers,
+    List<String> allowedHeaders,
+  ) {
+    if (allowedHeaders.isEmpty) return const {};
+
+    final allowedSet = allowedHeaders.map((h) => h.toLowerCase()).toSet();
```
fuzzybinary: I'd need to double check, but can we transform any `allowedHeaders` to a lowercase set once instead of on every call to `_extractHeaders`?

Aka, hold both `_requestHeaders` and `_responseHeaders` as a `Set<String>` that is pre-filtered and pre-transformed to lowercase.  It may be a minor performance win but it's still fairly simple for us to do.

packages/datadog_flutter_plugin/lib/src/rum/resource_headers_extractor.dart
```diff
@@ -0,0 +1,208 @@
...
+    List<String> custom,
+  ) {
+    final seen = <String>{};
+    final result = <String>[];
+    for (final h in [...defaults, ...custom]) {
+      final lower = h.toLowerCase();
+      if (seen.add(lower)) {
+        result.add(lower);
+      }
+    }
```
fuzzybinary: To the above point - looks like you're already pre-transforming to a lowercased `Set` but then returning it as a `List`.

Instead, let's make this a `Set<String>` which you can just return `seen` and not keep `result` at all.

packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
```diff
@@ -221,6 +221,17 @@ class DatadogRumConfiguration {
   /// [RumVitalOperationEvent]s before they are sent to Datadog.
   RumVitalOperationEventMapper? vitalOperationStepEventMapper;
 
+  /// Captures HTTP request and response headers from intercepted requests and
+  /// includes them in RUM resource events. Applies across all Datadog HTTP
+  /// tracking clients ([DatadogTrackingHttpClient], [DatadogClient], and
+  /// [DatadogDioInterceptor]).
+  ///
+  /// Requires native SDK 3.8.0+ on both Android and iOS for headers to appear
+  /// in events.
```
fuzzybinary: You can remove this. We don't let users choose which native SDK they use. We pin versions when we release.

packages/datadog_flutter_plugin/test/rum/resource_headers_extractor_test.dart
```diff
@@ -0,0 +1,350 @@
...
+            lessThanOrEqualTo(128));
+      });
+
+      test('truncates without splitting multi-byte characters', () {
+        final extractor = ResourceHeadersExtractor(
+          includeDefaults: false,
+          captureHeaders: ['x-emoji'],
+        );
+        // Each emoji is 4 bytes in UTF-8. Fill to just over 128 bytes.
+        final emoji = '\u{1F600}'; // 4 bytes
```
fuzzybinary: You can actually just put an emoji here.

fuzzybinary (changes requested):
Just a few minor comments, but I think it's worth switching `_requestHeaders` and `_responseHeaders` to pre-transformed `Set<String>`s instead of re-transforming every time.

The doc comment about the required native version also should be removed.

Also - I think it might be worth adding integration tests. They're held in `example/integration_test` in both `datadog_dio` and `datadog_tracking_http_client`.

Also - do we need to do any header capture in GraphQL or gRPC?

kikoveiga (commented):
Addressed your reviews, thank you!

Good point about GraphQL, I think it makes sense to add it in a follow-up PR. We don't support gRPC for now.

packages/datadog_flutter_plugin/test/rum/resource_headers_extractor_test.dart
kikoveiga: Nice, thanks!

packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
kikoveiga: Removed.

packages/datadog_flutter_plugin/lib/src/rum/resource_headers_extractor.dart
kikoveiga: Fixed that, thank you!

packages/datadog_flutter_plugin/lib/src/rum/resource_headers_extractor.dart
```diff
@@ -0,0 +1,208 @@
...
+    if (allowedHeaders.isEmpty) return const {};
+
+    final allowedSet = allowedHeaders.map((h) => h.toLowerCase()).toSet();
+
+    final result = <String, String>{};
+    var totalBytes = 0;
+
+    for (final entry in headers.entries) {
+      if (result.length >= _maxHeadersCount) break;
+      if (totalBytes >= _headerSizeLimitBytes) break;
```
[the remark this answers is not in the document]
kikoveiga: Fixed, aligned behavior with Android & iOS now.

packages/datadog_dio/example/integration_test/dio_interceptor_headers_test.dart
```diff
@@ -0,0 +1,110 @@
...
+    expect(getEvent.responseHeaders!['content-type'], isNotNull);
+
+    final postEvent = view2.resourceEvents
+        .firstWhereOrNull((e) => e.url == scenarioConfig.firstPartyPostUrl);
+    expect(postEvent, isNotNull);
+    expect(postEvent!.requestHeaders, isNotNull);
+    expect(postEvent.requestHeaders!['x-datadog-origin'], 'rum');
+    expect(postEvent.responseHeaders, isNotNull);
+    expect(postEvent.responseHeaders!['content-type'], isNotNull);
+  }, skip: kIsWeb);
```
fuzzybinary, line 109: Seeing this skip raises a good point I didn't think of - we should probably try to support header capture list to web. Can you create a ticket for that?

fuzzybinary (approved):
I think there's a way we can support this on web as well, it just needs to be passed to the RUM configuration at the right time. Let's create a ticket to support it or at least document that it isn't supported on Web.

packages/datadog_dio/example/integration_test/dio_interceptor_headers_test.dart
kikoveiga: Created [RUM-16328](https://datadoghq.atlassian.net/browse/RUM-16328) and will handle it soon.

PR #1031, opened 2026-05-14
chore(sr): Add a tolerant golden file comparator.

fuzzybinary:
### What and why?

Differences in macOS or iOS SDK versions can cause some very slight differences in images during golden testing. 

This creates a `TolerantGoldenFileComparator` that is used for all of our SR golden tests, that allows the images to differ by 1%.  In testing , the highest difference I saw was about .8% when dealing with different versions of macOS / XCode.

### Review checklist


JuanNaranjoDD (approved):
I’ve been struggling with this problem for a long time, so this is a huge improvement.

PR #1037, opened 2026-05-22
fix(sr): render Switch thumbIcon and use Material Icons for Checkbox glyphs

JuanNaranjoDD:
### What and why?

Two rendering fixes for Session Replay's input-widget recorders:

1. Switch ignored widget.thumbIcon. Material Switch lets callers paint an icon inside the thumb (e.g. a check when on, an X when off); the recorder was emitting a plain SRShapeWireframe for the thumb, so the glyph never appeared in replays.
2. Checkbox glyphs drifted from the real widget. The check / dash / mask were emitted as Unicode characters (✓, —, x) rendered in sans-serif. The replay player fell back to whatever system font was available, so stroke weight and proportions didn't match what users actually see on-device.

Both now emit text wireframes backed by the Material Icons font (Icons.check, Icons.remove, Icons.close) using codePoint + fontFamily, which the player already ships. The checkbox symbol scale was also bumped from 0.7 → 0.9 to match the visual weight of the real icon.

### How?

- switch_recorder.dart: resolve widget.thumbIcon (falling back to theme.switchTheme.thumbIcon), compute a thumbIconColor from the icon's own color or the M3 color scheme (with the same disabled-state opacity as the thumb), and emit the thumb as an SRTextWireframe carrying the icon's codePoint + fontFamily. SwitchNode gained thumbIcon / thumbIconColor fields.
- cupertino_switch_recorder.dart: passes thumbIcon: null and a transparent thumbIconColor since Cupertino switches don't render a thumb icon, keeps the shared SwitchNode shape.
- checkbox_recorder.dart: replaces the _checkmark / _dash / _maskedSymbol string constants with IconData constants; emits String.fromCharCode(icon.codePoint) and the icon's fontFamily instead of 'sans-serif'. _textScale raised 0.7 → 0.9.
- Tests updated to assert against the icon codepoints and the new thumb wireframe type (SRTextWireframe). Two new Switch cases with thumbIcon added to the golden harness; the four affected goldens regenerated.

### Review checklist


fuzzybinary (approved):
Yeah these look way better. Good work!

PR #1039, opened 2026-05-24
fix(gql_link): reuse TracingContext between RUM resource and wire headers

### LINKED ISSUE #1038: Should `DatadogGqlLink` reuse the outer `TracingContext` when injecting trace headers?
### Question

Hi team, first of all, thanks for maintaining this SDK. I ran into something while wiring up `datadog_gql_link` and I'm not sure whether it's a real issue or something I've misconfigured, so I wanted to ask before assuming anything.
  ## Setup
  We use Datadog RUM + APM with a Flutter app that talks to a GraphQL backend.
  **Datadog configuration:**
  ```dart
  configuration = DatadogConfiguration(
    clientToken: FennelEnv.getValue(FennelEnvValue.datadogClientToken),
    env: env,
    site: DatadogSite.us1,
    nativeCrashReportEnabled: true,
    loggingConfiguration: DatadogLoggingConfiguration(),
    rumConfiguration: DatadogRumConfiguration(
      applicationId: 'xxx',
    ),
    firstPartyHosts: ['x.com', 'x-dev.com'],
  )..enableHttpTracking(
      ignoreUrlPatterns: [RegExp(r'/graphql/?$')],
    );
  ```
  **GraphQL link:**
  ```dart
  link: Link.from([
    DatadogGqlLink(
      DatadogSdk.instance,
      Uri.parse('${BuildConfig.graphql_url}/'),
    ),
    DioLink('${BuildConfig.graphql_url}/', client: dio),
  ]),
  ```
  This matches the README recommendation, `DatadogGqlLink` above the terminating link, and `ignoreUrlPatterns` on `enableHttpTracking` so the native HTTP tracking does not double-report the `/graphql` endpoint.
  ## What I'm seeing
  **Without `DatadogGqlLink`** (plain `DioLink`, native HTTP tracking only): "View Trace in APM" from a `/graphql` RUM resource resolves correctly and shows the full backend hierarchy. The only thing I lose is the GraphQL-specific RUM attributes (operation name, type, etc.).
  > RUM resource → View Trace in APM showing the full backend hierarchy without DatadogGqlLink]_

Image

Image

  **With `DatadogGqlLink` added** (configured exactly as above, per the docs): the GraphQL attributes show up correctly on the RUM resource
  (operation name, type, variables, great!), but "View Trace in APM" now only shows the synthesized Flutter client span. The backend
  hierarchy is gone.

  > RUM resource → View Trace in APM showing only the client span after adding DatadogGqlLink]_

Image

Image

  ## What I noticed in the source

  While trying to understand what changed, I came across this in `datadog_gql_link.dart` (inside `_injectTracingHeaders`):

  ```dart
  // No tracing context, generate one ourselves
  final tracingContext = generateTracingContext(datadogSdk, rum);
  ```

  This is called from `DatadogGqlLink.request()`, which a few lines earlier already generated a `TracingContext` and used it to stamp the
  RUM resource:

  ```dart
  TracingContext? tracingContext;
  try {
    if (tracingHeaderTypes.isNotEmpty) {
      tracingContext = generateTracingContext(datadogSdk, rum);  // (A)
    }

    request = _injectTracingHeaders(request);                    // generates (B) inside
  } ...
  final resourceId = _startRumResource(
      request, internalAttributes, tracingContext, userAttributes); // stamps (A)
  ```
Since `generateTracingContext` calls `TracingId.traceId()`, which is freshly random per call with no cache, my reading is that the RUM resource gets stamped with one trace_id and the wire headers carry a different one. If that's right, the backend builds the APM trace under the wire trace_id, but RUM points "View Trace in APM" at the other one, so the link cannot resolve.
I might be misreading something here, which is why I wanted to ask rather than assume.
## What I tried locally
I patched `_injectTracingHeaders` to accept the outer `TracingContext` and prefer it over generating a new one:
  ```dart
  Request _injectTracingHeaders(Request request, TracingContext? outerContext) {
    ...
    final tracingContext =
        outerContext ?? generateTracingContext(datadogSdk, rum);
    ...
  }
  ```
  After this change, "View Trace in APM" resolves to the full backend hierarchy again, and the GraphQL attributes on the RUM resource are still there.

  > _[ RUM resource → View Trace in APM showing the full backend hierarchy after the patch]_

Image

## My questions

1. Is my reading of the two `generateTracingContext` calls correct, or is there something that ties them together that I'm missing?

2. I've opened a PR alongside this question so you can see the exact change. Happy to close the PR if my diagnosis is off, or to take a different approach if you'd prefer one.

3. If this isn't the right fix and I've misdiagnosed something, could you point me toward the path I should follow instead?
Happy to add more context, logs, or a minimal reproduction if it helps.

https://github.com/DataDog/dd-sdk-flutter/pull/1039

AcarFurkan:
### What and why?
`_injectTracingHeaders` was generating its own TracingContext instead of reusing the one already created in request(), so the RUM resource and the wire headers carried different trace_ids and "View Trace in APM" could not resolve to the backend trace.

### How?
`_injectTracingHeaders` now accepts the outer `TracingContext` from `request()` and reuses it when present, falling back to generating one only when the caller didn't (the existing call site skips generation when `tracingHeaderTypes` is empty). The change is API-additive the new parameter is nullable and the fallback preserves the previous behavior for any other caller. If needed, a description of how this PR accomplishes what it does.

Fix: #1038

### Review checklist


packages/datadog_gql_link/lib/src/datadog_gql_link.dart
```diff
@@ -229,7 +229,7 @@ class DatadogGqlLink extends Link {
     return resourceId;
   }
 
-  Request _injectTracingHeaders(Request request) {
+  Request _injectTracingHeaders(Request request, TracingContext? outerContext) {
```
fuzzybinary: I think this can be `TracingContext` (non-null).  This function should never be passed null and shouldn't generate its own tracing context.

packages/datadog_gql_link/test/datadog_gql_link_test.dart
```diff
@@ -1043,10 +1043,75 @@ query UserInfo($id: ID!) {
...
+        final rumSpanId = BigInt.parse(
+            capturedAttrs[DatadogRumPlatformAttributeKey.spanID] as String);
+        expect(rumTraceId, headerIds.traceId);
+        expect(rumSpanId, headerIds.spanId);
+      });
     });
   }
 }
 
+({BigInt traceId, BigInt spanId}) extractTraceAndSpanFromHeaders(
```
fuzzybinary, line 1248: You could use `TracingId.fromString` instead of this helper. If `TracingId` isn't already exported, feel free to export it from the main package in `lib/datadog_internal.dart`

fuzzybinary (changes requested):
Minor changes, but I'd like to get this in.

packages/datadog_gql_link/lib/src/datadog_gql_link.dart
AcarFurkan: Yep, sounds good

packages/datadog_gql_link/test/datadog_gql_link_test.dart
AcarFurkan: addressed

fuzzybinary (approved):

fuzzybinary:
@AcarFurkan I'm going to cherry pick this into a point release. I should hopefully be able to get that out tomorrow.

AcarFurkan:
That's amazing. Thanks for the quick review&response!

PR #1047, opened 2026-06-03
feat(datadog_flags): add precompute core

leoromanovsky:
## Motivation

Flutter needs a native Dart Datadog Flags package that can request precomputed assignments without bridging to the iOS or Android SDKs. The validated end-to-end POC in #1045 proves the full path, but it is too large to review as one change. This first stack PR isolates the transport and parsing foundation so reviewers can validate the precompute API boundary before typed evaluation, telemetry, persistence, and examples are layered on top.

## System Scope

This PR creates the package shell, internal precompute HTTP fetcher, request models, auth headers, site mapping, and fixture-backed parsing tests.

```mermaid
flowchart LR
    app[Flutter app] --> package[datadog_flags package]
    package --> facade[DatadogFlags facade and client]
    facade --> typed[Typed evaluation API]
    facade --> context[Evaluation context]
    context --> precompute[Precompute assignments fetcher]
    precompute --> api[Datadog precompute API]
    precompute --> models[Assignment models and parser]
    typed --> store[Assignment repository and persistence]
    typed --> rum[RUM feature flag context]
    typed --> exposure[Exposure EVP intake]
    typed --> flageval[Flag-evaluation EVP intake]
    example[Example app flag screen] --> app
    example --> counter[Forwarding request counters]
    counter --> precompute
    counter --> exposure
    counter --> flageval
    publishing[Pub package metadata and docs] --> package

    class package,models,precompute active;
```

## Stack Position

```mermaid
flowchart LR
    pr1047[#1047<br/>Precompute core] --> pr1048[#1048<br/>Typed API]
    pr1048 --> pr1049[#1049<br/>Exposures]
    pr1049 --> pr1050[#1050<br/>Flagevals]
    pr1050 --> pr1051[#1051<br/>Persistence]
    pr1051 --> pr1052[#1052<br/>Publishing]
    pr1052 --> pr1053[#1053<br/>Example app]

    class pr1047 current;
```

## Changes

This adds a `datadog_flags` package with the customer-facing shell for Datadog Feature Flags and Experimentation plus the internal precompute transport needed by later stack slices. The public barrel no longer exports the precompute fetcher, assignment model, or JSON helpers; those remain internal implementation details under `lib/src`.

The internal fetcher posts to `/precompute-assignments`, sends `Content-Type: application/vnd.api+json` plus `dd-client-token`, includes `dd-application-id` only when configured, and sends a `source` payload with `sdk_name` and `sdk_version` matching the iOS and Android request shape. It supports custom headers/endpoints for tests and dogfooding, allows evaluation contexts without a targeting key, scalarizes targeting attributes for the server contract, and maps Datadog sites including US1 staging.

The assignment parser treats missing or null variation values as malformed flag entries and drops those flags without dropping valid assignments. The package also mirrors the two canonical precompute fixture cases currently in `DataDog/ffe-system-test-data#12` so local tests can validate parser behavior while the shared fixture PR remains open. The mirrored Stack 1 fixtures now contain only precompute request/response data; later evaluation and telemetry expectations belong in the later stack PRs.

Validation:

- `dart run build_runner build --delete-conflicting-outputs` in `packages/datadog_flags`: passed and regenerated serializers; build_runner printed the existing analyzer/SDK-version warning
- `dart format --set-exit-if-changed packages/datadog_flags`: passed, 0 changed
- `dart analyze packages/datadog_flags`: passed, no issues
- `dart test` in `packages/datadog_flags`: passed, 12 tests
- `dart test --platform chrome` in `packages/datadog_flags`: passed, 10 browser-safe tests
- `flutter test test` in `packages/datadog_flags`: passed, 12 tests
- `git diff --check && git diff --cached --check`: passed

## Decisions

This PR intentionally does not include the typed flagging API, RUM feature-flag tagging, exposure emission, flagevaluation aggregation, persistence, publishing polish, or example-app UI. Those remain in later stack PRs so each review has a clear behavioral boundary.

The README is intentionally customer-facing and avoids documenting the internal precompute transport. Publishing remains a later stack concern; this PR keeps the package metadata valid without adding a publish guard that the repository release flow does not require.

The precompute fetcher remains internal rather than part of the public package barrel. The mirrored fixtures are a temporary local copy of the shared precompute fixture cases and should be replaced or linked to the shared fixture source once `DataDog/ffe-system-test-data#12` lands.


packages/datadog_flags/README.md
```diff
@@ -0,0 +1,59 @@
+# Datadog Flags for Flutter
+
+`datadog_flags` is a Dart-native package for Datadog feature flag assignments.
+This initial package is unpublished and contains the precompute transport layer
+that later stacked PRs will build into a complete Flutter flagging client.
+
+This package does not bridge to native iOS or Android flagging SDKs.
+
+## Precompute Fetching
```
leoromanovsky: this is completely not relevant for users - there should be no publish API related to actual precompute fetching. Customers only want to perform flag evaluation which might not be possible yet, that's ok.

leoromanovsky: this is completely not relevant for users - there should be no publish API related to actual precompute fetching. Customers only want to perform flag evaluation which might not be possible yet, that's ok.

leoromanovsky: this is completely not relevant for users - there should be no publish API related to actual precompute fetching. Customers only want to perform flag evaluation which might not be possible yet, that's ok.

packages/datadog_flags/README.md
```diff
@@ -0,0 +1,59 @@
...
+dart test
+```
+
+The included request example can make a real precompute call:
+
+```bash
+DD_CLIENT_TOKEN=<client-token> \
+DD_ENV=staging \
+DD_TARGETING_KEY=test-subject \
+dart run example/precompute_request.dart
```
leoromanovsky: this readme should be helpful to CUSTOMERs

packages/datadog_flags/example/precompute_request.dart
```diff
@@ -0,0 +1,72 @@
...
+  if (clientToken == null || clientToken.isEmpty) {
+    stderr.writeln('Set DD_CLIENT_TOKEN before running this example.');
+    exitCode = 64;
+    return;
+  }
+
+  final customEndpoint = Platform.environment['DD_FLAGS_ENDPOINT'];
+  final client = http.Client();
+  try {
+    final fetcher = FlagAssignmentsFetcher(
```
leoromanovsky: if this shouldn't be public, then remove this example it's ok not to have one

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,124 @@
...
+  return switch (value?.toLowerCase()) {
+    'boolean' => FlagVariationType.boolean,
+    'bool' => FlagVariationType.boolean,
+    'string' => FlagVariationType.string,
+    'integer' => FlagVariationType.integer,
+    'int' => FlagVariationType.integer,
+    'float' => FlagVariationType.float,
+    'double' => FlagVariationType.float,
+    'object' => FlagVariationType.object,
+    'json' => FlagVariationType.object,
```
leoromanovsky: that's pretty verbose where'd you get that?

packages/datadog_flags/lib/src/datadog_context.dart
```diff
@@ -0,0 +1,45 @@
...
+
+enum DatadogFlagsSite {
+  us1,
+  us3,
+  us5,
+  eu1,
+  ap1,
+  ap2,
+  us1Fed,
+}
```
leoromanovsky, line 22: does this exist in the repo already somewhere? if not can it be moved to a more central location?

packages/datadog_flags/lib/src/datadog_context.dart
```diff
@@ -0,0 +1,45 @@
...
+// Copyright 2019-Present Datadog, Inc.
+
+enum DatadogFlagsSite {
+  us1,
+  us3,
+  us5,
+  eu1,
+  ap1,
+  ap2,
+  us1Fed,
```
leoromanovsky: no fed support

packages/datadog_flags/lib/src/datadog_context.dart
```diff
@@ -0,0 +1,45 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+enum DatadogFlagsSite {
+  us1,
```
leoromanovsky: also add staging support for us1

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,122 @@
...
+import 'package:http/http.dart' as http;
+
+import 'assignment.dart';
+import 'datadog_context.dart';
+import 'flags_configuration.dart';
+import 'flags_context.dart';
+import 'flags_error.dart';
+import 'json_value.dart';
+
+class FlagAssignmentsFetcher {
```
leoromanovsky, line 18: this should be private

packages/datadog_flags/lib/src/flags_context.dart
```diff
@@ -0,0 +1,34 @@
...
+      targetingKey: json['targetingKey'] as String,
+      attributes: Map<String, Object?>.from(
+        json['attributes'] as Map<String, Object?>? ?? const {},
+      ),
+    );
+  }
+
+  Map<String, Object?> toJson() {
+    return {
+      'targetingKey': targetingKey,
```
leoromanovsky: P0: it's ok to not have a targeting key

packages/datadog_flags/lib/src/json_value.dart
```diff
@@ -0,0 +1,43 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+Object? sanitizeJsonValue(Object? value) {
```
leoromanovsky, line 12: what does this do? is it necessary?

packages/datadog_flags/test/fixtures/precomputed/cases/all-types-success.json
```diff
@@ -0,0 +1,133 @@
...
+              ]
+            },
+            "reason": "TARGETING_MATCH",
+            "doLog": true
+          }
+        }
+      }
+    }
+  },
+  "evaluations": [
```
leoromanovsky: what is this list? should it be a separate file?

packages/datadog_flags/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: datadog_flags
+description: Native Dart precompute client for Datadog feature flag assignments.
```
leoromanovsky: This is the wrong description; we are building a product for feature flags and experimentation.

packages/datadog_flags/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
+name: datadog_flags
+description: Datadog Feature Flags and Experimentation SDK for Flutter.
```
fuzzybinary: This doesn't actually specify a Flutter requirement. Will it in the future? Could this be used in just stand alone Dart apps?

packages/datadog_flags/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
...
+version: 0.0.1
+publish_to: none
```
fuzzybinary: This should be removed. We intend on publishing this to pub, and there's no harm in removing it while we're still developing it. It will only publish when we manually do so.

packages/datadog_flags/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+## 0.0.1
+
+- Add an unpublished Dart package scaffold for Datadog Flags.
```
fuzzybinary: Our release script actually generates the CHANGELOG based on Conventional Commits so that we don't have to remember to add things to the CHANGELOG manually.

packages/datadog_flags/test/precomputed_fixture_files_test.dart
```diff
@@ -0,0 +1,118 @@
...
+
+import 'package:datadog_flags/datadog_flags.dart';
+import 'package:datadog_flags/src/assignment.dart';
+import 'package:datadog_flags/src/flag_assignments_fetcher.dart';
+import 'package:http/http.dart' as http;
+import 'package:http/testing.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('mirrored canonical precompute fixture files load as JSON', () {
```
fuzzybinary: Is this a test just for making sure that the test case files are valid JSON?

packages/datadog_flags/test/precomputed_fixture_files_test.dart
```diff
@@ -0,0 +1,118 @@
...
+    final casesDirectory = [
+      Directory('test/fixtures/precomputed/cases'),
+      Directory('packages/datadog_flags/test/fixtures/precomputed/cases'),
```
fuzzybinary: I'm assuming this was added because something was failing depending on where the tests were run, but probably only handles if the tests are run from the root of the repo or from the package itself. We should likely find a more resilient way to do this if we want to keep this test.

packages/datadog_flags/test/precomputed_fixture_files_test.dart
```diff
@@ -0,0 +1,118 @@
...
+      'defaults-and-emission-gates.json',
+    ]);
+
+    for (final file in files) {
+      final decoded = jsonDecode(file.readAsStringSync());
+      expect(decoded, isA<Map<String, Object?>>());
+    }
+  });
+
+  test('mirrored canonical precompute fixtures parse as assignments', () async {
```
fuzzybinary: The better way to do this is to generate a test for each case, not have a single test for all cases. So:
```dart
for (final fixture in _fixtureCases()) {
  test('mirrored canonical fixture ${fixture} parses as assignment', () async {
```
This allows you to have a test failure for each file that fails instead of one that fails if any of them mismatch.

packages/datadog_flags/test/precomputed_fixture_files_test.dart
```diff
@@ -0,0 +1,118 @@
...
+    ),
+  );
+}
+
+Map<String, Object?> _flagsFrom(Map<String, Object?> fixture) {
+  final response = fixture['response'] as Map<String, Object?>;
+  final data = response['data'] as Map<String, Object?>;
+  final attributes = data['attributes'] as Map<String, Object?>;
+  return attributes['flags'] as Map<String, Object?>;
+}
```
fuzzybinary: In may be worth bringing in `json_serializable` to parse these bodies for you.

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'json_value.dart';
+
+enum FlagVariationType { boolean, string, integer, float, object, unknown }
+
+class FlagAssignment {
```
fuzzybinary: This class can probably be flagged as `@immutable`

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+String variationTypeToString(FlagVariationType type) {
+  return switch (type) {
+    FlagVariationType.boolean => 'boolean',
+    FlagVariationType.string => 'string',
+    FlagVariationType.integer => 'integer',
+    FlagVariationType.float => 'float',
+    FlagVariationType.object => 'object',
+    FlagVariationType.unknown => 'unknown',
+  };
+}
```
fuzzybinary: enums support methods and factory constructors in Dart. So these can be added as members:

```dart
enum FlagVariationType {
  boolean,
  stirng,
  integer,
 // etc,
  unknown;

  factory FlagVariationType.fromString(String str) {
    switch(str) {
      // cases
    }
  }
  
  String toShortString() {
    switch(this) {
      // cases
    } 
  } 
}
```

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'json_value.dart';
+
+enum FlagVariationType { boolean, string, integer, float, object, unknown }
```
dd-oleksii: what's `unknown` for?

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    return switch (requestedType) {
+      FlagVariationType.boolean when variationValue is bool => variationValue,
+      FlagVariationType.string when variationValue is String => variationValue,
+      FlagVariationType.integer when variationValue is int => variationValue,
+      FlagVariationType.float when variationValue is double => variationValue,
+      FlagVariationType.object => variationValue,
+      _ => null,
+    };
```
dd-oleksii: minor: this switch suggest that it is possible to create a `FlagAssignment` with a value that doesn't match its type. It's better to move this type check to constructor or provide named constructors that accept properly-typed values:
```dart
class FlagAssignment {
  // ...
  FlagAssignment.integer({
    required this.allocationKey,
    required this.variationKey,
    required integer variationValue, // <-- specific type
    required this.reason,
    required this.doLog,
  }) :
    variationType = FlagVariationType.integer,
    variationValue = variationValue;
```

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  static const defaultAssignment = FlagAssignment(
+    allocationKey: '',
+    variationKey: '',
+    variationType: FlagVariationType.unknown,
+    variationValue: null,
+    reason: 'DEFAULT',
```
dd-oleksii: minor: looks dangerous. I think "default" assignment should be an error. Also not sure why we even need a default

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+    doLog: false,
+  );
+}
+
+FlagVariationType normalizeVariationType(String? typeName, Object? value) {
+  final normalizedType = typeName?.toLowerCase();
+  if (normalizedType == 'number') {
```
dd-oleksii:
```suggestion
  if (typeName == 'number') {
```

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (value is num) {
+      return FlagVariationType.float;
+    }
+  }
+  return variationTypeFromString(typeName);
+}
+
+FlagVariationType variationTypeFromString(String? value) {
+  return switch (value?.toLowerCase()) {
```
dd-oleksii: minor: no need to lowercase
```suggestion
  return switch (value) {
```

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+
+FlagVariationType normalizeVariationType(String? typeName, Object? value) {
+  final normalizedType = typeName?.toLowerCase();
+  if (normalizedType == 'number') {
+    if (value is int) {
+      return FlagVariationType.integer;
+    }
+    if (value is num) {
+      return FlagVariationType.float;
+    }
```
dd-oleksii: major: I don't think we should remap the type returned from the server based on the value. This may lead to issues.

Imagine a float flag with variants `1.5` and `2` — depending on the resolved value, it may flip-flop between integer and float types.

If the server returns `number`, we should keep it as is and do the check at runtime. If user asks for int, we should check the value. If user asks for float, we can always return the value we have.

packages/datadog_flags/lib/src/assignment.dart
dd-oleksii: major: add `number` here

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,122 @@
...
+  final DatadogFlagsConfiguration configuration;
+  final http.Client httpClient;
+
+  FlagAssignmentsFetcher({
+    required this.datadogContext,
+    required this.configuration,
+    required this.httpClient,
+  });
+
+  Future<Map<String, FlagAssignment>> fetch(
```
dd-oleksii: minor: can we return the full response data here? this is currently missing `createdAt` and `environment`

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,120 @@
...
+}
+
+FlagVariationType variationTypeFromString(String? value) {
+  return switch (value?.toLowerCase()) {
+    'boolean' => FlagVariationType.boolean,
+    'string' => FlagVariationType.string,
+    'integer' => FlagVariationType.integer,
+    'float' => FlagVariationType.float,
+    'object' => FlagVariationType.object,
+    _ => FlagVariationType.unknown,
```
dd-oleksii: minor: throw?

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,122 @@
...
+    try {
+      final decoded = _asObject(jsonDecode(response.body), 'response');
+      final data = _asObject(decoded['data'], 'data');
+      final attributes = _asObject(data['attributes'], 'attributes');
+      final flags = _asObject(attributes['flags'], 'flags');
+      final assignments = <String, FlagAssignment>{};
+      for (final entry in flags.entries) {
+        final assignment = _parseAssignment(entry.value);
+        if (assignment == null ||
+            assignment.variationType == FlagVariationType.unknown) {
```
dd-oleksii: minor: can we throw instead of returning `unknown`?

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,122 @@
...
+  throw FormatException('$name must be a JSON object');
+}
+
+FlagAssignment? _parseAssignment(Object? value) {
+  try {
+    return FlagAssignment.fromJson(_asObject(value, 'flag assignment'));
+  } catch (_) {
+    return null;
+  }
+}
```
dd-oleksii: nitpick: if we throw instead of returning `unknown`, we can inline this function for more obvious control flow in `fetch()`:
```dart
  try {
    return FlagAssignment.fromJson(_asObject(entry.value, 'flag assignment'));
  } catch (_) {
    continue;
  }
```

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,122 @@
...
+    return Map<String, Object?>.from(value);
+  }
+  throw FormatException('$name must be a JSON object');
+}
+
+FlagAssignment? _parseAssignment(Object? value) {
+  try {
+    return FlagAssignment.fromJson(_asObject(value, 'flag assignment'));
+  } catch (_) {
+    return null;
```
dd-oleksii: minor: worth logging the error?

packages/datadog_flags/lib/src/json_value.dart
```diff
@@ -0,0 +1,38 @@
...
+  if (value == null ||
+      value is String ||
+      value is bool ||
+      value is int ||
+      value is double) {
+    return value;
+  }
+  if (value is num) {
+    return value.toDouble();
+  }
```
dd-oleksii: minor: `num` is only implemented by `int` and `double`, so this seems redundant?
```suggestion
```

packages/datadog_flags/lib/datadog_flags.dart
```diff
@@ -0,0 +1,14 @@
...
+
+export 'src/datadog_context.dart';
+export 'src/flags_configuration.dart';
+export 'src/flags_context.dart';
+export 'src/flags_error.dart';
+export 'src/flags_site.dart';
+
+import 'src/flags_context.dart';
+
+typedef FlagsEvaluationContext = DatadogFlagsEvaluationContext;
```
dd-oleksii: nitpick: why not call it `FlagsEvaluationContext` directly?

packages/datadog_flags/.gitignore
```diff
@@ -0,0 +1,31 @@
+# Miscellaneous
+*.class
+*.log
+*.pyc
+*.swp
+.DS_Store
+.atom/
+.buildlog/
+.history
+.svn/
```
dd-oleksii, line 10: aside: glad we're excluding svn repos 😆

packages/datadog_flags/lib/datadog_flags.dart
```diff
@@ -0,0 +1,14 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+export 'src/datadog_context.dart';
+export 'src/flags_configuration.dart';
+export 'src/flags_context.dart';
+export 'src/flags_error.dart';
+export 'src/flags_site.dart';
```
dd-oleksii: minor: do we actually want to export all of this? (I'd prefer using `export '...' show ...;` form to more tightly control what is our public surface and what's not)

packages/datadog_flags/lib/src/assignment.dart
fuzzybinary: `decodeVariationValue` could also possibly be a member.

packages/datadog_flags/lib/src/assignment.dart
fuzzybinary: It's actually better for our client SDKs to not throw, as an uncaught throw could potentially crash the application.

packages/datadog_flags/lib/src/datadog_context.dart
```diff
@@ -0,0 +1,42 @@
...
+        ),
+      DatadogFlagsSite.eu1 => Uri.parse('https://preview.ff-cdn.datadoghq.eu'),
+      DatadogFlagsSite.ap1 => Uri.parse(
+          'https://preview.ff-cdn.ap1.datadoghq.com',
+        ),
+      DatadogFlagsSite.ap2 => Uri.parse(
+          'https://preview.ff-cdn.ap2.datadoghq.com',
+        ),
+    };
+  }
```
fuzzybinary: Dart supports "Enhanced Enums". https://dart.dev/language/enums#declaring-enhanced-enums

I think it makes sense to store the endpoint with the num itself which avoids this switch.

packages/datadog_flags/lib/src/datadog_context.dart
```diff
@@ -0,0 +1,42 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'flags_site.dart';
+
+class DatadogFlagsContext {
```
fuzzybinary: I think this class and other data classes can be marked `@immutable` and likely also `final`.

I'm not great at remembering to put `final` on classes but it's also a fairly recent addition to Dart.

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,122 @@
...
+        'Unexpected flag assignments response status ${response.statusCode}.',
+      );
+    }
+
+    try {
+      final decoded = _asObject(jsonDecode(response.body), 'response');
+      final data = _asObject(decoded['data'], 'data');
+      final attributes = _asObject(data['attributes'], 'attributes');
+      final flags = _asObject(attributes['flags'], 'flags');
+      final assignments = <String, FlagAssignment>{};
```
fuzzybinary: I think it makes sense to use [json_serializable](https://pub.dev/packages/json_serializable) for this.

packages/datadog_flags/lib/src/flags_error.dart
```diff
@@ -0,0 +1,40 @@
...
+  clientNotInitialized,
+  invalidConfiguration,
+}
+
+class FlagsException implements Exception {
+  final FlagsErrorType type;
+  final String message;
+  final Object? cause;
+
+  const FlagsException(this.type, this.message, [this.cause]);
```
fuzzybinary: In my experience, having `cause` be an optional named parameter instead of an optional positional parameter will save you a lot of headaches if the signature of this class has to change in the future.

packages/datadog_flags/lib/src/flags_site.dart
```diff
@@ -0,0 +1,6 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+enum DatadogFlagsSite { us1, us1Staging, us3, us5, eu1, ap1, ap2 }
```
fuzzybinary: I don't think this has to be in a separate file. It can be held in the configuration file.

packages/datadog_flags/lib/datadog_flags.dart
fuzzybinary: Agreed. If we're just going to typedef it on export I'd rather have it name consistently internally.

packages/datadog_flags/test/precompute_core_test.dart
```diff
@@ -0,0 +1,343 @@
...
+import 'package:http/http.dart' as http;
+import 'package:http/testing.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('FlagAssignmentsFetcher', () {
+    test(
+      'builds the precompute request expected by canonical fixtures',
+      () async {
+        final fixture = _requestFixture();
```
fuzzybinary: Since this is always the same and always final, you can make it a member of the group:

```dart
group('FlagAssignmentsFetcher', () {
  final requestFixture = {
    'context': {
      'targetingKey': 'precomputed-user',
      'attributes': {'plan': 'pro', 'platform': 'flutter'},
    },
    'response': {'data': _emptyAssignments()},
  };
  
  test('etc, () {
```

packages/datadog_flags/test/precompute_core_test.dart
```diff
@@ -0,0 +1,343 @@
...
+DatadogFlagsContext _contextFor(DatadogFlagsSite site) {
+  return DatadogFlagsContext(clientToken: 'token', env: 'dev', site: site);
+}
+
+http.Client _jsonClient(List<http.Request> requests, Object? body) {
+  return MockClient((request) async {
+    requests.add(request);
+    return http.Response(jsonEncode(body), 200);
+  });
+}
```
fuzzybinary: I'm not really a fan of a lot of these helpers that just return another object. I'd rather be more explicit in the test, which will give us more control of the tests in the future without touching other tests, and keeps more of the context local to the test.

You also could use a mocking library for the MockClient. My preferred is [mocktail](https://pub.dev/packages/mocktail).

packages/datadog_flags/test/precompute_core_test.dart
```diff
@@ -0,0 +1,343 @@
...
+            'env': {'dd_env': 'dev'},
+            'subject': {
+              'targeting_attributes': {'companyId': '1'},
+            },
+          },
+        },
+      });
+    });
+
+    test('maps all supported sites including US1 staging', () {
```
fuzzybinary:
```suggestion
    test('maps all supported sites', () {
```

packages/datadog_flags/test/precompute_core_test.dart
```diff
@@ -0,0 +1,343 @@
...
+          datadogContext: _contextFor(DatadogFlagsSite.us1),
+          configuration: const DatadogFlagsConfiguration(),
+          httpClient: MockClient((_) async => http.Response('nope', 500)),
+        );
+
+        await expectLater(
+          nonSuccessFetcher.fetch(
+            const DatadogFlagsEvaluationContext(targetingKey: 'subject'),
+          ),
+          throwsA(
```
fuzzybinary, line 322: Do we intend to have a wrapper that clients will use that doesn't throw? Or are we expecting clients to catch these?

packages/datadog_flags/test/precomputed_fixture_files_test.dart
```diff
@@ -0,0 +1,118 @@
...
+    }
+  });
+}
+
+List<Map<String, Object?>> _fixtureCases() {
+  final casesDirectory = [
+    Directory('test/fixtures/precomputed/cases'),
+    Directory('packages/datadog_flags/test/fixtures/precomputed/cases'),
+  ].firstWhere((directory) => directory.existsSync());
+
```
fuzzybinary: This is essentially duplicate logic from above. If we want to keep the first test, these should be merged.

And also we should find a better way to get the cases Directory that's not reliant on the current working directory.

packages/datadog_flags/pubspec.yaml
fuzzybinary: Stand alone Dart apps aren't common, but if we can avoid a Flutter dependency, it might be worth it. In which case, this should specify it is for Client applications (not server applications).

packages/datadog_flags/README.md
```diff
@@ -0,0 +1,34 @@
+# Datadog Flags for Flutter
+
+`datadog_flags` is the native Flutter SDK for Datadog Feature Flags and
+Experimentation. It lets Flutter applications evaluate Datadog-backed feature
```
fuzzybinary: Flutter mentioned here but as of right now, we don't have a dependency on Flutter or the Flutter SDK. We should talk about whether that can stay the case or not.

packages/datadog_flags/README.md
```diff
@@ -0,0 +1,34 @@
...
+flags without bridging through the native iOS or Android flagging SDKs.
+
+This package does not bridge to native iOS or Android flagging SDKs.
```
fuzzybinary:
```suggestion
flags.
```

I don't think we need to explicitly mention this in the summary, let alone twice.  It's fine to have in a "Features" or "Details" section.

packages/datadog_flags/README.md
```diff
@@ -0,0 +1,34 @@
...
+- Supports Datadog US1 staging for dogfooding and local validation.
+
+## Local Validation
+
+From this package:
+
+```bash
+dart analyze .
+dart test
+```
```
fuzzybinary: Move to CONTRIBUTING.md.

This is fine for now, but we'll want to use `melos` at the root of the package in the long term. Contributors should also read the CONTRIBUTING.md at the root of the package as it has the generalized guidelines for contributing to our packages.

packages/datadog_flags/README.md
```diff
@@ -0,0 +1,34 @@
...
+Customer-facing usage documentation will live here once the typed flag
+evaluation API lands.
+
+## Behavior
+
+- Native Dart implementation for Flutter applications.
+- No runtime bridge to the native iOS or Android flagging SDKs.
+- Designed for Datadog Feature Flags and Experimentation, including local typed
+  flag evaluation and SDK telemetry in later stack slices.
+- Supports Datadog US1 staging for dogfooding and local validation.
```
fuzzybinary: I think this is getting a bit too implementation and process specific for a README that's meant for customer consumption. Some of this fits but some of it isn't really needed.

fuzzybinary (changes requested):
A bunch of places I think we can improve, especially if we use extended enums and extension methods, we can remove a bunch of the stand alone helpers that are scattered about.

I would also seriously think about using `json_serializable` for more robust json serialization / deserialization in many places. It does add a dependency but I think it's worth it in this case.

OliviaShoup (dismissed):

packages/datadog_flags/.gitignore
leoromanovsky: It's coming back in style!

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -3,19 +3,70 @@
...
+  float('float'),
+  number('number'),
+  object('object');
+
+  final String wireName;
+
+  const FlagVariationType(this.wireName);
+
+  factory FlagVariationType.fromWireName(String? value) {
+    return switch (value) {
```
fuzzybinary: If you set them up like this, you can now actually enumerate over the values looking for one that matches.

```dart
  factory FlagVariationType.fromWireName(String name) {
    for(final e in values) {
      if (e.wireName == name) return e;
    }
    throw FormatException('Unsupported flag variation type: $value')
  }
```
As null as an invalid value, I would change the signature to not accept an optional string.

packages/datadog_flags/test/precomputed_fixture_files_test.dart
```diff
@@ -12,80 +12,92 @@ import 'dart:io';
...
+  );
+  return jsonDecode(file.readAsStringSync()) as Map<String, Object?>;
+}
+
+Directory _packageRoot() {
+  if (Platform.script.scheme == 'file') {
+    final scriptRoot = _findPackageRoot(File.fromUri(Platform.script).parent);
+    if (scriptRoot != null) {
+      return scriptRoot;
     }
```
fuzzybinary: This whole series of functions led me down an incredible rabbit hole of how the hell you're supposed to find content files when doing Dart testing and... wow is there no good answer, which is amazing.

However, since these tests likely won't work for web / wasm anyway, it might be worth trying this to get the location of the root of this package on disk, which should be reliable.

```dart
var uri = await Isolate.resolvePackageUri(Uri.parse('package:datadog_flags/datadog_flags.data'));
```

fuzzybinary (commented):
First pass through new changes, I want to do another pass through everything.

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,130 @@
...
+  const FlagAssignment._({
+    required this.allocationKey,
+    required this.variationKey,
+    required this.variationType,
+    required this.variationValue,
+    required this.reason,
+    required this.doLog,
+  });
+
+  factory FlagAssignment.fromJson(Map<String, Object?> json) {
```
fuzzybinary, line 86: Any reason to not use `json_serializable` for this as well?

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,130 @@
...
+      'allocationKey': allocationKey,
+      'variationKey': variationKey,
+      'variationType': variationType.wireName,
+      'variationValue': sanitizeJsonValue(variationValue),
+      'reason': reason,
+      'doLog': doLog,
+    };
+  }
+
+  Object? typedValue(FlagVariationType requestedType) {
```
fuzzybinary: I agree with a previous comment about this method. It's doing a *lot* of work that's fairly confusing because it assumes it's possible to create a FlagAssignment with a `variationType` that doesn't match its `variationValue`.

Since everything in this class is final, we can make `fromJson` fail if it attempts to create a FlagAssignment that is invalid, either by having it throw or by changing it to a static function that returns null on an invalid value. We then potentially don't need this method at all, and `int -> double` conversions can be handled by the client.

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
```diff
@@ -0,0 +1,111 @@
...
+        cause: error,
+      );
+    }
+  }
+
+  Map<String, String> _headers() {
+    return {
+      'Content-Type': 'application/vnd.api+json',
+      'dd-client-token': datadogContext.clientToken,
+      if (datadogContext.applicationId != null)
```
fuzzybinary: nit and a bit of a weird syntax but just FYI:
If you wanted to avoid the force unwrap (!), Dart has a very strange syntax for doing local asignment:
```dart
if (datadogContext.applicationId case final applicationId?)
  'dd-application-id': applicationId
```

packages/datadog_flags/lib/src/flags_context.dart
```diff
@@ -0,0 +1,38 @@
...
+
+  const FlagsEvaluationContext({
+    this.targetingKey,
+    this.attributes = const {},
+  });
+
+  factory FlagsEvaluationContext.fromJson(Map<String, Object?> json) {
+    final targetingKey = json['targetingKey'];
+    return FlagsEvaluationContext(
+      targetingKey: targetingKey is String ? targetingKey : null,
```
fuzzybinary: I know it's very simple but this could also be done with JsonSerializable.

packages/datadog_flags/lib/src/json_value.dart
fuzzybinary: It's checking if all the values in the object are serializable to Json by `dart:convert` without requiring a secondary serialization library. It's similar to [this function](https://github.com/DataDog/dd-sdk-flutter/blob/1467e9ebf927c21dea7895f3e5e8a386105cd1d4/packages/datadog_flutter_plugin/lib/src/helpers.dart#L153) which is used in the main library after the fact to check for bad values.

fuzzybinary (commented):
This mostly looks good, but I do think we should really have a long think about the `typedValue` method. I think it's likely better to completely avoid bad values in FlagAssignments than to check them on every fetch, and it would make the code way simpler.

packages/datadog_flags/lib/src/flag_assignments_fetcher.dart
leoromanovsky: Thanks for catching yet; yes agreed this is not a good pattern in any language.

packages/datadog_flags/lib/src/assignment.dart
leoromanovsky: Cool; will use the tool in all the places.

packages/datadog_flags/test/precomputed_fixture_files_test.dart
leoromanovsky: Thank you for doing that!

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,103 @@
...
+      FlagVariationType.integer when value is int => value,
+      FlagVariationType.float when value is num => value.toDouble(),
+      FlagVariationType.number when value is num => value,
+      FlagVariationType.object => sanitizeJsonValue(value),
+      _ => throw FormatException('Invalid variation value for $wireName'),
+    };
+  }
+}
+
+String _variationTypeToJson(FlagVariationType type) => type.wireName;
```
fuzzybinary: Is this needed? I think it can be lined in the `@JsonKey` attribute.

packages/datadog_flags/lib/src/flags_context.dart
```diff
@@ -0,0 +1,42 @@
...
+part 'flags_context.g.dart';
+
+@immutable
+@JsonSerializable()
+final class FlagsEvaluationContext {
+  static const empty = FlagsEvaluationContext();
+
+  @JsonKey(includeIfNull: false)
+  final String? targetingKey;
+  @JsonKey(fromJson: _attributesFromJson, toJson: sanitizeJsonValue)
```
fuzzybinary: I'm not sure the `fromJson` here is necessary? I think JsonSerializable does the conversion properly, and you can specify a default value with `defaultValue`. It's worth checking anyway.

fuzzybinary (commented):
Two very minor nits, just cleaning up some internal methods that I think might be overly complicated. I'm also going to have Codex do a pass before approving.

fuzzybinary:
@codex review

packages/datadog_flags/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+dependencies:
+  http: ^1.0.0
+  json_annotation: ^4.9.0
+  meta: ^1.0.0
+
+dev_dependencies:
+  build_runner: 2.4.15
+  json_serializable: 6.9.5
+  lints: ^6.0.0
```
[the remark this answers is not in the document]
fuzzybinary: This is a good catch.

@leoromanovsky Change `lints` to be '>=5.0.0', even though this is a dev dependency. We do minimum version checks on the nightlies and having this at ^6 will break those min version checks.

packages/datadog_flags/test/precompute_core_test.dart
leoromanovsky: Public typed evaluation should not throw for flag decisions. The next API layer should catch provider/fetch/parser failures internally, record the error reason for telemetry, and return the caller-provided default value. These fetcher exceptions are internal signals for that wrapper, not something customers should catch.

leoromanovsky:
@codex review

fuzzybinary (dismissed):
LGTM! Thanks for addressing everything!

packages/datadog_flags/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
+name: datadog_flags
```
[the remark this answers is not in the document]
leoromanovsky: We decided not to.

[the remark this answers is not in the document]
fuzzybinary: We also never run melos publish over the entire repo so this is fine.

packages/datadog_flags/lib/src/assignment.dart
```diff
@@ -0,0 +1,103 @@
...
+final class FlagAssignment {
+  final String allocationKey;
+  final String variationKey;
+  @JsonKey(
+    fromJson: FlagVariationType.fromWireName,
+    toJson: FlagVariationType.toWireName,
+  )
+  final FlagVariationType variationType;
+  @JsonKey(toJson: sanitizeJsonValue)
+  final Object? variationValue;
```
dd-oleksii: why is this optional?

packages/datadog_flags/lib/src/datadog_context.dart
```diff
@@ -0,0 +1,39 @@
...
+
+  final String flagsEndpointUrl;
+
+  const DatadogFlagsSite(this.flagsEndpointUrl);
+
+  Uri get flagsEndpoint => Uri.parse(flagsEndpointUrl);
+}
+
+@immutable
+final class DatadogFlagsContext {
```
dd-oleksii: minor: is "config" a better name? `DatadogFlagsContext` is a weird name

packages/datadog_flags/lib/src/precompute_request.dart
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  Map<String, Object?> toJson() => _$PrecomputeRequestDataToJson(this);
+}
+
+@immutable
+@JsonSerializable(createFactory: false, explicitToJson: true)
+final class PrecomputeRequestAttributes {
+  final PrecomputeRequestEnv env;
+  final PrecomputeRequestSubject subject;
```
dd-oleksii, line 65: minor: can we add `source` here? (sdk name + version)

packages/datadog_flags/lib/src/precompute_request.dart
```diff
@@ -0,0 +1,95 @@
...
+  Map<String, Object?> toJson() => _$PrecomputeRequestEnvToJson(this);
+}
+
+@immutable
+@JsonSerializable(createFactory: false)
+final class PrecomputeRequestSubject {
+  @JsonKey(name: 'targeting_key', includeIfNull: false)
+  final String? targetingKey;
+  @JsonKey(name: 'targeting_attributes', toJson: sanitizeJsonValue)
+  final Map<String, Object?> targetingAttributes;
```
dd-oleksii, line 109: minor: jfyi, the server does not currently accept nested attributes. only strings, numbers, booleans and nulls are allowed

packages/datadog_flags/lib/src/evaluation_context.dart
dd-oleksii, line 1: minor: rename the file to `evaluation_context.dart`. "flags context" is too ambiguous

dd-oleksii (dismissed):

packages/datadog_flags/lib/src/evaluation_context.dart
leoromanovsky: Done in `c97d72e9`: renamed `flags_context.dart` / generated part to `evaluation_context.dart` while keeping the public `FlagsEvaluationContext` type.

packages/datadog_flags/lib/src/precompute_request.dart
leoromanovsky: Adjusted in `1e923eee`: `targeting_attributes` now serialize through a scalar-only sanitizer. Strings/numbers/booleans/nulls pass through; nested objects/lists are JSON-encoded into strings so the request body stays within the server contract. Added request-body coverage for this.

packages/datadog_flags/lib/src/precompute_request.dart
leoromanovsky: Done in `09c2a27c`: the precompute body now includes `attributes.source` with `sdk_name` and `sdk_version`, matching the iOS/Android request shape. The request test asserts `dd-sdk-flutter` plus the configured SDK version.

fuzzybinary (dismissed):

fuzzybinary (approved):

typotter (approved):

OliviaShoup (approved):

PR #1048, opened 2026-06-03
feat(datadog_flags): add typed API

leoromanovsky:
## Motivation

Stack 1 can fetch and parse precomputed assignments, but application code still needs a small, predictable flagging API. This PR adds the Dart-native API layer so client applications can initialize a flag client for one evaluation context and synchronously resolve typed flag details with caller-provided defaults.

The main design constraint from Stack 1 review is that `datadog_flags` should stay plain Dart. RUM integration, exposure intake, flagevaluation batching, persistence, and Flutter example UI remain later stack items so this PR can focus on public evaluation semantics.

## System Scope

This PR implements the highlighted typed API layer and leaves telemetry and app integration to later stacks.

```mermaid
flowchart LR
    app["Dart client app"] --> package["datadog_flags package"]
    package --> facade["DatadogFlags singleton facade"]
    facade --> client["DatadogFlagsClient"]
    client --> context["FlagsEvaluationContext"]
    context --> init["initialize(context)"]
    init --> fetcher["Precompute assignments fetcher"]
    fetcher --> api["Datadog precompute API"]
    fetcher --> models["Generated request/response models"]
    client --> state["Client-owned assignment state"]
    state --> details["Typed FlagDetails + default errors"]
    shutdown["shutdown()"] --> state
    cli["Standalone Dart CLI example"] --> facade

    exposure["Exposure EVP intake"] -. later .-> package
    flageval["Flag-evaluation EVP intake"] -. later .-> package
    persistence["Last-known assignment persistence"] -. later .-> state
    flutter["Flutter dogfooding app"] -. later .-> package
    rum["RUM feature flag integration"] -. later integration .-> flutter

    class facade,client,context,init,state,details,shutdown,cli active;
```

## OpenFeature Compatibility Design

This PR intentionally shapes the Dart API so an async OpenFeature provider extension can wrap it later without redesigning the core client. The goal is not to depend on an OpenFeature Dart package in Stack 2; it is to preserve the same concepts and method boundaries used by OpenFeature and the existing Datadog client SDKs.

| Concept | OpenFeature / Datadog precedent | Dart API in this PR |
| --- | --- | --- |
| Provider lifecycle | OpenFeature providers expose lifecycle boundaries for initialization and shutdown. | [`DatadogFlagsClient`](https://github.com/DataDog/dd-sdk-flutter/blob/1998e573d300eae060b7dc5211453f6a2ae44c01/packages/datadog_flags/lib/src/flags_client.dart#L14-L46) exposes `initialize(context)` as the async precompute boundary and `shutdown()` as the client cleanup boundary. |
| Detailed evaluation | OpenFeature detailed evaluation returns value plus variant, reason, and error fields. The JS browser provider returns `ResolutionDetails<T>` from `resolve*Evaluation` methods, and the Swift provider converts Datadog `FlagDetails<T>` into OpenFeature `ProviderEvaluation<T>`. | [`FlagDetails<T>`](https://github.com/DataDog/dd-sdk-flutter/blob/1998e573d300eae060b7dc5211453f6a2ae44c01/packages/datadog_flags/lib/src/flags_client.dart#L49-L62) carries `value`, `variant`, `reason`, and `error`; the client exposes details-only typed methods so app code can always inspect why a default was returned. |
| Evaluation context | OpenFeature evaluation context has an optional targeting key and custom attributes. Android direct `FlagsClient` and iOS direct `FlagsClient` both require a context step before local evaluation. | [`FlagsEvaluationContext`](https://github.com/DataDog/dd-sdk-flutter/blob/1998e573d300eae060b7dc5211453f6a2ae44c01/packages/datadog_flags/lib/src/evaluation_context.dart#L15-L31) has optional `targetingKey` plus JSON attributes. |
| Direct client as provider substrate | Android documents direct `FlagsClient` as the advanced substrate underneath the OpenFeature provider, and the direct Android interface says it follows OpenFeature closely to simplify provider implementation. | [`DatadogFlags`](https://github.com/DataDog/dd-sdk-flutter/blob/1998e573d300eae060b7dc5211453f6a2ae44c01/packages/datadog_flags/lib/src/datadog_flags.dart#L20-L99) owns enablement and named client construction through `sharedClient`, so a later Dart OpenFeature extension can adapt those clients into provider methods without changing evaluation semantics. |
| Default semantics | OpenFeature basic evaluation returns the caller default on evaluation errors; the provider spec allows errors to be represented through language idioms or error fields on details. Android direct implementation returns defaults while attaching `ErrorCode` in resolution details. | [`DefaultDatadogFlagsClient`](https://github.com/DataDog/dd-sdk-flutter/blob/1998e573d300eae060b7dc5211453f6a2ae44c01/packages/datadog_flags/lib/src/default_flags_client.dart#L121-L182) returns caller defaults and sets `FlagEvaluationError` for provider-not-ready, flag-not-found, and type-mismatch paths. Fetch failures are caught before returning to app code, and misconfiguration produces a no-op client instead of throwing. |
| Multiple mobile subjects | Mobile apps commonly evaluate flags for more than one subject, such as logged-out and logged-in users or Datadog-style org-level and user-level targeting. | This stack supports separate named clients so each client can own one current evaluation context and assignment cache. |
| Isolate-local state | Dart isolates do not share memory, so SDK state and client caches are not shared with background isolates. | `DatadogFlags` and its clients are isolate-local. Background isolates must use their own `DatadogFlags` instance, create the clients they need, and initialize those clients independently. |
| Event delivery lifecycle | OpenFeature evaluation does not require applications to manually flush provider events. Datadog's exposure and flagevaluation delivery should be a runtime concern owned by the SDK. | This stack exposes no `flush` API. Later telemetry stacks should batch, drain, and shut down internally instead of asking app code to manage event delivery. |

References:

- OpenFeature evaluation API: https://openfeature.dev/docs/reference/concepts/evaluation-api/
- OpenFeature provider resolution requirements: https://openfeature.dev/specification/sections/providers/#22-flag-value-resolution
- OpenFeature evaluation context requirements: https://openfeature.dev/specification/sections/evaluation-context/#31-fields
- Datadog JS browser provider typed resolution: https://github.com/DataDog/openfeature-js-client/blob/60b0554f57b44ed75ec4d1fc60394c392b1d0b90/packages/browser/src/openfeature/provider.ts#L266-L305
- Datadog Swift OpenFeature provider typed evaluation: https://github.com/DataDog/dd-openfeature-provider-swift/blob/f8aa96323c66ff2cfb9c41a482498c9bbd456b70/Sources/DatadogOpenFeatureProvider/DatadogProvider.swift#L65-L89
- Datadog Android direct `FlagsClient` API: https://github.com/DataDog/dd-sdk-android/blob/2a08737bc3878f317e11bf6bc855f485b8f29f15/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsClient.kt#L61-L149
- Datadog Android OpenFeature usage docs: https://docs.datadoghq.com/feature_flags/client/android/
- Datadog iOS direct `FlagsClient` usage docs: https://docs.datadoghq.com/feature_flags/client/ios/

## Stack Position

```mermaid
flowchart LR

    class pr1048 current;
```

## Changes

This adds the `DatadogFlags.instance` enable/shared-client facade, `DatadogFlagsClient`, typed details methods, and client-owned assignment state for the latest initialized context. `DatadogFlags` owns client construction; `DatadogFlagsClient` represents an evaluator for one current context. `initialize(context)` fetches assignments through the Stack 1 precompute fetcher; typed details methods then resolve from the current in-memory state.

The public evaluation path returns caller defaults for provider-not-ready, missing flag, and type mismatch outcomes. Fetch failures are caught by the client so application code does not see thrown errors during normal flagging decisions. Missing Datadog configuration leaves the facade in a valid no-op state instead of throwing, so app code can continue to evaluate defaults. Object defaults are returned as provided rather than being JSON-validated during evaluation.

This also keeps the package Dart-only: tests use `package:test`, the analyzer config uses `package:lints`, and the core package does not depend on Flutter or `datadog_flutter_plugin`. The README documents multiple named clients for multiple mobile subjects, and documents that background isolates must use their own `DatadogFlags` instance and create their own clients. A standalone Dart CLI example package exercises the public API without exposing the internal precompute transport as customer API; local validation instructions live in `CONTRIBUTING.md`, not the customer README.

Example behavior check:

```sh
cd packages/datadog_flags/example
DD_CLIENT_TOKEN=fake-token \
dart run datadog_flags_example:typed_evaluation \
  --env staging \
  --flag-key checkout.enabled \
  --flag-type boolean \
  --targeting-key test-subject
```

```text
key: checkout.enabled
value: false
variant: (none)
reason: (none)
error: providerNotReady
```

## Decisions

This PR keeps assignment state in memory only. Each named client owns one current evaluation context and assignment cache. Last-known assignment persistence, context-matched restore, and reset persistence are reserved for the persistence stack.

This PR treats `DatadogFlags` and `DatadogFlagsClient` state as isolate-local. Apps that evaluate flags from background isolates must use their own `DatadogFlags` instance, create the needed clients, and initialize those clients in those isolates; cross-isolate assignment sharing is intentionally out of scope for this stack.

This PR intentionally does not add RUM feature flag tagging. Keeping `datadog_flags` Dart-only avoids coupling the core SDK to Flutter runtime APIs; RUM integration should be handled in a later Flutter integration layer or example-app stack.

This PR also does not emit exposures or flagevaluation metrics. Those side effects are intentionally reserved for the telemetry stacks so reviewers can validate each HTTP boundary independently. Event delivery should be managed internally when batching lands, so this stack does not expose a `flush` method.

The package changelog is intentionally absent in this stack because repository release notes are generated from Conventional Commits.


packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -0,0 +1,132 @@
...
+import 'flags_error.dart';
+import 'flags_repository.dart';
+
+class DatadogFlags {
+  static DatadogFlagsConfiguration _configuration =
+      const DatadogFlagsConfiguration();
+  static DatadogFlagsContext? _datadogContext;
+  static http.Client? _httpClient;
+  static final Map<String, DatadogFlagsClient> _clients = {};
+  static bool _enabled = false;
```
fuzzybinary: I highly recommend we look into a singleton pattern for this instead of having these as static variables.

packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -0,0 +1,132 @@
...
+    _httpClient?.close();
+    _httpClient = null;
+    _datadogContext = null;
+    _enabled = false;
+
+    final datadogContext = configuration.datadogContext;
+    if (datadogContext == null) {
+      throw FlagsException.invalidConfiguration(
+        'DatadogFlagsConfiguration.datadogContext is required.',
+      );
```
fuzzybinary: It is generally my preference that we avoid throwing as much as we can in client SDKs and instead emit a Warning or Error before leaving the SDK in a valid, no-op state. The last thing we want is for a minor misconfiguration in Datadog to end up crashing end user applications.

packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -0,0 +1,228 @@
...
+    return getObjectDetails(key: key, defaultValue: defaultValue).value;
+  }
+
+  @override
+  Future<void> reset() {
+    return _repository.reset();
+  }
+
+  @override
+  Future<void> dispose() async {}
```
fuzzybinary: Is there something this method is meant to do on other clients?

packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -0,0 +1,228 @@
...
+
+  FlagDetails<T> _getDetails<T>({
```
fuzzybinary: Any reason not to make this public?

packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -0,0 +1,228 @@
...
+      value: typedValue.value as T,
+      variant: assignment.variationKey,
+      reason: assignment.reason,
+    );
+  }
+
+  ({bool matched, T? value}) _typedValue<T>(
+    FlagAssignment assignment,
+    FlagVariationType requestedType,
+  ) {
```
fuzzybinary: Doesn't `Respository.flagAssignment` already do this check?

packages/datadog_flags/lib/src/flags_client.dart
```diff
@@ -0,0 +1,84 @@
...
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'datadog_flags.dart';
+import 'flags_context.dart';
+import 'flags_details.dart';
+
+abstract interface class DatadogFlagsClient {
+  static const defaultName = 'default';
```
fuzzybinary: nit: I feel like `defaultName` makes more sense in DatadogFlags, and `DatadogFlagsClient` should just accept whatever its given (and have no default value).

packages/datadog_flags/lib/src/flags_client.dart
```diff
@@ -0,0 +1,84 @@
...
+    String name = defaultName,
+  }) {
+    return DatadogFlags.createClient(name: name);
+  }
+
+  static DatadogFlagsClient shared({
+    String name = defaultName,
+  }) {
+    return DatadogFlags.sharedClient(name: name);
+  }
```
fuzzybinary: I think we should remove these. It creates a circular reference to `DatadogFlags` which should control all construction as it's the main factory.

packages/datadog_flags/lib/src/flags_details.dart
```diff
@@ -0,0 +1,22 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'flags_error.dart';
+
+class FlagDetails<T> {
```
fuzzybinary: This can be held in `flags_client.dart` I think.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,52 @@
...
+    final PrecomputedAssignments assignments;
+    try {
+      assignments = await fetcher.fetch(context);
+    } catch (_) {
+      if (requestId == _contextRequestId) {
+        _context = null;
+        _flags = const {};
+      }
+      return;
+    }
```
fuzzybinary: It might be worth a comment here explaining what's happening and what _contextRequestId / requestId is doing.

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -0,0 +1,307 @@
...
+    final details = client.getBooleanDetails(
+      key: 'show-paywall',
+      defaultValue: false,
+    );
+    expect(details.value, isFalse);
+    expect(details.error, FlagEvaluationError.providerNotReady);
+  });
+}
+
+Future<DatadogFlagsClient> createClient({
```
fuzzybinary: These should either be private or declared in `main` so they don't leak.

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -0,0 +1,307 @@
...
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:datadog_flags/datadog_flags.dart';
+import 'package:http/http.dart' as http;
+import 'package:http/testing.dart';
+import 'package:test/test.dart';
+
+void main() {
+  late List<http.Request> requests;
```
fuzzybinary: Any reason this is global to the test file instead of declared in individual tests?

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -0,0 +1,307 @@
...
+
+  setUp(() async {
+    requests = [];
+    await DatadogFlags.disable();
+  });
+
+  tearDown(() async {
+    await DatadogFlags.disable();
+  });
```
fuzzybinary: If we remove the global statics we can remove these setUp / tearDown methods.

fuzzybinary (changes requested):

packages/datadog_flags/lib/src/default_flags_client.dart
leoromanovsky: Good observation thank you.

packages/datadog_flags/README.md
```diff
@@ -4,13 +4,62 @@
...
+final enabled = flags.getBooleanValue(
+  key: 'checkout.enabled',
+  defaultValue: false,
+);
+```
+
+## Multiple Contexts and Isolates
+
+Use separate clients for separate mobile subjects, such as logged-out and
+logged-in users or org-level and user-level targeting:
```
drichards-87:
```suggestion
Use separate clients for different mobile subjects, such as logged-out and
logged-in users or org-level and user-level targeting:
```

packages/datadog_flags/README.md
```diff
@@ -4,13 +4,62 @@
...
+final userFlags = await datadogFlags.createClient(name: 'user');
+await userFlags.setEvaluationContext(
+  const FlagsEvaluationContext(targetingKey: 'user-456'),
+);
+```
+
+Clients are local to the Dart isolate where they are created. Background
+isolates do not share `DatadogFlags` state or client assignment caches with the
+main isolate, so they must call `DatadogFlags.instance.enable()`, create the
+clients they need, and set evaluation contexts independently.
```
drichards-87:
```suggestion
main isolate, so each background isolate must call `DatadogFlags.instance.enable()`, create the clients it needs, and set evaluation contexts independently.
```

packages/datadog_flags/README.md
```diff
@@ -4,13 +4,62 @@
...
+
+## Behavior
+
+- Unknown or malformed individual flag assignments are ignored so one bad flag
+  does not prevent other assignments from loading.
```
drichards-87:
```suggestion
- Unknown or malformed individual flag assignments are ignored so that one invalid assignment does not prevent other assignments from loading.
```

packages/datadog_flags/README.md
```diff
@@ -4,13 +4,62 @@
...
+- Typed evaluations return caller-provided defaults instead of throwing when
+  assignments are unavailable, a flag is missing, or a flag has the wrong type.
+- Typed details include provider-not-ready, flag-not-found, or type-mismatch
+  errors when defaults are used.
```
drichards-87:
```suggestion
  errors when default values are used.
```

drichards-87 (approved):
Left some suggestions from Docs and approved the PR.

packages/datadog_flags/example/bin/typed_evaluation.dart
fuzzybinary, line 1: This is okay for this PR, but in reality `example` should be a full application with its own pubspec and everything. I should have noticed this before but it slipped through.

It might also make sense to make this more of an actual command line application that takes the key, type, targeting key and env as parameters, and only looks at the env variables for the token and application id. It would be a more full featured example.

packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -0,0 +1,136 @@
...
+    String name = defaultClientName,
+  }) async {
+    return _client(name);
+  }
+
+  DatadogFlagsClient sharedClient({
+    String name = defaultClientName,
+  }) {
+    return _client(name);
+  }
```
fuzzybinary: If `sharedClient` and `createClient` are going to do the same thing I think it's okay to just have `sharedClient`.

packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -0,0 +1,136 @@
...
+    final httpClient = _httpClient;
+    if (!_enabled || datadogContext == null || httpClient == null) {
+      return null;
+    }
+
+    return _FlagsRuntime(
+      configuration: _configuration,
+      datadogContext: datadogContext,
+      httpClient: httpClient,
+    );
```
fuzzybinary: Does it makes sense to make this an actual member variable that's created on `enable`? If it's `null`, enable hasn't been called or has failed, and we can get rid of the _enabled check.

packages/datadog_flags/lib/src/datadog_flags.dart
fuzzybinary: Also `createClient` no longer needs to be async. That strikes me as something in the implementation drastically changed here. Was that intentional?

fuzzybinary (commented):
Very very minor places I think we can clean up a bit. Otherwise looking good.

packages/datadog_flags/example/bin/typed_evaluation.dart
leoromanovsky: Done in `43fa3201`: `example` is now a standalone Dart package with its own `pubspec.yaml`, executable entrypoint, CLI parameters for env/site/key/type/targeting context, and env vars only for `DD_CLIENT_TOKEN` / `DD_APPLICATION_ID`.

packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -0,0 +1,115 @@
...
+import 'flags_configuration.dart';
+import 'flags_repository.dart';
+import 'no_op_flags_client.dart';
+
+/// Entry point for configuring and creating Datadog feature flag clients.
+///
+/// `DatadogFlags` state is local to the current Dart isolate. Background
+/// isolates must use their own [DatadogFlags] instance, create any clients they
+/// need, and set their own evaluation contexts before evaluating flags.
+class DatadogFlags {
```
dd-oleksii, line 20: minor: imo it's better to get rid of singletons and remove all of this

packages/datadog_flags/lib/src/flags_client.dart
```diff
@@ -0,0 +1,88 @@
...
+
+import 'flags_context.dart';
+import 'flags_error.dart';
+
+/// Evaluates feature flags for one current evaluation context.
+///
+/// Create separate clients for separate mobile subjects, such as logged-out and
+/// logged-in users. Clients are local to the Dart isolate where they are
+/// created and must be recreated in background isolates.
+abstract interface class DatadogFlagsClient {
```
dd-oleksii, line 14: minor: consider modeling this after (expected) openfeature Provider interfame:
- remove non-details methods
- add initialize/shutdown functions

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,55 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'assignment.dart';
+import 'flag_assignments_fetcher.dart';
+import 'flags_context.dart';
+
+class FlagsRepository {
```
dd-oleksii: minor: consider inlining it into "default flags client"

packages/datadog_flags/lib/src/no_op_flags_client.dart
```diff
@@ -0,0 +1,112 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'flags_client.dart';
+import 'flags_context.dart';
+import 'flags_error.dart';
+
+class NoOpDatadogFlagsClient implements DatadogFlagsClient {
```
dd-oleksii, line 10: minor: consider removing this along with the singleton

dd-oleksii (approved):
Looks good functionality-wise, though I suggest dropping singleton + align client interface with what we expect openfeature provider interface to be

PR #1049, opened 2026-06-03
feat(datadog_flags): add exposure logging

leoromanovsky:
## Motivation

After precompute fetching and typed local evaluation, successful flag evaluations need to produce exposure telemetry. This stack isolates exposure logging from flagevaluation aggregation so reviewers can validate one side effect: a successful `FlagDetails` evaluation should immediately attempt an exposure event when the assignment says `doLog: true`.

The customer problem is observability correctness. If a flag is part of an experiment allocation, evaluating that flag without an exposure means experiment analysis can undercount participants. At the same time, default paths, type mismatches, and missing flags must not create false exposures.

## System Scope

This PR implements the highlighted exposure path. It keeps the package Dart-only and leaves RUM feature-flag tagging, flagevaluation batching, persistence, and Flutter example UI to later stacks.

```mermaid
flowchart LR
    app["Dart client app"] --> package["datadog_flags package"]
    package --> facade["DatadogFlags facade"]
    facade --> client["DatadogFlagsClient"]
    client --> init["initialize(context)"]
    init --> fetcher["Precompute assignments fetcher"]
    fetcher --> precompute["Datadog precompute API"]
    client --> details["get*Details(...)"]
    details --> success["Successful assignment resolution"]
    success --> exposure["ExposureLogger"]
    exposure --> intake["EVP exposure intake"]
    exposure --> cache["In-memory exposure dedupe"]

    flageval["Flagevaluation aggregation"] -. later .-> package
    persistence["Last-known assignment persistence"] -. later .-> package
    rum["Flutter/RUM feature flag integration"] -. later .-> package
    example["Flutter dogfooding app"] -. later .-> package

    class exposure,intake,cache,success active;
```

## Stack Position

```mermaid
flowchart LR

    class pr1049 current;
```

## Changes and Decisions

This adds exposure intake endpoint mapping to `DatadogFlagsSite`, exposure-specific configuration knobs on `DatadogFlagsConfiguration`, and an internal `ExposureLogger`. `DefaultDatadogFlagsClient` now calls the logger only after a typed details method resolves a real assignment with the requested type. Provider-not-ready, flag-not-found, and type-mismatch paths still return the caller default and do not emit exposures.

The exposure request follows the mobile SDK shape: `POST /api/v2/exposures`, `Content-Type: text/plain;charset=UTF-8`, `DD-API-KEY`, `DD-EVP-ORIGIN`, `DD-EVP-ORIGIN-VERSION`, `DD-REQUEST-ID`, and a JSON exposure event with timestamp, allocation, flag, variant, and subject. The request includes `ddsource` as a query parameter like the iOS request builder. Failed exposure sends remove the dedupe key so a later evaluation can retry.

The dedupe key is `targetingKey + flagKey + allocationKey + variationKey`, matching the practical mobile behavior: repeated reads of the same assigned variant should not inflate exposure counts, but a different subject or assignment should be eligible to emit again.

This PR deliberately removes the old Flutter/RUM coupling that was present on the stale exposure branch. `datadog_flags` remains a plain Dart package with no `datadog_flutter_plugin` dependency. RUM feature-flag tagging belongs in a later Flutter integration layer; flagevaluation aggregation belongs in the next stack.

Relevant code:

- Exposure transport: https://github.com/DataDog/dd-sdk-flutter/blob/56e33b87a46fccad72babf67e83c1fbe46292178/packages/datadog_flags/lib/src/exposure_logger.dart
- Evaluation hook: https://github.com/DataDog/dd-sdk-flutter/blob/56e33b87a46fccad72babf67e83c1fbe46292178/packages/datadog_flags/lib/src/default_flags_client.dart
- Configuration: https://github.com/DataDog/dd-sdk-flutter/blob/56e33b87a46fccad72babf67e83c1fbe46292178/packages/datadog_flags/lib/src/flags_configuration.dart
- HTTP boundary tests: https://github.com/DataDog/dd-sdk-flutter/blob/56e33b87a46fccad72babf67e83c1fbe46292178/packages/datadog_flags/test/datadog_flags_test.dart

Mobile precedent:

- Android exposure request factory: https://github.com/DataDog/dd-sdk-android/blob/8b687f09c90ee2d0a91a8783e52fbb8e919d2739/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/net/ExposuresRequestFactory.kt
- iOS exposure logger: https://github.com/DataDog/dd-sdk-ios/blob/6da56aac76a733b024c1374c7f468e19d663d119/DatadogFlags/Sources/Client/ExposureLogger.swift
- iOS exposure request builder: https://github.com/DataDog/dd-sdk-ios/blob/6da56aac76a733b024c1374c7f468e19d663d119/DatadogFlags/Sources/Feature/ExposureRequestBuilder.swift

Validation evidence:

- `dart analyze packages/datadog_flags`
- `dart test` in `packages/datadog_flags` passed with exposure HTTP-boundary coverage
- `dart test --platform chrome` in `packages/datadog_flags` passed
- `flutter test test` in `packages/datadog_flags` passed
- `dart analyze .` in `packages/datadog_flags/example` passed
- `dart format --set-exit-if-changed packages/datadog_flags` passed
- `git diff --check && git diff --cached --check` passed

The exposure tests assert: no exposure before provider readiness, exactly one exposure for repeated successful evaluations of the same assignment, no exposure when tracking is disabled, no exposure when `doLog` is false, no exposure for missing/type-mismatched flags, and retry after an initial failed exposure send.


rtrieu (dismissed):

packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -93,6 +92,11 @@ class DatadogFlags {
     final client = DefaultDatadogFlagsClient(
       name: name,
       fetcher: fetcher,
+      exposureLogger: ExposureLogger(
```
fuzzybinary: Why not just pass the `Runtime`?

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -0,0 +1,94 @@
...
+          'DD-REQUEST-ID': const Uuid().v4(),
+        },
+        body: jsonEncode(event),
+      );
+      if (response.statusCode < 200 || response.statusCode >= 300) {
+        _loggedExposures.remove(exposureKey);
+      }
+    } catch (_) {
+      _loggedExposures.remove(exposureKey);
+    }
```
fuzzybinary, line 198: One thing I'm missing from the flags SDK at the moment is logging to the user when something goes wrong. It might be worth looking into if we can have an logging interface similar to what exists in the main package so that users can set a "verbose" diagnostic level and get information about success / failure. That helps with both their troubleshooting and ours when a support request comes in.

fuzzybinary (dismissed):
Looks good. Minor nit where I don't think we need to split out the `Runtime` parameters being sent to the exposure logger.

packages/datadog_flags/lib/src/datadog_flags.dart
leoromanovsky: Updated in `cb2a45fb` to pass a shared internal `FlagsRuntime` into `ExposureLogger` instead of splitting the runtime fields at the call site.

packages/datadog_flags/lib/src/exposure_logger.dart
leoromanovsky: Agreed. I’m keeping this PR focused on exposure emission and will add a diagnostic logging interface in its own follow-up PR stack so fetch, evaluation, and exposure failures can share the same shape.

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -0,0 +1,88 @@
...
+import 'package:uuid/uuid.dart';
+
+import 'assignment.dart';
+import 'evaluation_context.dart';
+import 'flags_runtime.dart';
+import 'json_value.dart';
+
+class ExposureLogger {
+  final FlagsRuntime runtime;
+  final Set<String> _loggedExposures = {};
```
dd-oleksii: major: is there a specific reason it deviates from our usual exposure cache?

packages/datadog_flags/lib/src/datadog_flags_config.dart
```diff
@@ -6,39 +6,66 @@
...
 
   Uri get flagsEndpoint => Uri.parse(flagsEndpointUrl);
+  Uri get intakeEndpoint => Uri.parse(intakeEndpointUrl);
 }
 
 @immutable
 final class DatadogFlagsConfig {
   static const defaultSdkName = 'dd-sdk-flutter';
   static const defaultSdkVersion = '0.0.1';
+  static const defaultSource = 'flutter';
```
dd-oleksii: minor: I don't think source and sdk name/version should be configurable

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -0,0 +1,88 @@
...
+        'attributes': sanitizeJsonValue(evaluationContext.attributes),
+      });
+      final event = {
+        'timestamp': configuration.dateProvider().millisecondsSinceEpoch,
+        'allocation': {'key': assignment.allocationKey},
+        'flag': {'key': flagKey},
+        'variant': {'key': assignment.variationKey},
+        'subject': subject,
+      };
+      final response = await runtime.httpClient.post(
```
dd-oleksii: major: is it supposed to send every exposure event individually?

packages/datadog_flags/lib/src/flags_runtime.dart
```diff
@@ -0,0 +1,21 @@
...
+
+import 'package:http/http.dart' as http;
+
+import 'datadog_flags_config.dart';
+import 'flags_configuration.dart';
+
+class FlagsRuntime {
+  final DatadogFlagsConfiguration configuration;
+  final DatadogFlagsConfig datadogConfig;
+  final http.Client httpClient;
```
dd-oleksii, line 14: nitpick: "runtime" is a weird name for a bunch of configurations and an http client. From the use, it seems like it wants to be an http client for datadog apis (and configs select endpoints)

packages/datadog_flags/lib/src/exposure_logger.dart
leoromanovsky: Thanks for noticing this. Sorry, I was too casual modeling the behavior on ios/android... just found bugs in both of them (fixing).

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -283,11 +283,11 @@ void main() {
       final request = _exposureRequests(requests).single;
       expect(
         request.url.toString(),
-        'https://browser-intake-datadoghq.com/api/v2/exposures?ddsource=flutter',
+        'https://browser-intake-datadoghq.com/api/v2/exposures?ddsource=dart-client',
```
fuzzybinary, line 287: Be careful here -- this impacts billing and depending on a number of factors may need some backend code to work properly.

packages/datadog_flags/lib/src/version.dart
```diff
@@ -0,0 +1,6 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+const ddPackageVersion = '0.0.1';
```
leoromanovsky, line 6: @fuzzybinary Do you have some suggestions for emitting the currently published version from the pubspec.yml? I wasn't able to solve this well in eppo either (https://github.com/Eppo-exp/dart-sdk/blob/main/lib/src/sdk_version.dart) with Flutter SDKs.

packages/datadog_flags/test/datadog_flags_test.dart
leoromanovsky: Thanks. Anything specific to know about? We are ingesting exposures via EVP across the other 8 released SDKs in a similar pattern and volume. A local cache reduces the number of events to just the canonical ones needed for experiment analysis.

Or do you mean adding an abstraction in front of it?

packages/datadog_flags/lib/src/flags_runtime.dart
leoromanovsky: Suggestions are welcome.

packages/datadog_flags/lib/src/version.dart
fuzzybinary: No, unfortunately. Once we can support Dart 3.10+, we can potentially automatically update a file with build hooks (like [this plugin](https://pub.dev/packages/pubspec_version_hook)) but for now we have to do it manually.

In the main package we update a `version.dart` file on release, and if actually if you put the right file in the right location with the right name, you can have the releaser take care of it for you: https://github.com/DataDog/dd-sdk-flutter/blob/e8f7928808c3cb66347b777d95882f63892d05f7/tools/releaser/lib/version_updater.dart#L120

leoromanovsky: Thank you!

I noticed the releaser already looks for exactly `packages/[package]/lib/src/version.dart` and rewrites lines starting with `const ddPackageVersion`. I believe our file matches that contract.

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -0,0 +1,205 @@
...
+    return endpoint.replace(
+      queryParameters: {
+        ...endpoint.queryParameters,
+        'ddsource': datadogFlagsSource,
+      },
+    );
+  }
+}
+
+final class _ExposureCacheKey {
```
fuzzybinary, line 228: Tag these with structs with `@immutable`.

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -0,0 +1,205 @@
...
+
+  Future<void> flush({bool rescheduleOnFailure = false}) async {
+    _flushTimer?.cancel();
+    _flushTimer = null;
+
+    final inFlight = _flushInFlight;
+    if (inFlight != null) {
+      await inFlight;
+      return flush(rescheduleOnFailure: rescheduleOnFailure);
+    }
```
fuzzybinary: This feels more complicated and potentially more fragile than it needs to be. I worry about what might happen if several flush's get backed up if, for example, talking to the Datadog exposure endpoint is repeatedly timing out.

Instead, in a case where a flush is requested and one is pending (which can be a bool), use the _flushTimer with a smaller delay (say 500ms instead of 10 seconds). The timer checks on a regular basis, but there's only ever one `flush` request in flight at a time. Once it succeeds in a full flush, it goes back to waiting for the next exposure with a 10 second delay.

fuzzybinary: 🤔 I guess the point is to try to get the `shutdown` flush to clear all pending flushes and exposures.... I may have to think about this some more, because I still a bit worried about multiple flushes backing up.

fuzzybinary (commented):
Minor nit about `@immutable`.

For the flushing - the approach feels fragile but I'll have to think about if there's a better way to await all pending flushes, then perform a final flush in order to make "shutdown" make sense.

packages/datadog_flags/lib/src/exposure_logger.dart
leoromanovsky: Updated with tests.

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -129,15 +193,21 @@ class ExposureLogger {
...
 
-  void _scheduleFlush() {
-    _flushTimer ??= Timer(_exposureFlushDelay, () {
+  void _scheduleFlush({Duration delay = _exposureFlushDelay}) {
+    if (_shutdownDrainActive || _flushTimer != null) {
+      return;
+    }
+
+    final flushDelay =
+        _uploadInFlight == null ? delay : _exposureFlushRetryDelay;
```
fuzzybinary, line 207: If the delay is determined here and it ignores the parameter in certain circumstances, maybe we don't need the parameter?

packages/datadog_flags/test/datadog_flags_test.dart
fuzzybinary: To be honest I'm not 100% sure how this is for intake other than RUM. If you can add a source to the exposure endpoint without requiring too much work than you're probably fine.

fuzzybinary (approved):
Yeah this way of doing flushing is way better and avoids a bunch of flushes in flight at the same time.

One minor nit on the the delay parameter but otherwise this LGTM.

packages/datadog_flags/test/datadog_flags_test.dart
leoromanovsky: Revalidated against ios/android - yes not positive the `ddsource` will persist but it's optional.

packages/datadog_flags/lib/src/exposure_logger.dart
leoromanovsky: Will fix in https://github.com/DataDog/dd-sdk-flutter/pull/1050

PR #1050, opened 2026-06-03
feat(datadog_flags): add flagevaluation metrics

leoromanovsky:
## Motivation

Exposure logging captures assignment exposure, but Flutter also needs SDK-side evaluation metrics emitted by the other Datadog flagging clients. This stack adds flagevaluation aggregation after the typed API and exposure layers are in place, so typed evaluations immediately feed telemetry without adding batching or transport knobs to the OpenFeature-shaped client API.

## System Scope

This PR adds SDK-side flagevaluation aggregation, timer/max-batch flush behavior, failed-flush restore, and success/error evaluation payloads.

```mermaid
flowchart LR
    app[Flutter app] --> package[datadog_flags package]
    package --> facade[DatadogFlags facade]
    facade --> client[Named flags client]
    client --> typed[Typed evaluation API]
    client --> context[Evaluation context]
    context --> precompute[Precompute assignments fetcher]
    precompute --> api[Datadog precompute API]
    typed --> exposure[Exposure EVP intake]
    typed --> flageval[Flag-evaluation aggregator]
    flageval --> intake[Flag-evaluation EVP intake]

    class flageval,intake active;
```

## Stack Position

```mermaid
flowchart LR
    pr1047[#1047<br/>Precompute core] --> pr1048[#1048<br/>Typed API]
    pr1048 --> pr1049[#1049<br/>Exposures]
    pr1049 --> pr1050[#1050<br/>Flagevals]
    pr1050 --> pr1051[#1051<br/>Persistence]
    pr1051 --> pr1052[#1052<br/>Publishing]
    pr1052 --> pr1053[#1053<br/>Example app]

    class pr1050 current;
```

## Changes

This adds `EvaluationAggregator`, flagevaluation intake configuration, batch context generation, and client wiring. Successful, defaulted, and error evaluations are aggregated by flag, assignment, targeting key, evaluation context, and error state. Successful typed evaluations still log exposures first, then record the evaluation aggregate from the same synchronous evaluation call.

The flagevaluation transport follows the iOS/browser-style JSON intake envelope used for this Dart package. The event context currently includes Datadog `env` and optional RUM application id because this package remains Dart-only and does not depend on Flutter runtime/device metadata.

## Native References

The batching and lifecycle choices intentionally follow the mobile SDKs:

- iOS creates an internal `EvaluationAggregator`, clamps `evaluationFlushInterval` to 1s-60s, and keeps `maxAggregations` internal: [FlagsFeature.swift](https://github.com/DataDog/dd-sdk-ios/blob/main/DatadogFlags/Sources/Feature/FlagsFeature.swift), [EvaluationAggregator.swift](https://github.com/DataDog/dd-sdk-ios/blob/main/DatadogFlags/Sources/Client/EvaluationAggregator.swift).
- Android coerces `evaluationFlushInterval` to 1s-60s and creates the max aggregation limit inside `EvaluationsFeature`: [FlagsConfiguration.kt](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/FlagsConfiguration.kt), [EvaluationsFeature.kt](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationsFeature.kt).
- Android shutdown drains pending events and bounds the executor wait, which this PR mirrors by adding a timeout to flagevaluation uploads: [EvaluationEventsProcessor.kt](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationEventsProcessor.kt).

## Decisions

This PR keeps persistence out of scope. It records only in-memory evaluation aggregates and restores failed flushes inside the aggregator buffer.

`evaluationFlushInterval` remains public but is coerced to the native SDK 1s-60s bounds. The max aggregation count is now internal, matching iOS and Android, with a test-only hook for deterministic batch-flush coverage. Public callers keep the OpenFeature-shaped surface: initialize a client context, synchronously evaluate typed details with defaults, and call `shutdown()` for lifecycle cleanup.

Tests now cover both the client HTTP boundary and `EvaluationAggregator` directly. The dedicated aggregator tests cover aggregation-key behavior, runtime-default/error separation, failed-upload restore, internal max-batch auto flush, disabled tracking, and bounded shutdown.

Local validation: `dart analyze .`, `dart test`, `dart test --platform chrome`, `dart format --set-exit-if-changed .`, and `git diff --check` all pass in `packages/datadog_flags`.


packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -128,6 +135,12 @@ class DefaultDatadogFlagsClient implements DatadogFlagsClient {
   }) {
     final context = _context;
     if (context == null) {
+      _evaluationAggregator.recordEvaluation(
+        flagKey: key,
+        assignment: null,
+        evaluationContext: FlagsEvaluationContext.empty,
+        error: FlagEvaluationError.providerNotReady.name,
```
sameerank: I assume these are in camelCase 

https://github.com/DataDog/dd-sdk-flutter/blob/70c54c2b8875d29e2c9264169be025e8a21521d7/packages/datadog_flags/lib/src/flags_error.dart#L6

[But in iOS these are SCREAMING_SNAKE_CASE](https://github.com/DataDog/dd-sdk-ios/blob/608d7b7c0b700f2bb48304764502b95e431a92c6/DatadogFlags/Sources/Client/FlagsClient.swift#L44-L48)

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -637,20 +672,299 @@ void main() {
...
+      expect(success['targeting_key'], 'user-123');
+      expect(success['context'], {
+        'evaluation': {'plan': 'pro'},
+      });
+      expect(success.containsKey('runtime_default_used'), isFalse);
+
+      final mismatch = _evaluation(
+        evaluations,
+        flagKey: 'show-paywall',
+        error: FlagEvaluationError.typeMismatch.name,
```
sameerank: Also related to error message casing, and these tests reinforce the divergent behavior

packages/datadog_flags/lib/src/evaluation_aggregator.dart
```diff
@@ -0,0 +1,361 @@
...
+
+  Map<String, Object?> _datadogContext() {
+    final applicationId = runtime.datadogConfig.applicationId;
+    return _removeNullValues({
+      'env': runtime.datadogConfig.env,
+      'rum': applicationId == null
+          ? null
+          : {
+              'application': {'id': applicationId},
+              'view': null,
```
sameerank: In iOS, it *looks* like [this field, `view`, is included and set to `nil`](https://github.com/DataDog/dd-sdk-ios/blob/4e6b50831ca347e28c4ebf1237dd708afb00cdde/DatadogFlags/Sources/Feature/EvaluationRequestBuilder.swift#L84-L87), but actually the purpose of it was to indicate what is stated in the comment "viewURL not available in RUMCoreContext"

And since [RUMInfo is a Codable](https://github.com/DataDog/dd-sdk-ios/blob/e3581cd4e84a14fbf2bdee74e36f87c4a2caf46c/DatadogFlags/Sources/Models/FlagEvaluationEvent.swift#L51), `view` gets removed by `JSONEncoder` which [skips nils](https://developer.apple.com/documentation/swift/keyedencodingcontainer/encodeifpresent(_:forkey:)-45la3). 

In Android, on the other hand, [it is actually possible to get the view from the context](https://github.com/DataDog/dd-sdk-android/blob/288b50a5a5d6766601e70202a4e828207603b376/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/EvaluationsFeature.kt#L117), so it is set to a non-null value

So basically two things to consider here in relation to the other client SDKs:
1. `_removeNullValues` only strips this object at the top level, so `'view': null` is included in the generated context, unlike in iOS where it is skipped.
2. The reason to set this to `null` is because you cannot access view name from any of the runtime APIs, unlike Android where it is available.

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -637,20 +672,299 @@ void main() {
...
+    final evaluations =
+        (body['flagEvaluations'] as List<Object?>).cast<Map<String, Object?>>();
+
+    final evaluation = evaluations.single;
+    expect(evaluation['flag'], {'key': 'show-paywall'});
+    expect(evaluation['runtime_default_used'], isTrue);
+    expect(evaluation['error'], {
+      'message': FlagEvaluationError.providerNotReady.name,
+    });
+    expect(evaluation.containsKey('targeting_key'), isFalse);
```
sameerank, line 784: This is related to my comment about `.empty`. This assertion differs from the behavior in iOS

packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -128,6 +135,12 @@ class DefaultDatadogFlagsClient implements DatadogFlagsClient {
   }) {
     final context = _context;
     if (context == null) {
+      _evaluationAggregator.recordEvaluation(
+        flagKey: key,
+        assignment: null,
+        evaluationContext: FlagsEvaluationContext.empty,
```
sameerank, line 141: [iOS sets `targetingKey` to "" when empty](https://github.com/DataDog/dd-sdk-ios/blob/b94f1f2c91aac73521536be3bb5b742e1fdd8fbe/DatadogFlags/Sources/Models/FlagsEvaluationContext.swift#L32). [Android does the same](https://github.com/DataDog/dd-sdk-android/blob/f12ec28661eec4100b9d75749216ba287f4fe4af/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/model/EvaluationContext.kt#L54). This `.empty` doesn't set anything

https://github.com/DataDog/dd-sdk-flutter/blob/70c54c2b8875d29e2c9264169be025e8a21521d7/packages/datadog_flags/lib/src/evaluation_context.dart#L16

And later on the non-inclusion of `targetingKey` is asserted, which doesn't match iOS nor Android

https://github.com/DataDog/dd-sdk-flutter/blob/5b1448c4535d5a8b9c3f37cd82834ab9aea2d271/packages/datadog_flags/test/datadog_flags_test.dart#L784

packages/datadog_flags/lib/src/evaluation_aggregator.dart
```diff
@@ -0,0 +1,361 @@
...
+        datadogConfig.intakeEndpoint().replace(path: '/api/v2/flagevaluation');
+    return endpoint.replace(
+      queryParameters: {
+        ...endpoint.queryParameters,
+        'ddsource': datadogFlagsSource,
+      },
+    );
+  }
+
+  Map<String, Object?> _datadogContext() {
```
sameerank, line 249: For comparison, [here's the analogous context in iOS](https://github.com/DataDog/dd-sdk-ios/blob/e3581cd4e84a14fbf2bdee74e36f87c4a2caf46c/DatadogFlags/Sources/Models/FlagEvaluationEvent.swift#L20).

```
internal struct EvaluationContext: Codable {
    let geo: GeoInfo?
    let device: DeviceInfo
    let os: OSInfo
    let service: String
    let version: String
    let env: String
    let rum: RUMInfo?
```

In the PR description, you've justified why fields like `geo`, `device`, and  `os` are missing

> The event context currently includes Datadog env and optional RUM application id because this package remains Dart-only and does not depend on Flutter runtime/device metadata.

But I am surprised that `service` and/or `version` aren't available since neither is device metadata

packages/datadog_flags/lib/src/default_flags_client.dart
leoromanovsky: Thank you!

sameerank (approved):
Thanks for addressing my feedback!

packages/datadog_flags/lib/src/evaluation_aggregator.dart
```diff
@@ -0,0 +1,382 @@
...
+  bool _shutdownDrainActive = false;
+
+  EvaluationAggregator(this.runtime) {
+    if (runtime.configuration.trackEvaluations) {
+      _periodicFlushTimer = Timer.periodic(
+        runtime.configuration.evaluationFlushInterval,
+        (_) => unawaited(flush()),
+      );
+    }
+  }
```
fuzzybinary, line 43: Would it make more sense to just not create an aggregator if `trackEvaluations` is false? It adds a lot of `?.` to the client, but then makes this class way cleaner as it doesn't have to worry about the state of `trackEvaluations`.

packages/datadog_flags/lib/src/evaluation_aggregator.dart
```diff
@@ -0,0 +1,382 @@
...
+
+  @visibleForTesting
+  static Duration uploadTimeout = defaultUploadTimeout;
+
+  final FlagsRuntime runtime;
+  final Map<String, _AggregatedEvaluation> _aggregations = {};
+  Future<bool>? _uploadInFlight;
+  Future<void>? _shutdownInFlight;
+  Timer? _periodicFlushTimer;
+  Timer? _retryFlushTimer;
```
fuzzybinary, line 33: Something to think about - instead of using a timer for the retry, you could potentially have one upload future that is attempting to perform the Flush. If it fails (and `rescheduleOnFailure` is true) it awaits `Future.delayed` for the specified amount of time and then loops back to the top. You can even have a set number of retries easily this way.

This prevents the complication of having two timers that might overlap. The _periodicFlushTimer sees that an upload is still in flight (that could have tried multiple times) and early outs.

fuzzybinary (commented):
A bit late for my review here, as it's already merged, but some thoughts for improvement.

PR #1051, opened 2026-06-03
feat(datadog_flags): add assignment persistence

leoromanovsky:
## Motivation

The Dart flags client should retain the last successful assignment set so apps can continue from known-good flag state when startup or network conditions are degraded. This should match the existing mobile SDK behavior: stored assignments are allowed only for the same evaluation context, and live precompute remains the source of truth when it succeeds.

## Reference Implementations

- iOS reads/writes `FlagsData` through the SDK data store: [FlagsDataStore.swift](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogFlags/Sources/Client/FlagsDataStore.swift#L16-L46)
- iOS repository reads stored state on startup, waits briefly for disk reads, writes live success, and keeps stale state only for matching context: [FlagsRepository.swift](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogFlags/Sources/Client/FlagsRepository.swift#L78-L139), [failure handling](https://github.com/DataDog/dd-sdk-ios/blob/develop/DatadogFlags/Sources/Client/FlagsRepository.swift#L161-L216)
- Android repository loads and writes persisted state through `DataStoreHandler`: [DefaultFlagsRepository.kt](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/repository/DefaultFlagsRepository.kt#L33-L70)
- Android evaluation flow always fetches precompute and transitions to stale only when cached context matches: [EvaluationsManager.kt](https://github.com/DataDog/dd-sdk-android/blob/develop/features/dd-sdk-android-flags/src/main/kotlin/com/datadog/android/flags/internal/evaluation/EvaluationsManager.kt#L61-L105)

## System Scope

```mermaid
flowchart LR
    app["Client app"] --> facade["DatadogFlags"]
    facade --> client["DefaultDatadogFlagsClient"]
    facade --> noop["NoOpDatadogFlagsClient"]
    facade --> config["DatadogFlagsConfiguration"]
    facade --> ddconfig["DatadogFlagsConfig"]
    facade --> http["HTTP client"]

    client --> repository["FlagsRepository"]
    client --> exposure["ExposureLogger"]
    client --> evaluations["EvaluationAggregator"]

    repository --> fetcher["FlagAssignmentsFetcher"]
    repository --> storeInterface["DatadogFlagsStore"]
    storeInterface --> storeData["FlagsData"]
    fetcher --> precompute["Precompute assignments API"]

    exposure --> exposureIntake["EVP exposures intake"]
    evaluations --> evalIntake["EVP flagevaluation intake"]

    classDef public fill:#e7f6ea,stroke:#18794e,stroke-width:2px,color:#111;
    classDef internal fill:#eef5ff,stroke:#1d4ed8,stroke-width:2px,color:#111;
    class facade,client,config,ddconfig,storeInterface,storeData public;
    class repository,fetcher,exposure,evaluations,noop internal;
```

## Changes

This PR keeps persistence deliberately small. `DatadogFlagsStore` is the public store contract and it now operates on typed `FlagsData` rather than raw JSON maps. `FlagsData` remains responsible for JSON conversion so a future disk-backed store can serialize explicitly without leaking repository internals into the store interface.

`FlagsRepository` now owns assignment state, store reads and writes, and stale-initialize protection. Startup first attempts a bounded store read, publishes cached assignments only when the stored context matches the requested context, and then always requests live precompute. Live precompute remains authoritative: successful live assignments replace stored assignments and are written back to the store. If live precompute fails, stored assignments are used only when they were already proven to match the requested context; otherwise evaluations continue returning programmatic defaults.

The initialize flow is guarded by a small cancel token. Starting a new initialize or clearing memory cancels the previous token, so stale async reads or precompute responses cannot publish assignments after they are obsolete. This preserves the existing behavior where two in-flight precompute requests can resolve out of order without the older response clobbering the newer state. `DatadogFlags.reset()` now calls the polymorphic client `reset()` method, which clears memory and stored assignments without shutting down evaluation or exposure plumbing.

```mermaid
flowchart TD
    init["initialize(context)"] --> token["Create cancel token"]
    token --> hasStore{"Store configured?"}
    hasStore -- "yes" --> readStore["Read stored assignments with bounded timeout"]
    hasStore -- "no" --> skipStore["Skip stored assignment read"]
    readStore --> currentCheck["Continue only if token is not canceled"]
    skipStore --> currentCheck
    currentCheck --> matchStored{"Stored context matches requested context?"}
    matchStored -- "yes" --> publishStored["Publish stored assignments"]
    matchStored -- "no" --> defaults["Keep provider-not-ready defaults"]
    publishStored --> requestLive["Request live precompute"]
    defaults --> requestLive
    requestLive --> liveSuccess{"Live precompute success?"}
    liveSuccess -- "yes" --> publishLive["Publish live assignments"]
    publishLive --> writeStore["Write latest FlagsData to store"]
    writeStore --> latestValues["Evaluate latest values"]
    liveSuccess -- "no" --> storedPublished{"Stored assignments already published?"}
    storedPublished -- "yes" --> staleValues["Evaluate stored values"]
    storedPublished -- "no" --> programmaticDefaults["Return programmatic defaults"]

    classDef live fill:#e7f6ea,stroke:#18794e,stroke-width:2px,color:#111;
    classDef fallback fill:#fff4cc,stroke:#c37b00,stroke-width:2px,color:#111;
    classDef guard fill:#f3f4f6,stroke:#6b7280,stroke-width:2px,color:#111;
    class requestLive,publishLive,writeStore,latestValues live;
    class readStore,publishStored,staleValues fallback;
    class token,currentCheck,storedPublished,programmaticDefaults guard;
```

## Decisions

- Core package stays Dart-only: no Flutter dependency, no `shared_preferences`, no platform channels.
- This PR does not add a concrete disk-backed store. The follow-up can implement one against `DatadogFlagsStore` without changing repository semantics.
- Store behavior is last-known assignment storage, not a read-through cache.
- Context match is mandatory before stored assignments can affect evaluation.
- Stored assignments are startup fallback: they do not replace current in-memory assignments for the same context.
- `shutdown()` clears memory only; `DatadogFlags.reset()` clears memory and stored assignments.

## Stack Position

```mermaid
flowchart LR

    class pr1051 current;
```

## Validation

- `dart analyze .`
- `dart test`
- `dart test --platform chrome`
- `dart format --set-exit-if-changed .`
- `git diff --check`

The tests cover out-of-order live precompute responses, cancel-token reset behavior, matching-context stored startup fallback, matching stored data not replacing current in-memory assignments, mismatched-context cache rejection, reset clearing stored assignments, pending store writes racing reset, and browser-safe compilation.


packages/datadog_flags/lib/src/datadog_flags.dart
```diff
@@ -61,15 +58,23 @@ class DatadogFlags {
...
   Future<void> reset() async {
-    await Future.wait(_clients.values.map((client) => client.shutdown()));
+    await Future.wait(
+      _clients.values.map((client) async {
+        if (client is DefaultDatadogFlagsClient) {
+          await client.reset();
+        } else {
+          await client.shutdown();
+        }
+      }),
```
fuzzybinary: Why use a separate method if it's the Default vs any other flag client? From a polymorphism perspective, shouldn't we use `reset` in all circumstances here?

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,219 @@
...
+import 'evaluation_context.dart';
+import 'flag_assignments_fetcher.dart';
+import 'flags_store.dart';
+import 'json_value.dart';
+
+class FlagsRepository {
+  static const defaultStoreReadTimeout = Duration(milliseconds: 100);
+
+  @visibleForTesting
+  static Duration storeReadTimeout = defaultStoreReadTimeout;
```
fuzzybinary: Can this be a member variable instead of a static variable?

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,219 @@
...
+  FlagsEvaluationContext? get context => _state?.context;
+
+  FlagAssignment? flagAssignment(String key) => _state?.flags[key];
+
+  Future<void> initialize(FlagsEvaluationContext context) async {
+    final requestId = ++_contextRequestId;
+    final cached = await _readCached();
+    if (requestId != _contextRequestId) {
+      return;
+    }
```
fuzzybinary: I kinda understand what we're going for here, but it might be better to look into [`CancelableOperation `](https://pub.dev/documentation/async/latest/async/CancelableOperation-class.html) instead (or alternatively [`CancelableCompleter`](https://pub.dev/documentation/async/latest/async/CancelableCompleter-class.html) which could prevent having to keep and check this context state variable. Instead, you store the current operation and, before you run a new one, cancel the old one, which I believe would complete the pending `Future`.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,219 @@
...
+
+    await _publishLiveAssignments(
+      requestId: requestId,
+      context: context,
+      liveAssignments: _fetchLiveAssignments(context),
+      clearOnFailure: matchingCached == null,
+    );
+  }
+
+  Future<({PrecomputedAssignments? assignments})> _fetchLiveAssignments(
```
fuzzybinary: Why is this returning a record?

fuzzybinary: Actually, this whole `_publish` / `_fetch` system could be drastically simplified by removing `_fetchLiveAssignments` completely.

```dart
Future<void> _publishLiveAssignments({
    required int requestId,
    required FlagsEvaluationContext context,
    required bool clearOnFailure,
}) async {
  try {
    final assignments = await fetcher.fetch(context);
    await _publishAssignments(
      requestId: requestId,
      context: context,
      assignments: assignments,
    );
  } catch(_) {
    if (clearOnFailure && requestId == _contextRequestId) {
      _state = null;
    }
  }
}
```

packages/datadog_flags/lib/datadog_flags.dart
```diff
@@ -9,4 +9,5 @@ export 'src/datadog_flags.dart' show DatadogFlags;
 export 'src/flags_client.dart' show DatadogFlagsClient, FlagDetails;
 export 'src/flags_configuration.dart' show DatadogFlagsConfiguration;
 export 'src/flags_error.dart' show FlagEvaluationError;
+export 'src/flags_store.dart' show DatadogFlagsStore;
```
sameerank: Maybe the PR description needs to be updated? This diff doesn't match the description

> `FileDatadogFlagsStore` is exported from `datadog_flags_io.dart` so the default `datadog_flags.dart` entrypoint remains browser-safe

There are a couple other mentions of `datadog_flags_io.dart` too which doesn't show up in the diff

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  Future<void> _publishAssignments({
+    required int requestId,
+    required FlagsEvaluationContext context,
+    required PrecomputedAssignments assignments,
+  }) async {
+    if (requestId != _contextRequestId) {
+      return;
+    }
```
fuzzybinary: Even if you don't end up using a `CancelableFuture`, if we're having to check this much, it starts to be worth encapsulating the entire operation into its own Operation class, which checks at each step if has been canceled and terminates the operation if so.

I worry that the extra `_contextRequestId` introduces mental overhead that's just not needed if the full operation was its own class. It's easier to test as well.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,219 @@
...
+    return current != null &&
+        _contextsMatch(current.context, data.context) &&
+        data.date.isBefore(current.date);
+  }
+
+  DateTime _nextStateDate() {
+    final now = dateProvider();
+    final current = _state?.date;
+    if (current != null && !now.isAfter(current)) {
+      return current.add(const Duration(microseconds: 1));
```
fuzzybinary: Why do we add a microsecond here?

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -0,0 +1,219 @@
...
+  if (left.targetingKey != right.targetingKey) {
+    return false;
+  }
+
+  try {
+    return jsonEncode(_sortedJson(sanitizeJsonValue(left.attributes))) ==
+        jsonEncode(_sortedJson(sanitizeJsonValue(right.attributes)));
+  } catch (_) {
+    return false;
+  }
```
fuzzybinary, line 161: Is this actually the fastest way to do this? Maybe it doesn't matter because it doesn't happen often or because the contexts are relatively small, but it feels like serializing them to JSON would have a lot of unnecessary overhead especially as you're already having to perform a sort on all the keys, so you're visting everything anyway....

How restrictive would it be to write overload `equals` for `FlagsEvaluationContext`?

packages/datadog_flags/lib/src/flags_store.dart
```diff
@@ -0,0 +1,78 @@
...
+
+import 'dart:convert';
+
+import 'package:meta/meta.dart';
+
+import 'assignment.dart';
+import 'evaluation_context.dart';
+
+abstract class DatadogFlagsStore {
+  Future<Map<String, Object?>?> read(String clientName);
```
fuzzybinary: Does it makes sense for this to operate on `FlagsData` instead of JSON directly?

fuzzybinary (commented):
Just some questions -- I do think it might be worth looking into either using a `CancelableFuture` or refactoring the caching operations into class to avoid the constant checking of the _contextRequestId.

Just let me know if you don't think it's worth it though.

packages/datadog_flags/test/datadog_flags_test.dart
```diff
@@ -185,10 +186,10 @@ void main() {
     final second = client.initialize(
       const FlagsEvaluationContext(targetingKey: 'user-second'),
     );
-    await Future<void>.delayed(Duration.zero);
+    await _waitUntil(() => responseCompleters.length == 1);
 
-    expect(responseCompleters, hasLength(2));
-    responseCompleters[1].complete(
+    expect(responseCompleters, hasLength(1));
```
sameerank: Intentional change to the test? The test asserts

https://github.com/DataDog/dd-sdk-flutter/blob/d0c33a36d7433a3ace911efe090cc3fd5937995f/packages/datadog_flags/test/datadog_flags_test.dart#L170

i.e. that for 2 in-flight responses, a delayed first response won't clobber a newer second one. And this guard against stale responses is no longer tested with only one response

https://github.com/DataDog/dd-sdk-flutter/blob/71f5afccd358ca560e3801c1b1b7a2fdb0e95220/packages/datadog_flags/lib/src/flags_repository.dart#L102-L104

And I'd double check if this was a necessary change to support persistence

packages/datadog_flags/lib/src/flags_repository.dart
sameerank: Fine as a class variable for the current implementation, but might work better as an instance variable if we write tests and don't want timeouts for different stores to interfere with each other

packages/datadog_flags/lib/datadog_flags.dart
leoromanovsky: Yes! Thanks for noticing. Original PR had 2k LOC and some bike-shedding that I undid forgot to update the body.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -18,102 +17,55 @@ class FlagsRepository {
...
 
   Future<void> _publishAssignments({
-    required int requestId,
-    required FlagsEvaluationContext context,
+    required _InitializeOperation operation,
     required PrecomputedAssignments assignments,
   }) async {
-    if (requestId != _contextRequestId) {
+    if (!operation.isCurrent) {
       return;
```
fuzzybinary: This is already waaay better, but I think it still requires a weird bit of state sharing. 

Since these methods are only really accessed from `_InitializeOperation`, you can easily just have a `cancel` function on `_InitializeOperation` which is called before it's nulled out in `clearMemory() or replaced when `initialize()` is called.

Then in the operation, before calling `_publishAssignments` or `_publishStoredAssignments`, first check to see if the operation has been canceled, and early out if it has.

You could have `isCurrentOperation` as a failsafe, but it really should be owned by the repository so that `_InitializeOperation` is more encapsulated, and doesn't have to look at what operation is on the repository.

I'm actually a bit surprised Dart is letting `_InitializeOperation` access private members of `FlagsRepository`.....

packages/datadog_flags/lib/src/flags_repository.dart
leoromanovsky: Code now uses direct deep JSON-value comparison, not JSON stringification.

packages/datadog_flags/lib/src/flags_repository.dart
leoromanovsky: Thanks for asking; I over-thought the tie-breaker between disk cache and network. Removing it.

sameerank (dismissed):

fuzzybinary (approved):

PR #1057, opened 2026-06-08
fix(gql): Fix GQL errors to omit unused properties.

fuzzybinary:
### What and why?

When the Android SDK deserializes the GQL Errors property, JSON `null`s are deserialized as JSonNull, not a Kotlin null, which breaks its translation into the Attributes map.

The easy fix is to omit any properties that are missing, which makes them null on the Kotlin side.

### Review checklist


fuzzybinary:
@codex review

kikoveiga (approved):

Issue #1059, opened 2026-06-09
Migrate Plugin to Built-in Kotlin


pgiacomo69:
### Feature description

I am using datadog_flutter_plugin in my Flutter app.

Starting with Android Gradle Plugin (AGP) 9.0, support for applying the Kotlin Gradle Plugin (KGP) has been removed. Because this plugin applies KGP, it causes a compilation error that prevents my app from building. Here is an example of the error [Issue #181383](https://github.com/flutter/flutter/issues/181383).

Flutter has temporarily added support to allow KGP while apps and plugins migrate to AGP 9.0+, but this support will be removed in a future version of Flutter.

### Proposed solution

Please migrate this plugin to use built-in Kotlin to ensure your plugin users can successfully build their apps in future versions of Flutter.

Here is the Flutter [migration guide for plugin authors](https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors).

fuzzybinary:
Hi @pgiacomo69,

Thanks for reporting. This is on our radar but since Flutter added a workaround I hadn't prioritized it.

It looks like the guide also supplies a way to support Flutter version before 3.44, so we should be able to supply this soon without changing our minimum supported version above Dart 3.6 in 3.x release.

fuzzybinary:
We don't fully support AGP 9 and built in Kotlin yet because some dependencies don't support it, but much or our work for this now complete.  The next major version of the plugin should fully support AGP 9 and built-in Kotlin.

PR #1065, opened 2026-06-23
feat: Update all gradle files to support AGP 9

fuzzybinary:
### What and why?

Upgrade all of our gradle files to support AGP 9 and its built in Kotlin support.

Because our minimum supported version of Flutter is 3.27, we need to add the check for AGP 9 and re-add Kotlin if it's not there.

refs: #1059

### Review checklist


kikoveiga:
Should we bump to AGP 9 in the example apps to test the sdk is behaving well there? There are some weird changes with this new agp, we are having some problems on the native side so better be sure.
Also, there is still an occurrence of `android.builtInKotlin=false` in `datadog_webview_tracking`, which means for AGP 9 there will be no Kotlin dependency there and the `kotlin {}` block will fail. Can we remove that?

fuzzybinary:
@kikoveiga yeah it's a good callout. I noticed last night that the upgrade didn't include upgrading the examples.

I'll to the upgrade and see what happens. Hopefully it doesn't break our min-version checks.

fuzzybinary:
This ended up being a good callout. While we've updated to AGP 9, we have some dependencies that haven't yet, and that causes some interesting issues. We now guard against both an AGP>= 9 and `builtInKotlin` both being true before attempting to use the `kotlin` block.

examples/native-hybrid-app/android/app/build.gradle
```diff
@@ -5,10 +5,14 @@ buildscript {
 
 plugins {
     id 'com.android.application'
-    id 'org.jetbrains.kotlin.android'
     id("com.datadoghq.dd-sdk-android-gradle-plugin") version "1.24.0"
 }
 
+def agpMajor = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
```
kikoveiga: nit and same in other files:

```suggestion
import com.android.Version // in the beginning of the file
def agpMajor = Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0] as int
```

examples/native-hybrid-app/android/app/build.gradle
```diff
@@ -34,14 +38,17 @@ android {
...
-        jvmTarget = '1.8'
-    }
     buildFeatures {
         viewBinding true
     }
 }
 
+kotlin {
+    compilerOptions {
+        jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8
```
kikoveiga: Same as above:
```suggestion
import org.jetbrains.kotlin.gradle.dsl.JvmTarget

        jvmTarget = JvmTarget.JVM_1_8
```

packages/datadog_session_replay/example/assets/fonts/Open_Sans/README.txt
kikoveiga, line 1: Do we need this `README.txt` file and all the unused `ttf` fonts? Or we can just include the ones needed?

packages/datadog_session_replay/example/assets/fonts/OpenSans-Regular.ttf
kikoveiga, line 1: So we can remove the `assets/fonts/OpenSans-regular.ttf` file.

kikoveiga (dismissed):
Looks good, there are still 2 example apps on AGP 8 though:
- `examples/native-hybrid-app`
- `packages/datadog_inappwebview_tracking/example`

fuzzybinary:
> Looks good, there are still 2 example apps on AGP 8 though:
> 
> * `examples/native-hybrid-app`
> * `packages/datadog_inappwebview_tracking/example`

@kikoveiga Yeah, good eyes. `Inappwebview` itself hasn't updated to AGP 9, and it's possible we will have to deprecate that packages as a result (it appears to be abandoned).

Native hybrid app needs a full rework for AGP 9, but is just an example so I'm okay leaving it for now.

packages/datadog_session_replay/example/assets/fonts/Open_Sans/README.txt
fuzzybinary: Technically, no, however I'm inclined to keep them just to keep everything intact from a Licensing perspective (and also in case we decide to use them :) )

fuzzybinary: Actually, let me check if `pub publish` sends these up. If it does, I'll remove them.

kikoveiga (approved):

PR #1068, opened 2026-06-25
feat: Migrate main package to a federated plugin.

fuzzybinary:
### What and why?

As we start looking to potentially add more platforms, it's starting to make sense for us to move from a single monolithic package to a Flutter [federated plugin](https://docs.flutter.dev/packages-and-plugins/developing-packages#federated-plugins).  Over the long term, this will allow us to release the various platform libraries independently, especially if there are minor bug fixes, improvements, or features that are specific to one platform.

Because a lot of the Datadog Flutter Plugin relies on sending information to the native platform, a lot more data is held in the "platform_interface" (datadog_flutter_plugin_platform_interface) than normal.  This is because the platform specific plugins can include the `platform_interface` but not the main package. The only portion in the main package are the features that are completely implemented in Dart.

This first portion only tackles the main package `datadog_flutter_plugin`, and should do so without any external changes. If users pulled this now, the change would be transparrent to them.

However, because of the size of this change, we're going to bundle it with other breaking changes slated for v4, so this is targeted to a temporary `v4` branch, instead of at `develop`

### Review checklist


janine-c (approved):
Looks like the only files the docs team owns were renamed, without there being content for us to review. Nothing to look at, but approving to unblock 🙂

packages/datadog_flutter_plugin/datadog_flutter_plugin/test/helper_test.dart
```diff
@@ -1,5 +1,8 @@
-import 'package:datadog_flutter_plugin/src/helpers.dart';
-import 'package:datadog_flutter_plugin/src/internal_logger.dart';
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2026-Present Datadog, Inc.
```
kikoveiga, line 3: I see this license part is not the same in all files, e.g. sometimes the year is `2016`, `2019`, etc. An idea: have a script/ci step to check for this, as we have on [android](https://github.com/DataDog/dd-sdk-android/blob/478f4af36faed2319c48650d16342b0d54ffe589/detekt_custom_general.yml#L56).

packages/datadog_flutter_plugin/datadog_flutter_plugin_platform_interface/lib/src/logs/ddlogs_noop_platform.dart
```diff
@@ -2,12 +2,15 @@
...
+import '../internal_logger.dart';
 import 'ddlogs_platform_interface.dart';
+import 'log_configuration.dart';
+import 'log_level.dart';
 
 class DdNoOpLogsPlatform extends DdLogsPlatform {
   @override
-  Future<void> enable(DatadogSdk core, DatadogLoggingConfiguration config) {
+  Future<void> enable(
+      InternalLogger logger, DatadogLoggingConfiguration config) {
```
kikoveiga, line 13: Is this a normal formatting in flutter? To me it looks weird, do we enforce it with any tool in this project?

packages/datadog_flutter_plugin/datadog_flutter_plugin/lib/src/rum/ddrum.dart
```diff
@@ -12,70 +12,50 @@ import 'package:flutter/services.dart';
...
+        DdRumPlatform,
+        LateConfigurationProperty;
+import 'package:datadog_flutter_plugin_platform_interface/datadog_flutter_plugin_platform_interface.dart'
+    show
+        DatadogAttachConfiguration,
+        DatadogRumConfiguration,
+        ResourceHeadersExtractor,
+        RumActionType,
+        RumErrorSource,
+        RumFeatureOperationFailureReason,
```
kikoveiga, line 32: If all this is coming in v4 only, might be a good time to rename all `FeatureOperation` public apis to just `Operation`, as well as internal apis, classes, variables, etc. This doesn't need to be done here ofc, could be in a follow-up ticket.

packages/datadog_flutter_plugin/datadog_flutter_plugin_android/example/android/local.properties
kikoveiga, line 1: This should be removed and added to `.gitignore`.

packages/datadog_flutter_plugin/datadog_flutter_plugin_android/example/android/gradle.properties
```diff
@@ -0,0 +1,16 @@
...
+#
+
+org.gradle.jvmargs=-Xmx4G
+android.useAndroidX=true
+android.defaults.buildfeatures.buildconfig=true
+android.nonTransitiveRClass=false
+android.nonFinalResIds=false
+kotlin_version=2.1.0
+# This builtInKotlin flag was added automatically by Flutter migrator
+android.builtInKotlin=false
```
kikoveiga, line 14: For v4, we should discuss this. With the new Android API 37, there is a new AGP 9 which provides built-in kotlin, which we will want to enable by default. We will support this in Android v4, so it would make sense for Flutter v4 to support it as well I would say. This means we would also have to bump a bunch of dependencies here, including the `gradle` version to `9.x`.

packages/datadog_flutter_plugin/datadog_flutter_plugin_android/example/android/.gradle/8.11.1/checksums/checksums.lock
kikoveiga, line 1: I think `.gradle/` folder should be gitgnored and all these checksums files removed.

packages/datadog_flutter_plugin/datadog_flutter_plugin_ios/ios/datadog_flutter_plugin_ios/.swiftpm/xcode/xcuserdata/jeff.ward.xcuserdatad/xcschemes/xcschememanagement.plist
kikoveiga, line 1: Should this be gitignored and removed as well?

packages/datadog_flutter_plugin/datadog_flutter_plugin_ios/melos_datadog_flutter_plugin_ios.iml
kikoveiga, line 1: Can we remove all these `melos_*.iml` files?

packages/datadog_flutter_plugin/datadog_flutter_plugin_android/example/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java
kikoveiga, line 1: Is this file needed in the repo?

packages/datadog_flutter_plugin/datadog_flutter_plugin/example/ios/RunnerTests/RunnerTests.swift
kikoveiga, line 1: Why is this needed?

packages/datadog_flutter_plugin/datadog_flutter_plugin/lib/src/tracing/baggage_helpers.dart
```diff
@@ -142,13 +144,17 @@ String _constructBaggageHeader(
   return header;
 }
 
-String mergeW3CBaggageHeader(TracingContext context, String? baggageHeader) {
-  final logger = DatadogSdk.instance.internalLogger;
+String mergeW3CBaggageHeader(
+  TracingContext context,
+  String? baggageHeader, [
+  InternalLogger? logger,
```
kikoveiga, line 150: Is a non-null `logger` here better? It's also consistent with the other functions in this file.

packages/datadog_flutter_plugin/integration_test_app/ios/Flutter/AppFrameworkInfo.plist
```diff
@@ -20,7 +20,5 @@
   <string>????</string>
   <key>CFBundleVersion</key>
   <string>1.0</string>
-  <key>MinimumOSVersion</key>
-  <string>13.0</string>
```
kikoveiga, line 24: I see this is the only `AppFrameworkInfo.plist` file where this happened, is it intentional?

kikoveiga (commented):
- `generate_env.sh` needs to be updated to support the new dirs, specifically L15, L16 & L33.
- Do we need to add `CHANGELOG`, `README` & `LICENSE` files to the 4 new sub-packages?

packages/datadog_flutter_plugin/datadog_flutter_plugin/test/helper_test.dart
fuzzybinary: It's the year the file is created (or the license was added). I'm not sure which is more correct -- the year the library is created or the year the file is created.... I'll look it up.

packages/datadog_flutter_plugin/datadog_flutter_plugin_platform_interface/lib/src/logs/ddlogs_noop_platform.dart
fuzzybinary: Yes. `dartfmt` is incredibly strict, and CI (now) checks it.

Usually preference for this would be to add a trailing comma, which slightly changes the format to:
```dart
Future<void> enable(
  InternalLogger logger,
  DatadogLoggingConfiguration config,
) {
```

But this is also correct by Dart standard.

packages/datadog_flutter_plugin/datadog_flutter_plugin/lib/src/rum/ddrum.dart
fuzzybinary: Yes, but that will be a separate PR. This is only focused on federating the packages.

packages/datadog_flutter_plugin/datadog_flutter_plugin_android/example/android/gradle.properties
fuzzybinary: Yeah -- so ignore this for now. `develop` has [this PR](https://github.com/DataDog/dd-sdk-flutter/pull/1065) which updates to AGP 9.

BUT as I found out, until *all* of our dependencies are on AGP 9 we still have to make sure the `android.builtInKotlin` is false.

v3 will actually update to mostly support AGP 9, but full support will be added in v4 with the JNI update.

packages/datadog_flutter_plugin/datadog_flutter_plugin_ios/ios/datadog_flutter_plugin_ios/.swiftpm/xcode/xcuserdata/jeff.ward.xcuserdatad/xcschemes/xcschememanagement.plist
fuzzybinary: Yes, all *.xcuserdata should be. I'll go back through and double check.

packages/datadog_flutter_plugin/datadog_flutter_plugin_ios/melos_datadog_flutter_plugin_ios.iml
fuzzybinary: I *think* those are supposed to be committed, but I will double check.

packages/datadog_flutter_plugin/datadog_flutter_plugin_android/example/android/app/src/main/java/io/flutter/plugins/GeneratedPluginRegistrant.java
fuzzybinary: I think these are generally committed, but I'll double check.

packages/datadog_flutter_plugin/datadog_flutter_plugin/example/ios/RunnerTests/RunnerTests.swift
fuzzybinary: It's likely not. I actually thought I deleted it.

packages/datadog_flutter_plugin/integration_test_app/ios/Flutter/AppFrameworkInfo.plist
fuzzybinary: I think with the upgrade of Flutter it may have removed this intentionally. I will double check.

fuzzybinary:
@kikoveiga I think I addressed everything. Please give a quick once over and make sure I got everything.

kikoveiga (approved):
Looks good! Just some nits:

- There is still the question on the nullable `logger` in `baggage_helpers.dart`.
- Found some `unnecessary_import` when running stuff locally in `datadog_webview.dart:5`, `datadog_webview_tracking_test.dart:5,` and `datadog_inappwebview_user_script.dart:5`.
- Do we need an iOS person to review that part of the code?

fuzzybinary:
> Looks good! Just some nits:
> 
> * There is still the question on the nullable `logger` in `baggage_helpers.dart`.

Going to keep it for now. I'll probably revisit it though.

> * Found some `unnecessary_import` when running stuff locally in `datadog_webview.dart:5`, `datadog_webview_tracking_test.dart:5,` and `datadog_inappwebview_user_script.dart:5`.

👍 I'll actually address that in an immediate followup PR, as well as figure out why they're not causing build failures (my guess is they're just infos?).

> * Do we need an iOS person to review that part of the code?

The iOSs code is mostly moved or boilerplate. I think we're okay.

PR #1069, opened 2026-06-25
feat(rum): parse route query string into view.url and url_query attributes

seunghwanly:
DatadogNavigationObserver only used `route.settings.name` as the view identifier, so mobile RUM views never populated `url_query` facets the way the browser SDK does.

`defaultViewInfoExtractor` now parses a query string in the route name:
- the full path+query is reported as `view.url` so Datadog can derive the standard `@view.url_query.*` facets server-side (as on web)
- each query parameter is also added as a `url_query.<key>` view attribute (`@context.url_query.<key>`) as a guaranteed fallback

Routes without a query string keep the previous behavior exactly.

### What and why?

`DatadogNavigationObserver` identifies RUM views solely by `route.settings.name`. On the browser SDK, Datadog derives the `@view.url_query.*` facets from the page URL, but on mobile a view's `url` is just the route name with no query component — so those facets are never available for mobile views.

This PR teaches the default view extractor to understand query strings in route names. Apps that navigate with query-bearing routes (for example `Navigator.pushNamed(context, '/products?category=shoes&id=123')`) can then search, filter, and group mobile RUM views by individual query parameters — the same workflow already available on web.

### How?

`defaultViewInfoExtractor` now delegates to a new `rumViewInfoFromRouteName` helper:

- **No query string** → behavior is unchanged: the route name is used as-is, with no extra attributes.
- **Has a query string** (for example `/products?category=shoes&id=123`):
  - `view.url` carries the full path + query (`/products?category=shoes&id=123`), letting Datadog derive the standard `@view.url_query.*` facets server-side.
  - `view.name` is the clean path (`/products`).
  - Each query parameter is additionally emitted as a `url_query.<key>` view attribute (surfaced as `@context.url_query.<key>`), a guaranteed fallback that does not rely on backend URL parsing. Multi-valued parameters (`?id=1&id=2`) are reported as a list.

`RumViewInfo` gains `viewKey` (→ `view.url`) and `viewName` (→ `view.name`) getters so `startView` / `stopView` / `markViewFirstBuildComplete` consistently key off the query-bearing URL. When there is no distinct path, `viewName` is `null`, keeping the previous `startView(name, null, attributes)` call identical. The same parsing also flows through `DatadogRouteAwareMixin`.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
@@ -15,9 +15,8 @@ class RumViewInfo {
 
   /// An optional path or URL for the view, reported as `view.url`. When this
   /// contains a query string (for example `/products?category=shoes`), Datadog
-  /// parses it server-side into the standard `@view.url_query.*` facets, the
-  /// same way it does for the browser SDK. When `null`, [name] is used as the
-  /// key and the previous behavior is preserved.
+  /// derives the standard `@view.url_query.*` facets from it server-side. When
+  /// `null`, [name] is used as the view key.
   final String? path;
 
   /// Any attributes to be associated with this view
@@ -35,8 +34,7 @@ class RumViewInfo {
   String get viewKey => path ?? name;
 
   /// The display name passed to [DatadogRum.startView]. Returns `null` when no
-  /// distinct [path] is provided, preserving the original behavior where the
-  /// view key and name were identical.
+  /// distinct [path] is provided, so the view key is used as the name.
   String? get viewName => path != null ? name : null;
 }
 
@@ -66,15 +64,12 @@ RumViewInfo? defaultViewInfoExtractor(Route<dynamic> route) {
 /// it contains.
 ///
 /// When [routeName] includes a query string (for example
-/// `/products?category=shoes&id=123`):
-///   * `view.url` carries the full path + query string, letting Datadog derive
-///     the standard `@view.url_query.*` facets server-side (as it does on web);
-///   * each query parameter is additionally reported as a `url_query.<key>`
-///     view attribute (surfaced as `@context.url_query.<key>`), a guaranteed
-///     fallback that does not depend on backend URL parsing.
+/// `/products?category=shoes&id=123`) the `view.url` will carry the full path +
+/// query string, which Datadog will add to its standard `@view.url_query.*` facets.
+/// Additionally, each parameter will be reported as a `url_query.<key>` view
+/// attribute, accessible through `@context.url_query.<key>`.
 ///
-/// Routes without a query string keep the previous behavior exactly: the name
-/// is used as-is and no extra attributes are added.
+/// Routes without a query string use the name as-is and add no extra attributes.
 RumViewInfo rumViewInfoFromRouteName(String routeName) {
   if (!routeName.contains('?')) {
     return RumViewInfo(name: routeName);
--- a/packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
@@ -66,8 +66,6 @@ RumViewInfo? defaultViewInfoExtractor(Route<dynamic> route) {
 /// When [routeName] includes a query string (for example
 /// `/products?category=shoes&id=123`) the `view.url` will carry the full path +
 /// query string, which Datadog will add to its standard `@view.url_query.*` facets.
-/// Additionally, each parameter will be reported as a `url_query.<key>` view
-/// attribute, accessible through `@context.url_query.<key>`.
 ///
 /// Routes without a query string use the name as-is and add no extra attributes.
 RumViewInfo rumViewInfoFromRouteName(String routeName) {
@@ -89,17 +87,9 @@ RumViewInfo rumViewInfoFromRouteName(String routeName) {
   // The path without the query string, used as the human-readable view name.
   final viewName = uri.path.isNotEmpty ? uri.path : routeName.split('?').first;
 
-  // Decompose the query string into per-parameter attributes. Multi-valued
-  // parameters (for example `?id=1&id=2`) are reported as a list.
-  final attributes = <String, Object?>{};
-  uri.queryParametersAll.forEach((key, values) {
-    attributes['url_query.$key'] = values.length == 1 ? values.first : values;
-  });
-
   return RumViewInfo(
     name: viewName.isNotEmpty ? viewName : routeName,
     path: routeName,
-    attributes: attributes,
   );
 }
 
--- a/packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
@@ -10,10 +10,14 @@ import '../../datadog_internal.dart';
 
 /// Information about a View that will be passed to [DatadogRum.startView]
 class RumViewInfo {
-  /// The name of the view
+  /// The name of the view, reported as `view.name` in the RUM explorer.
   final String name;
 
-  /// A path to the view
+  /// An optional path or URL for the view, reported as `view.url`. When this
+  /// contains a query string (for example `/products?category=shoes`), Datadog
+  /// parses it server-side into the standard `@view.url_query.*` facets, the
+  /// same way it does for the browser SDK. When `null`, [name] is used as the
+  /// key and the previous behavior is preserved.
   final String? path;
 
   /// Any attributes to be associated with this view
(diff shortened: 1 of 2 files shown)


packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
```diff
@@ -10,10 +10,14 @@ import '../../datadog_internal.dart';
...
-  /// The name of the view
+  /// The name of the view, reported as `view.name` in the RUM explorer.
   final String name;
 
-  /// A path to the view
+  /// An optional path or URL for the view, reported as `view.url`. When this
+  /// contains a query string (for example `/products?category=shoes`), Datadog
+  /// parses it server-side into the standard `@view.url_query.*` facets, the
+  /// same way it does for the browser SDK. When `null`, [name] is used as the
+  /// key and the previous behavior is preserved.
```
fuzzybinary: No need to mention "previous behavior" in public comments (there are a few instances of this). We only need to document current behavior. You can also remove the reference to the browser SDK and just document the behavior we support.

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
```diff
@@ -24,6 +28,16 @@ class RumViewInfo {
...
   });
+
+  /// The key that identifies this view to RUM, reported as `view.url`. Prefers
+  /// [path] (which may include a query string, so the backend can derive
+  /// `@view.url_query.*`) and falls back to [name].
+  String get viewKey => path ?? name;
+
+  /// The display name passed to [DatadogRum.startView]. Returns `null` when no
+  /// distinct [path] is provided, preserving the original behavior where the
+  /// view key and name were identical.
```
fuzzybinary: Remove reference to "original behavior".

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
```diff
@@ -36,16 +50,62 @@ class RumViewInfo {
...
+/// Builds a [RumViewInfo] from a raw route name, parsing any URL query string
+/// it contains.
+///
+/// When [routeName] includes a query string (for example
+/// `/products?category=shoes&id=123`):
+///   * `view.url` carries the full path + query string, letting Datadog derive
+///     the standard `@view.url_query.*` facets server-side (as it does on web);
+///   * each query parameter is additionally reported as a `url_query.<key>`
+///     view attribute (surfaced as `@context.url_query.<key>`), a guaranteed
+///     fallback that does not depend on backend URL parsing.
```
fuzzybinary: I don't think this needs to be bullet points, and this can be cleaned up slightly

```suggestion
/// When [routeName] includes a query string (for example
/// `/products?category=shoes&id=123`) the `view.url` will carry the full path + 
/// query string, which Datadog will add to its standard `@view.url_query.*` facets.
/// Additionally, each parameter will be reported reported as a `url_query.<key>` view 
/// attribute, accessible through `@context.url_query.<key>`.
```

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
```diff
@@ -36,16 +50,62 @@ class RumViewInfo {
...
+
+  // The path without the query string, used as the human-readable view name.
+  final viewName = uri.path.isNotEmpty ? uri.path : routeName.split('?').first;
+
+  // Decompose the query string into per-parameter attributes. Multi-valued
+  // parameters (for example `?id=1&id=2`) are reported as a list.
+  final attributes = <String, Object?>{};
+  uri.queryParametersAll.forEach((key, values) {
+    attributes['url_query.$key'] = values.length == 1 ? values.first : values;
+  });
```
fuzzybinary: Let me verify with the team we're okay adding this part. This is a great add because it bring Flutter's behavior more in line with the Browser SDK, but then this actually causes it to diverge a bit.

If we decide not to do this, you can still add this to your own product by overriding view info extractor.

packages/datadog_flutter_plugin/test/rum/navigation_observer_test.dart
```diff
@@ -244,6 +244,65 @@ void main() {
...
+          '/products?category=shoes&id=123',
+          '/products',
+          {'url_query.category': 'shoes', 'url_query.id': '123'},
+        ));
+    verify(() =>
+        mockRum.markViewFirstBuildComplete('/products?category=shoes&id=123'));
+  });
+
+  group('rumViewInfoFromRouteName', () {
+    test('route without a query string keeps the previous behavior', () {
```
fuzzybinary: Don't use "previous behavior" here. Describe the behavior you want. Ideally, we already had tests checking this but it's possible that's not the case.

fuzzybinary (changes requested):
This looks good to me mostly, mostly some comment style issues. Wait for me to confirm about whether we're okay automatically adding the query parameters to the view attributes before proceeding.

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
seunghwanly: You're right, I was over-documenting here. Thanks for catching it — fixed.

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
seunghwanly: Done, removed.

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
seunghwanly: Applied your suggestion.

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
seunghwanly: Completely understand, happy to follow whatever the team decides. 

For context: as a daily Datadog user, I've found it awkward to query or segment events by individual deep-link parameters in RUM and Product Analytics — that's what prompted this PR. I do see how it diverges from the other SDKs, though, so no concerns if you'd rather keep it out of the default.

packages/datadog_flutter_plugin/test/rum/navigation_observer_test.dart
seunghwanly: Renamed it to describe the behavior directly — wanted at least a minimal unit test here, so I made the title explicit about what it checks.

seunghwanly:
> This looks good to me mostly, mostly some comment style issues. Wait for me to confirm about whether we're okay automatically adding the query parameters to the view attributes before proceeding.

Thanks for the review! I've addressed all the comment/doc style points. I'll wait for the team's confirmation.

packages/datadog_flutter_plugin/lib/src/rum/navigation_observer.dart
fuzzybinary: Okay, the consensus at Datadog is we should not add these attributes. The backend should properly parse the query parameters if they're supplied in the `path` portion of the view, which you can then query like you would a Browser based view. and if it doesn't we should fix that on the backend.

Adding these as "built in" attributes supported by Datadog would require some coordination with other SDKs and supporting the attributes long term, where we already have a solution built into the URL parsing.

Once you have that removed we can look into adding this.

Thanks again!

fuzzybinary:
@seunghwanly See my above comment. If you could remove adding the query params to the attributes, we can get this built and submitted.

seunghwanly:
@fuzzybinary Thanks for the clarification :)

I think my previous change went a bit too far, so I removed the `url_query.<key>` custom attributes. I hope this PR now makes sure that when a `RumViewInfo.path` is provided, it is used as the RUM view key, so the native SDKs can report it as the View URL.

Thanks again for the review!

fuzzybinary (approved):

fuzzybinary:
Ugh... sorry @seunghwanly - your commits need to be signed. It looks like first one was but this last one wasn't. If you can get that fixed we'll get this merged in.

seunghwanly:
Ah, my bad — the previous commits weren’t signed. I’ve force-pushed the signed commits, so this should be ready to merge now. Thanks!

fuzzybinary (approved):

fuzzybinary:
Hi. sorry @seunghwanly one more time -- I need you to merge develop into your branch to fix a proguard issue caused by an update in our Android library.

seunghwanly:
> Hi. sorry @seunghwanly one more time -- I need you to merge develop into your branch to fix a proguard issue caused by an update in our Android library.

Done! 👍 Merged develop into the branch.

fuzzybinary:
@seunghwanly It also needs to be signed 😓

barboraplasovska (approved):

seunghwanly:
Sorry for the noise here. My local Git signing setup was in a bit of flux.. I’ve cleaned up the branch and pushed the intended signed commits now. Thanks

PR #1071, opened 2026-06-26
feat(datadog_flags_flutter): add Flutter integration package

leoromanovsky:
## Motivation

Flutter customers should be able to use Datadog feature flags with the Datadog Flutter SDK without coupling the flags integration release cadence to the core `datadog_flutter_plugin` package. The standalone `datadog_flags` package remains the right API for pure Dart apps and custom lifecycle control, while Flutter apps should have a small integration package that derives Datadog SDK configuration and adds successful flag evaluations to RUM.

## Changes

This PR moves the Flutter integration into a new `datadog_flags_flutter` package. Apps import `package:datadog_flags_flutter/datadog_flags_flutter.dart`, register `DatadogFlagsPluginConfiguration` with `DatadogConfiguration.addPlugin`, and then use `DatadogSdk.instance.flags?.sharedClient()` to evaluate flags through the same typed API provided by `datadog_flags`.

The new package depends on `datadog_flags: ^1.0.0` and `datadog_flutter_plugin: ^3.4.0`. The core Flutter plugin only keeps the small support needed by external plugin integrations: it exposes the active `DatadogConfiguration` during initialization and allows const plugin configuration classes. It does not depend on `datadog_flags` and does not export the flags API.

Successful evaluations still flow through the standalone Flags SDK for precompute assignment reads, exposure telemetry, flag evaluation telemetry, persistence, shutdown, and reset behavior. The Flutter package adds SDK-owned setup and RUM feature flag evaluation tagging through `DatadogRum.addFeatureFlagEvaluation`.

```mermaid
flowchart LR
  app["Flutter app"] --> flutterFlags["datadog_flags_flutter\nthis PR"]
  app --> core["datadog_flutter_plugin"]
  flutterFlags --> flags["datadog_flags\nstandalone Dart SDK"]
  flutterFlags --> rum["Datadog RUM\nfeature flag evaluations"]
  core --> flutterFlags
  flags --> precompute["Precompute assignments API"]
  flags --> evp["Exposure and flag evaluation telemetry"]
```

## Decisions

`datadog_flags_flutter` is a separate package so feature flag integration updates can ship independently from the core Flutter SDK after the one-time core support lands. The package requires `datadog_flutter_plugin` 3.4.0 because it needs the active SDK configuration to derive client token, environment, site, RUM application ID, service, and version.

Unsupported Datadog sites do not throw during setup. If a caller uses an unsupported Flutter SDK site without an explicit `DatadogFlagsConfig`, the integration leaves the underlying Flags SDK unconfigured and clients return programmatic defaults through the standalone SDK behavior.

The simple example now demonstrates the integrated package instead of manually enabling `DatadogFlags.instance`, while keeping credentials in the ignored `.env` flow.

## Validation

Screenshot 2026-06-29 at 6 45 01 PM

### DIFF
--- a/examples/simple_example/README.md
+++ b/examples/simple_example/README.md
@@ -23,7 +23,7 @@ names, org names, or customer-owned flag keys.
 
 ## Feature Flags
 
-The `Flags` screen initializes the Flutter-integrated flags package, refreshes
+The `Flags` screen initializes the `datadog_flags_flutter` package, refreshes
 assignments for one evaluation context, and evaluates boolean, string, integer,
 double, and JSON flags with programmatic defaults.
 
--- a/examples/simple_example/README.md
+++ b/examples/simple_example/README.md
@@ -6,7 +6,7 @@ This is a more realistic example of how you use the Datadog Flutter Plugin in a 
     * Automatic network tracing with `datadog_tracking_http_client`
     * User interaction tracking with RumUserActionDetector
     * Error/Crash handling with manually reported errors
-    * Basic feature flag initialization and typed evaluation with `datadog_flags`
+    * Basic feature flag initialization and typed evaluation with `datadog_flags_flutter`
 
 ## Setup
 
@@ -23,9 +23,9 @@ names, org names, or customer-owned flag keys.
 
 ## Feature Flags
 
-The `Flags` screen initializes `DatadogFlags`, refreshes assignments for one
-evaluation context, and evaluates boolean, string, integer, double, and JSON
-flags with programmatic defaults.
+The `Flags` screen initializes the Flutter-integrated flags package, refreshes
+assignments for one evaluation context, and evaluates boolean, string, integer,
+double, and JSON flags with programmatic defaults.
 
 To test feature flags in your own organization, customize the generated `.env`
 file:
--- a/examples/simple_example/lib/main.dart
+++ b/examples/simple_example/lib/main.dart
@@ -3,7 +3,6 @@
 // Copyright 2023-Present Datadog, Inc.
 
 import 'package:datadog_flutter_plugin/datadog_flutter_plugin.dart';
-import 'package:datadog_flags/datadog_flags.dart';
 import 'package:datadog_gql_link/datadog_gql_link.dart';
 import 'package:datadog_session_replay/datadog_session_replay.dart';
 import 'package:datadog_tracking_http_client/datadog_tracking_http_client.dart';
@@ -67,6 +66,11 @@ void main() async {
     siteName: dotenv.maybeGet('DD_SITE'),
     applicationId: datadogConfig.rumConfiguration?.applicationId,
   );
+  datadogConfig.addPlugin(
+    DatadogFlagsPluginConfiguration(
+      flagsConfiguration: flagsConfig.configuration,
+    ),
+  );
 
   // runUsingRunApp(datadogConfig, flagsConfig);
   runUsingAlternativeInit(datadogConfig, flagsConfig);
@@ -94,7 +98,6 @@ Future<void> runUsingAlternativeInit(
   };
 
   await DatadogSdk.instance.initialize(datadogConfig, TrackingConsent.granted);
-  await DatadogFlags.instance.enable(configuration: flagsConfig.configuration);
   final link = Link.from([
     DatadogGqlLink(DatadogSdk.instance, Uri.parse(graphQlUrl)),
     HttpLink(graphQlUrl),
@@ -112,19 +115,15 @@ Future<void> runUsingRunApp(
   FlagsExampleConfig flagsConfig,
 ) async {
   await DatadogSdk.runApp(datadogConfig, TrackingConsent.granted, () {
-    DatadogFlags.instance
-        .enable(configuration: flagsConfig.configuration)
-        .then((_) {
-      final link = Link.from([
-        DatadogGqlLink(DatadogSdk.instance, Uri.parse(graphQlUrl)),
-        HttpLink(graphQlUrl),
-      ]);
-      final graphQlClient = GraphQLClient(link: link, cache: GraphQLCache());
+    final link = Link.from([
+      DatadogGqlLink(DatadogSdk.instance, Uri.parse(graphQlUrl)),
+      HttpLink(graphQlUrl),
+    ]);
+    final graphQlClient = GraphQLClient(link: link, cache: GraphQLCache());
 
-      runApp(MyApp(
-        graphQLClient: graphQlClient,
-        flagsConfig: flagsConfig,
-      ));
-    });
+    runApp(MyApp(
+      graphQLClient: graphQlClient,
+      flagsConfig: flagsConfig,
+    ));
   });
 }
--- a/examples/simple_example/lib/main.dart
+++ b/examples/simple_example/lib/main.dart
@@ -25,46 +25,55 @@ void main() async {
   configureUrlStrategy();
 
   DatadogSdk.instance.sdkVerbosity = CoreLoggerLevel.debug;
+  final siteConfig =
+      FlagsExampleSiteConfig.fromName(dotenv.maybeGet('DD_SITE'));
 
   final datadogConfig = DatadogConfiguration(
     clientToken: dotenv.get('DD_CLIENT_TOKEN', fallback: ''),
     env: dotenv.get('DD_ENV', fallback: ''),
-    site: DatadogSite.us1,
-    loggingConfiguration: DatadogLoggingConfiguration(),
+    site: siteConfig.datadogSite,
+    loggingConfiguration: DatadogLoggingConfiguration(
+      customEndpoint: siteConfig.logsCustomEndpoint,
+    ),
     firstPartyHosts: ['localhost'],
     rumConfiguration: DatadogRumConfiguration(
-        applicationId: dotenv.get('DD_APPLICATION_ID', fallback: ''),
-        traceSampleRate: 100.0,
-        trackResourceHeaders: ResourceHeadersExtractor(
-          captureHeaders: [
-            'accept-ranges',
-            'content-disposition',
-            'server',
-            'user-agent',
-            'via',
-            'x-cache-hits',
-            'x-served-by',
-            'x-datadog-trace-id',
-            'x-datadog-parent-id',
-            'x-datadog-origin',
-            'traceparent',
-          ],
-        )),
-  )
-    ..enableHttpTracking(
+      applicationId: dotenv.get('DD_APPLICATION_ID', fallback: ''),
+      customEndpoint: siteConfig.rumCustomEndpoint,
+      traceSampleRate: 100.0,
+      trackResourceHeaders: ResourceHeadersExtractor(
+        captureHeaders: [
+          'accept-ranges',
+          'content-disposition',
+          'server',
+          'user-agent',
+          'via',
+          'x-cache-hits',
+          'x-served-by',
+          'x-datadog-trace-id',
+          'x-datadog-parent-id',
+          'x-datadog-origin',
+          'traceparent',
+        ],
+      ),
+    ),
+  )..enableHttpTracking(
       // Using ignoreUrlPatterns is needed if you want to combine HttpClient
       // tracking and GraphQL tracking through datadog_gql_link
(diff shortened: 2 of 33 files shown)


packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
```diff
@@ -0,0 +1,322 @@
...
+  return switch (site) {
+    DatadogSite.us1 => DatadogFlagsSite.us1,
+    DatadogSite.us3 => DatadogFlagsSite.us3,
+    DatadogSite.us5 => DatadogFlagsSite.us5,
+    DatadogSite.eu1 => DatadogFlagsSite.eu1,
+    DatadogSite.ap1 => DatadogFlagsSite.ap1,
+    DatadogSite.ap2 => DatadogFlagsSite.ap2,
+    DatadogSite.us1Fed => null,
+  };
+}
```
leoromanovsky: Code like this was written before except it supported staging; I am considering whether this is duplicated and should be in a more canonical place.

packages/datadog_flutter_plugin/lib/src/datadog_plugin.dart
```diff
@@ -7,6 +7,8 @@ import 'package:flutter/material.dart';
 import '../datadog_flutter_plugin.dart';
 
 abstract class DatadogPluginConfiguration {
+  const DatadogPluginConfiguration();
```
leoromanovsky, line 10: wdyt?

examples/simple_example/lib/screens/flags_screen.dart
```diff
@@ -22,23 +23,31 @@ class FlagsScreen extends StatefulWidget {
 }
 
 class _FlagsScreenState extends State<FlagsScreen> {
-  late final DatadogFlagsClient _client;
+  late final DatadogFlagsClient? _client;
```
fuzzybinary: Does this have to be `late` if it's optional? Maybe because it's `final`?

examples/simple_example/lib/screens/flags_screen.dart
```diff
@@ -22,23 +23,31 @@ class FlagsScreen extends StatefulWidget {
...
       _status = 'loading';
     });
+    final client = _client;
+    if (client == null) {
+      _evaluateFlags();
+      setState(() {
+        _status = 'flags plugin not configured';
+      });
+      return;
+    }
```
fuzzybinary: You could probably check for the client above the initial `setState()` call to set the status to loading. It avoids an extra unneeded `setState`, but that likely isn't a big deal.

examples/simple_example/lib/main.dart
```diff
@@ -67,6 +67,11 @@ void main() async {
     siteName: dotenv.maybeGet('DD_SITE'),
     applicationId: datadogConfig.rumConfiguration?.applicationId,
   );
+  datadogConfig.addPlugin(
```
fuzzybinary: you can use chaining here:
```dart
)..addPlugin(
  DatadogFlagsPluginConfiguration(
    flagsConfiguration: flagsConfig.configuration,
  ),
);
```

packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
```diff
@@ -0,0 +1,322 @@
...
+  final Map<String, DatadogFlutterFlagsClient> _clients = {};
+
+  Future<void> _ready = Future<void>.value();
+
+  /// Creates a Flutter flags integration plugin.
+  DatadogFlagsPlugin(
+    super.instance, {
+    DatadogFlagsConfiguration flagsConfiguration =
+        const DatadogFlagsConfiguration(),
+    bool rumIntegrationEnabled = true,
```
fuzzybinary: Shouldn't these be required?

packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
```diff
@@ -0,0 +1,322 @@
...
+    return _clients.putIfAbsent(
+      name,
+      () => DatadogFlutterFlagsClient(
+        name: name,
+        resolveDelegate: () async {
+          await ready;
+          return _flags.sharedClient(name: name);
+        },
+        addRumFeatureFlagEvaluation: _addRumFeatureFlagEvaluation,
+        rumIntegrationEnabled: _rumIntegrationEnabled,
```
fuzzybinary: It might be easier to have `addRumFeatureFlagEvaluation` be nullable, and pass `null` if `_rumIntegrationEnabled` is false. That simplifies the calling sites as well as you can use optional chaining instead of checking `_rumIntegrationEnabled` first.

packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
fuzzybinary: I think we had it prior to making the split to have `datadog_flags` be pure dart without a depenency on the flutter package. I think this is the canonical place for it.

packages/datadog_flags_flutter/test/datadog_flags_plugin_test.dart
```diff
@@ -0,0 +1,352 @@
...
+
+  @override
+  Future<void> disable() async {
+    disableCalled = true;
+  }
+}
+
+class FakeDatadogFlagsClient implements DatadogFlagsClient {
+  @override
+  final String name;
```
fuzzybinary: Think about using mocktail for this. It *drastically* simplifies fakes and mockss.

packages/datadog_flags_flutter/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Getting Started
+
+Add the package:
+
+```bash
+flutter pub add datadog_flags_flutter
+```
+
+This package requires `datadog_flutter_plugin` 3.4.0 or later.
```
fuzzybinary: 3.6 is specified in the Pubspsec.

packages/datadog_flutter_plugin/lib/src/datadog_plugin.dart
fuzzybinary: Should be fine I think?

packages/datadog_flutter_plugin/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.4.0
+
+* Expose the active SDK configuration to Flutter plugin integrations.
+
```
fuzzybinary: This should hopefully be pulled from conventional commits... I actually don't know what releaser does if a version header already exists for the version it wants to update.....

Either way this should probably go under an `## Unreleased` header.

fuzzybinary (commented):
This mostly LGTM.

Two points that we might want to address in this PR, or can be a follow up PR, however you want to run it:

First, background isolate support.  It's very possible users will want to query flags from background isolates. An easy way to support this is to just have the clients re-query for the flags in the background isolate. A better way would be to try to transmit the already fetched flag during initialization and / or have it subscribed to changes from the main isolate.  Setting up the plugin to be notified when users initialize a background isolate, but actually supporting it is maybe a bit of a heavy lift for this PR

Second, if we're going to be chasing Pub points, we should add an example specifically to the package. If can be the same / similar as the one that's in `simple_example` if you want, or even simpler, just showing basic setup. But we do get pub points for it.

packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
```diff
@@ -309,7 +309,11 @@ class DatadogFlutterFlagsClient implements DatadogFlagsClient {
   }
 }
 
-DatadogFlagsSite? _flagsSiteFor(DatadogSite site) {
+/// Returns the Flags SDK site matching a Datadog Flutter SDK [site].
+///
+/// Some Datadog Flutter SDK sites, such as FedRAMP, do not currently have a
+/// matching Datadog Flags endpoint.
```
fuzzybinary, line 311: is FedRAMP the correct public name for this?

packages/datadog_flags_flutter/test/datadog_flags_plugin_test.dart
fuzzybinary, line 1: I don't know about you but I much prefer this approach with mocktail.

packages/datadog_flags_flutter/README.md
```diff
@@ -92,6 +93,13 @@ Use `package:datadog_flags/datadog_flags.dart` directly for pure Dart apps or
 Flutter apps that need full lifecycle control without Datadog Flutter SDK
 integration.
 
+## Background Isolates
+
+Flutter background isolates do not share the main isolate's plugin state. If
+your app evaluates flags from a background isolate, initialize the Datadog SDK
+and `datadog_flags_flutter` in that isolate before creating or using a flags
```
fuzzybinary: We should change this to stay that, as of right now, we don't support background isolates for this package. It's unfortunately not as simple as just re-initializing the Datadog SDK / Flags Plugin.

You can leave in the language that they can manually create the flags client as workaround, but we should discuss _how_ we want to actually support multiple isolates long term.

fuzzybinary (dismissed):
LGTM other than some minor docs stuff. Let's make sure the docs team reviews before a final merge.

packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
leoromanovsky: FedRAMP is the target of Datadog GovCloud. I'm fine with whatever naming is clearer to customers.

packages/datadog_flags_flutter/test/datadog_flags_plugin_test.dart
leoromanovsky: Agreed!

packages/datadog_flags_flutter/README.md
```diff
@@ -0,0 +1,109 @@
...
+Flutter apps that need full lifecycle control without Datadog Flutter SDK
+integration.
+
+## Background Isolates
+
+`datadog_flags_flutter` does not currently support evaluation from background
+isolates. If your app needs to evaluate flags from a background isolate, create
+and initialize a standalone `datadog_flags` client in that isolate.
+
+# Contributing
```
domalessi:
```suggestion
## Contributing
```
Heading level breaks parallel structure with sibling H2 sections (Getting Started, Custom Flags Configuration, Background Isolates).

packages/datadog_flags_flutter/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+Pull requests are welcome. For more information, read the
+[contributing guide](../../CONTRIBUTING.md) in the root repository.
+
+# License
```
domalessi:
```suggestion
## License
```
Same heading-level inconsistency as above.

packages/datadog_flags_flutter/README.md
```diff
@@ -0,0 +1,109 @@
...
+);
+```
+
+Use `package:datadog_flags/datadog_flags.dart` directly for pure Dart apps or
+Flutter apps that need full lifecycle control without Datadog Flutter SDK
+integration.
+
+## Background Isolates
+
+`datadog_flags_flutter` does not currently support evaluation from background
```
domalessi:
```suggestion
`datadog_flags_flutter` does not support evaluation from background
```
"currently" implies temporariness and isn't needed here.

examples/simple_example/README.md
```diff
@@ -23,9 +23,9 @@ names, org names, or customer-owned flag keys.
 
 ## Feature Flags
 
-The `Flags` screen initializes `DatadogFlags`, refreshes assignments for one
-evaluation context, and evaluates boolean, string, integer, double, and JSON
-flags with programmatic defaults.
+The `Flags` screen initializes the Flutter-integrated flags package, refreshes
```
domalessi:
```suggestion
The `Flags` screen initializes the `datadog_flags_flutter` package, refreshes
```
Naming the package explicitly is clearer than "the Flutter-integrated flags package."

domalessi (dismissed):
A few inline comments but approving so you're not blocked on me!

domalessi (approved):

fuzzybinary (approved):

sameerank (approved):

Issue #950, opened 2026-02-18
WebView tracking is broken in iOS in Flutter 3.41


fuzzybinary:
### Describe the bug

The current methods we use to access the `FlutterPluginRegistry` (using the `FlutterAppDelegate` or the `FlutterRootViewController` attached to that delegate) no longer work given Flutter's transition to using `UIScene`.

Accessing the registry is necessary to get the WebViews that it creates to install our tracking. 

We're still looking into a potential solution to either access the registry, or the published plugin value. 

### Reproduction steps

Install `datadog_webview_tracking`.
Attempt to track the webview.

### Affected SDK versions

3.0.1

### Latest working SDK version

3.0.1

### Did you confirm if the latest SDK version fixes the bug?

Yes

### Flutter Version

3.41.0

fuzzybinary:
Part of the fix is now merged as part of https://github.com/flutter/flutter/pull/182753. I still need to have `webview_flutter` utilize this method for getting published values, but I'll work on that PR shortly.

As it stands, `datadog_webview_tracking` will unfortunately be broken for the full 3.41 release cycle, but should be fixed with 3.44.

ecolab-neil:
@fuzzybinary  this is exciting, can you keep us up to date at your convenience in the likely event that flutter 3.44 is released to stable on Wednesday

fuzzybinary:
This is now fixed in `datadog_webview_tracking 3.1.0` which requires Flutter 3.44.

cc @ecolab-neil

fuzzybinary:
Secondary note - this will unfortunately remain broken in Flutter 3.41, because the fix requires Flutter 3.44. If you are using `datadog_webview_plugin` you will need to use either Flutter < 3.41 with 3.0.x or Flutter > 3.44 with 3.1.0.

PR #1080, opened 2026-07-06
feat!: Update all packages to Dart 3.10 and Flutter 3.38

fuzzybinary:
### What and why?

We're going to set our minimum versions in v4 to Dart 3.10 and Flutter 3.38. This allows us to upgrade to the latest Objective-C, JNI, and FFI libraries.

In addition to this, we're trying to get as close to full SPM, AGP 9, and Built-In kotlin support as possible, though some dependent libraries still do not support built in Kotlin (such as `jni`) or `spm` (such as `flutter_inappwebview`).

All plugins should now build cleanly without too many warnings.

### Review checklist


leoromanovsky (dismissed):
Do you need any help re-verifying manually?

fuzzybinary:
@leoromanovsky I think I'm good. Going to wait on @kikoveiga as a second reviewer before merge but hopefully this is the last super large PR for the v4 conversion and upgrade.

packages/datadog_webview_tracking/pubspec.yaml
```diff
@@ -5,21 +5,21 @@ homepage: http://datadoghq.com
 repository: https://github.com/DataDog/dd-sdk-flutter
 
 environment:
-  sdk: ">=3.8.0 <4.0.0"
+  sdk: ">=3.10.0 <4.0.0"
   flutter: ">=3.44.0"
```
kikoveiga, line 9: This was already above the `3.38` baseline and mismatched with `sdk: ">=3,10,0"`, is it intended?

kikoveiga (commented):
- `packages/datadog_flags_flutter/pubspec.yaml:14` is still on `sdk: ">=3.6.0"` and `flutter: >="3.27.0"`, should we bump them?
- `packages/datadog_flutter_plugin/datadog_flutter_plugin_ios/example/pubspec.yaml:8` has `sdk: ^3.12.0` vs `3.10` everywhere else, is it inteded?

packages/datadog_webview_tracking/pubspec.yaml
fuzzybinary: This one should definitely be fixed... not sure why it's at 3.44+.

Good catch!

fuzzybinary:
Good catches on all the mismatches. I'll get them all fixed.  Thanks!

packages/datadog_webview_tracking/pubspec.yaml
fuzzybinary: Okay, looking into it, this is actually "correct". This newest version actually requires something that's only available in Flutter 3.44+, and fixes a break in Flutter 3.41 (see #950). I may need to supply a patch version of 3.1 that supports `datadog_flutter_plugin` up to Flutter 3.38.

Dealing with this type of thing is now tracked as RUMM-3555.

kikoveiga: Alright, thanks for taking a look!

kikoveiga (dismissed):

PR #1091, opened 2026-07-24
RUM-17728 Add Performance Timeseries configuration to RUM

barboraplasovska:
### What and why?

Exposes the native Timeseries RUM feature (dense memory/CPU sampling events) through the public RUM configuration API.

> [!WARNING]
> This is currently blocked from compiling: the pinned native SDK versions don't yet contain Timeseries. Do not open this for real review until native work is merged and released.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
@@ -64,6 +64,33 @@ typedef RumLongTaskEventMapper = RumLongTaskEvent? Function(
 typedef RumVitalOperationEventMapper = RumVitalOperationStepEvent? Function(
     RumVitalOperationStepEvent event);
 
+/// Configuration for the Performance Timeseries RUM feature, which samples
+/// memory and CPU usage at regular intervals throughout a session,
+/// independent of view lifecycle.
+class DdTimeseriesConfiguration {
+  /// Whether to enable collection of memory and CPU timeseries events.
+  ///
+  /// Defaults to `false`.
+  final bool enabled;
+
+  /// The number of samples collected before a timeseries batch is flushed.
+  ///
+  /// If not set, the native SDK default is used.
+  final int? bufferSize;
+
+  const DdTimeseriesConfiguration({
+    required this.enabled,
+    this.bufferSize,
+  });
+
+  Map<String, Object?> encode() {
+    return {
+      'enabled': enabled,
+      if (bufferSize != null) 'bufferSize': bufferSize,
+    };
+  }
+}
+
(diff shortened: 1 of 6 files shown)


packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
```diff
@@ -64,6 +64,41 @@ typedef RumLongTaskEventMapper = RumLongTaskEvent? Function(
 typedef RumVitalOperationEventMapper = RumVitalOperationStepEvent? Function(
     RumVitalOperationStepEvent event);
 
+/// A device timeseries type that can be collected by the Performance
+/// Timeseries RUM feature.
+///
+/// *Note*: This API is experimental and may change in the future.
+enum DdTimeseriesType {
```
fuzzybinary, line 71: We've dropped the `Dd` prefix on most objects, I think we can safely drop it from here and from the configuration below.

Also, I guess I'm late to the party on this but `TimeSeries` should really have been two words 😅

packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
```diff
@@ -64,6 +64,41 @@ typedef RumLongTaskEventMapper = RumLongTaskEvent? Function(
...
+  final List<DdTimeseriesType>? collectTypes;
+
+  const DdTimeseriesConfiguration({
+    this.collectTypes,
+  });
+
+  Map<String, Object?> encode() {
+    return {
+      if (collectTypes != null)
+        'collectTypes': collectTypes!.map((e) => e.toString()).toList(),
```
fuzzybinary, line 97: If we switch this to v4, we can use null aware collection elements here (introduced in Dart 3.8:
```dart
return {
  'collectTypes': ?collectTypes?.map((e) => e.toString()).toList(),
};
```

packages/datadog_flutter_plugin/lib/src/rum/rum_configuration.dart
```diff
@@ -298,6 +341,7 @@ class DatadogRumConfiguration {
       'attachVitalOperationStepEventMapper':
           vitalOperationStepEventMapper != null,
       'additionalConfig': additionalConfig,
+      if (timeseries != null) 'timeseries': timeseries!.encode(),
```
fuzzybinary, line 344: null aware collection elements would work here as well, but for simplicity it's likely also okay to pass `null` here and run the null check in the native code as well.

fuzzybinary (changes requested):
Very minor change to remove the Dd prefix. Otherwise looks good.

Let's have a discussion about whether this should target v3 or v4. Given the new requirement in `dd-sdk-ios` to target iOS 15, we may have to target v4 with this change.

Issue #942, opened 2026-02-09
[Android] SIGSEGV null pointer dereference in globalEnv_CallVoidMethod (libdartjni.so)


dfdgsdfg:
## Description

App crashes with a native SIGSEGV (null pointer dereference) on Android in release mode. The crash originates from `globalEnv_CallVoidMethod` in `libdartjni.so`, which calls `CallVoidMethodV` with a null `_jobject*`.

We suspect this is related to **Session Replay** capturing screen state, as the Dart VM frames show a repeated iterative pattern consistent with widget tree traversal.

## Environment

- **Flutter:** 3.38.9 (stable)
- **Dart:** 3.10.8
- **datadog_flutter_plugin:** 3.0.1
- **datadog_session_replay:** 1.0.0-preview.9
- **datadog_tracking_http_client:** 3.0.1
- **datadog_dio:** 2.0.0
- **datadog_inappwebview_tracking:** 2.0.0
- **jni (transitive):** 0.14.2
- **Android NDK:** 28.2.13676358
- **R8/ProGuard:** Disabled (`minifyEnabled` is false)
- **Device arch:** arm64-v8a

## Setup

Session Replay is enabled and wraps the entire app widget tree via `SessionReplayCapture`:

```dart
DatadogConfiguration(
  // ...
  nativeCrashReportEnabled: true,
  rumConfiguration: DatadogRumConfiguration(applicationId: applicationId),
  site: DatadogSite.us5,
)..enableSessionReplay(
  DatadogSessionReplayConfiguration(
    textAndInputPrivacyLevel: TextAndInputPrivacyLevel.maskSensitiveInputs,
    touchPrivacyLevel: TouchPrivacyLevel.show,
    replaySampleRate: replaySampleRate,
  ),
);
```

Debug mode already skips Datadog initialization entirely (to avoid the hot-restart JNI crash from #932).

## Tombstone (symbolicated)

```
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
Cause: null pointer dereference

#00 pc 0x3150bc  art::JValue art::InvokeVirtualOrInterfaceWithVarArgs<_jmethodID*>(
                   art::ScopedObjectAccessAlreadyRunnable const&, _jobject*, _jmethodID*, std::__va_list)
                   [libart.so]
#01 pc 0x620a84  art::JNI<false>::CallVoidMethodV(_JNIEnv*, _jobject*, _jmethodID*, std::__va_list)
[...]

PR #1093, opened 2026-07-24
fix(sr): Fix dangling bridge reference in JNI

fuzzybinary:
### What and why?

The FlutterSessionReplayBridge reference held by the processing isolate could become stale if the root isolate was killed or released its reference. Avoid this using JNI's `as` over a Dart cast to create a new reference on the processing isolate side.

Additionally, make sure the background isolates are shutdown properly, instead of forcing an immediate kill during app lifecycle shutdown.

refs: #942

### Review checklist

### DIFF
--- a/packages/datadog_session_replay/lib/src/processor/processor.dart
+++ b/packages/datadog_session_replay/lib/src/processor/processor.dart
@@ -2,6 +2,7 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2025-Present Datadog, Inc.
 
+import 'dart:async';
 import 'dart:isolate';
 
 import 'package:flutter/services.dart';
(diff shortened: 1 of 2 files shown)


packages/datadog_session_replay/lib/src/processor/processor.dart
```diff
@@ -47,8 +56,29 @@ class SessionReplayProcessor with WidgetsBindingObserver {
...
+      unawaited(_shutdown());
+    }
+  }
+
+  /// Asks the processor isolate to stop, then waits for it to finish any
+  /// in-flight snapshot and exit on its own. The isolate is only force
+  /// killed as a fallback if it doesn't exit within [_shutdownTimeout] -
+  /// killing it while it may still be mid-call into native JNI/ObjC code
+  /// can leave those references in a bad state.
+  Future<void> _shutdown() async {
```
kikoveiga, line 68: Should we add a unit test for this function? Ideally testing the 2 paths: graceful ack and timeout fallback.

kikoveiga (approved):

packages/datadog_session_replay/lib/src/processor/processor.dart
fuzzybinary: It's a good callout, but it's also not a simple unit test because of interaction with the background isolate and native functions. I'll see if I can figure out a way to write one though.

PR #1094, opened 2026-07-24
feat(sr): Cache sent resource identifiers via DataStore to avoid re-upload (RUM-11447)

JuanNaranjoDD:
### What and why?

Session Replay must not resend images whose MD5 hash was already sent to the resource endpoint in a previous session. Both platforms mirror their own native SDK's dedup design rather than each other, so their marking behavior differs.

Android introduces a dedicated ResourceDataStoreManager, mirroring native dd-sdk-android's class of the same name: it loads its known-hash entry eagerly on construction and marks an identifier as known optimistically, before the write is attempted, gated on isReady() so it never overwrites the DataStore entry before the initial load completes. This required constructing the writer inside onInitialize, after ResourceFeature registers, instead of as a property initializer, so the eager load has a feature to read from.

iOS keeps its existing design, mirroring native iOS: it seeds its identifier cache eagerly in init (its FeatureScope is already registered at that point) and marks an identifier as known only after its write actually succeeds.

Both platforms reset their cache after 30 days to bound unbounded local growth.

### How?

- Android: Added ResourceDataStoreManager (resource/ResourceDataStoreManager.kt), which wraps the feature's DataStoreHandler behind a single combined entry (DATASTORE_HASHES_ENTRY_KEY) storing a JSON blob of {last_update_date_ns, resource_hashes}. On construction it reads that entry via sdkCore.timeProvider-based elapsed time, discards it and clears the DataStore if it's older than 30 days, otherwise seeds an in-memory known-hash set from it. isReady() reports whether the initial read has completed (including failure paths), and cacheResourceHash() persists the updated set back to the DataStore. DefaultResourceWriter.write() now checks isPreviouslySentResource() first, marks the hash via cacheResourceHash() before writing (only if isReady(), to avoid clobbering data that hasn't loaded yet), then writes the resource event as before. Since the manager's eager load requires ResourceFeature to already be registered, DefaultResourceWriter/ResourceDataStoreManager construction moved from a property initializer into FlutterSessionReplayFeature.onInitialize(), after sdkCore.registerFeature(resourcesFeature).
- iOS: ResourcesWriter is unchanged in structure — it already seeded knownIdentifiers eagerly from scope.dataStore in init and reset the store after 30 days. Minor internal cleanup/refactor plus expanded test coverage (ResourcesWriterTest.swift, new DataStoreMock.swift) to exercise the seeding, reset, and mark-after-write paths more thoroughly.
- Added ResourceDataStoreManagerTest.kt and ResourceWriterTest.kt (mockk/JUnit5, matching this repo's existing Android test conventions) covering: fresh vs. previously-cached resources, DataStore persistence, stale-store reset after 30 days, and the various isReady()-gated write/cache orderings, including failure paths on read and delete.

### Review checklist

- [x] This pull request references a Github or JIRA issue — 11447

### DIFF
--- a/packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
+++ b/packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
@@ -6,6 +6,7 @@
 
 package com.datadoghq.flutter.sessionreplay.resource
 
+import com.datadog.android.api.InternalLogger
 import com.datadog.android.api.feature.FeatureSdkCore
 import com.datadog.android.api.storage.datastore.DataStoreReadCallback
 import com.datadog.android.api.storage.datastore.DataStoreWriteCallback
@@ -23,11 +24,11 @@ import java.util.concurrent.atomic.AtomicLong
 internal class ResourceDataStoreManager(
     private val featureSdkCore: FeatureSdkCore,
     private val gson: Gson = Gson(),
-    private val dataStoreResetTimeNs: Long = TimeUnit.DAYS.toNanos(30)
+    private val dataStoreResetTimeMs: Long = TimeUnit.DAYS.toMillis(30)
 ) {
     @Suppress("UnsafeThirdPartyFunctionCall") // map is initialized empty
     private val knownResources = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())
-    private val storedLastUpdateDateNs = AtomicLong(featureSdkCore.timeProvider.getDeviceElapsedTimeNanos())
+    private val storedLastUpdateDateMs = AtomicLong(featureSdkCore.timeProvider.getDeviceTimestampMillis())
     private val isInitialized = AtomicBoolean(false) // has init finished executing its async actions
 
     init {
@@ -40,10 +41,10 @@ internal class ResourceDataStoreManager(
                     return@lambda
                 }
 
-                val lastUpdateDateNs = storedData.lastUpdateDateNs
+                val lastUpdateDateMs = storedData.lastUpdateDateMs
                 val storedHashes = storedData.resourceHashes
 
-                if (didDataStoreExpire(lastUpdateDateNs)) {
+                if (didDataStoreExpire(lastUpdateDateMs)) {
                     deleteStoredHashesEntry(
                         callback = object : DataStoreWriteCallback {
                             override fun onSuccess() {
@@ -51,17 +52,27 @@ internal class ResourceDataStoreManager(
                             }
 
                             override fun onFailure() {
+                                featureSdkCore.internalLogger.log(
+                                    InternalLogger.Level.ERROR,
+                                    InternalLogger.Target.TELEMETRY,
+                                    { Constants.FAILED_TO_DELETE_STALE_ENTRY_ERROR_MESSAGE }
+                                )
                                 finishedInitializingManager()
                             }
                         }
                     )
                 } else {
-                    storedLastUpdateDateNs.set(lastUpdateDateNs)
+                    storedLastUpdateDateMs.set(lastUpdateDateMs)
                     knownResources.addAll(storedHashes)
                     finishedInitializingManager()
                 }
             },
             onFetchFailure = {
+                featureSdkCore.internalLogger.log(
+                    InternalLogger.Level.ERROR,
+                    InternalLogger.Target.TELEMETRY,
+                    { Constants.FAILED_TO_FETCH_ENTRY_ERROR_MESSAGE }
+                )
                 finishedInitializingManager()
             }
         )
@@ -86,7 +97,7 @@ internal class ResourceDataStoreManager(
 
     private fun writeResourcesToStore() {
         val data = ResourceHashesEntry(
-            lastUpdateDateNs = storedLastUpdateDateNs.get(),
+            lastUpdateDateMs = storedLastUpdateDateMs.get(),
             resourceHashes = knownResources.toList()
         )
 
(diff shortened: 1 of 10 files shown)


packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,152 @@
...
+    private val isInitialized = AtomicBoolean(false) // has init finished executing its async actions
+
+    init {
+        fetchStoredResourceHashes(
+            onFetchSuccessful = lambda@{ storedEntry ->
+                val storedData = storedEntry?.data
+
+                if (storedData == null) {
+                    finishedInitializingManager()
+                    return@lambda
```
fuzzybinary, line 41: Is the `@lambda` here necessary?

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,152 @@
...
+import java.util.concurrent.atomic.AtomicLong
+
+internal class ResourceDataStoreManager(
+    private val featureSdkCore: FeatureSdkCore,
+    private val gson: Gson = Gson(),
+    private val dataStoreResetTimeNs: Long = TimeUnit.DAYS.toNanos(30)
+) {
+    @Suppress("UnsafeThirdPartyFunctionCall") // map is initialized empty
+    private val knownResources = Collections.newSetFromMap(ConcurrentHashMap<String, Boolean>())
+    private val storedLastUpdateDateNs = AtomicLong(featureSdkCore.timeProvider.getDeviceElapsedTimeNanos())
```
fuzzybinary: This shouldn't be using `getDeviceElapsedTimeNanos`, which is using time since device boot. It's a high resolution timer meant to measure elapsed time, not absolute time.

It is likely better to use the system clock and milliseconds or seconds since epoch. This doesn't need to be in nanosecond precision as we're checking for a difference in days.

packages/datadog_session_replay/android/src/test/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManagerTest.kt
```diff
@@ -0,0 +1,289 @@
...
+        every { mockScope.dataStore } returns fakeDataStore
+        every { mockSdkCore.timeProvider } returns FakeTimeProvider { fakeElapsedTimeNs }
+    }
+
+    @Test
+    fun `M return false W isPreviouslySentResource {resource was not sent}`(
+        @StringForgery hash: String
+    ) {
+        // Given
+        setUp()
```
fuzzybinary: JUnit supports `@BeforeEach`. You shouldn't have to call `setUp` at the beginning of every test.

packages/datadog_session_replay/android/src/test/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceWriterTest.kt
```diff
@@ -0,0 +1,124 @@
...
+import org.junit.jupiter.api.extension.ExtendWith
+
+@ExtendWith(ForgeExtension::class)
+internal class ResourceWriterTest {
+    val mockSdkCore = mockk<FeatureSdkCore>()
+    val mockScope = mockk<FeatureScope>()
+    val mockDataStoreManager = mockk<ResourceDataStoreManager>()
+
+    fun setUp() {
+        every {
```
fuzzybinary, line 33: `@BeforeEach` here too.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,152 @@
...
+                        }
+                    )
+                } else {
+                    storedLastUpdateDateNs.set(lastUpdateDateNs)
+                    knownResources.addAll(storedHashes)
+                    finishedInitializingManager()
+                }
+            },
+            onFetchFailure = {
+                finishedInitializingManager()
```
fuzzybinary, line 76: This should probably be sent to telemetry if this happens.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
```diff
@@ -0,0 +1,152 @@
...
+
+                if (didDataStoreExpire(lastUpdateDateNs)) {
+                    deleteStoredHashesEntry(
+                        callback = object : DataStoreWriteCallback {
+                            override fun onSuccess() {
+                                finishedInitializingManager()
+                            }
+
+                            override fun onFailure() {
+                                finishedInitializingManager()
```
fuzzybinary, line 60: Failure here should probably be sent to telemetry.

fuzzybinary (changes requested):
A few minor changes. Specifically, `getDeviceElapsedTimeNanos` is not the right time measure here.

packages/datadog_session_replay/android/src/main/kotlin/com/datadoghq/flutter/sessionreplay/resource/ResourceDataStoreManager.kt
JuanNaranjoDD: Same pattern used in the Android SDK, but I also dug into why it's actually needed: fetchStoredResourceHashes isn't an inline function, so the lambda is a real closure. Kotlin only generates an implicit label automatically when a lambda is passed to an inline function — that's not the case here. So I need an explicit label to do a local early-return, hence lambda@.

fuzzybinary (approved):

PR #1095, opened 2026-07-28
feat: Add support for Windows CI

fuzzybinary:
### What and why?

> [!NOTE]
> The .gitlab job for windows integration testing has been commented out, because our current CI runners cannot create a graphics stack to actually run them. Other changes will remain intact because they are useful for linux and macOS integration tests.

Add support for windows CI for unit and integration testing.

Integration tests on Desktop must be run one file at a time because of a bug in the Flutter tool, so a script has been introduced to do that

Additionally, since many of the melos steps need to be executed cross platform, we've had to abstract some of them into Dart scripts. The two major examples of this are:

  * `run_with_junit` which pipes flutter test output to `tojunit`, but preserves the exit / status code of `flutter test` command. This is used because Windows pipes will only return the status code of the last element in the pipe, which doesn't preserve the pass / fail state of the tests
  * `generate_env` since old `.env` file generation used `tee`, which is not available on Windows, we've created a custom script that generates the `.env` files properly cross platform.

refs: RUM-17631

### Review checklist


fuzzybinary:
@codex review

ci/windows/ci-pre.ps1
```diff
@@ -0,0 +1,34 @@
+# PowerShell equivalent of the `.pre` bash template in .gitlab-ci.yml, run inside the
+# Windows CI toolchain container (see ci/windows/Dockerfile). Unlike bash, PowerShell
+# does not stop the script when a native executable (flutter/dart/melos) exits non-zero,
+# so each call is followed by an explicit $LASTEXITCODE check.
+$ErrorActionPreference = "Stop"
+
+flutter upgrade --force
+if ($LASTEXITCODE -ne 0) { throw "flutter upgrade failed with exit code $LASTEXITCODE" }
```
awforsythe, line 8: Apparently, as of PowerShell 7.1 (released Nov. 2022), you can omit these per-command exit code checks (and get `set -e` behavior) by setting `$PSNativeCommandUseErrorActionPreference = $true`.

This is fine as-is though; I don't know what PowerShell version(s) we run, and the throw-per-command style is still typical.

awforsythe (approved):

ci/windows/ci-pre.ps1
fuzzybinary: TIL. We can check what version another day and  maybe remove these from multiple CI scripts in the future.

PR #1099, opened 2026-08-03
feat(rum): use tooltips as action descriptions in RumUserActionDetector

### LINKED ISSUE #1033: [Feature request] Support using `tooltip` as a description in RumUserActionDetector
### Feature description

Plese support tooltip including icons such as hello world "Increment" button or other tooltips, this is common such as a camera icon to take a picture, share icon etc.

if there is workaround please let me know and document if possible

### Proposed solution

Support tooltip

### Other relevant information

Hello world uses fab and tooltip so it would help evaluation and "first hour" experience

brunovsiqueira:
### What and why?

Closes #1033.

`RumUserActionDetector` builds action descriptions from `RumUserActionAnnotation`, child `Text`, `Semantics` labels, and `Icon.semanticsLabel`, but ignores tooltips. As a result, common buttons that only carry a `tooltip` — like the stock counter app's `FloatingActionButton(tooltip: 'Increment')` or an `IconButton(tooltip: ...)` — report `InkWell(unknown)` / `IconButton(unknown)`.

This PR uses tooltip messages as action descriptions, and recognizes `FloatingActionButton` as its own element type. With it, the `flutter create` counter app reports `FloatingActionButton(Increment)` out of the box.

### How?

The description priority is now: `RumUserActionAnnotation` → the tapped widget's own `tooltip` property (`IconButton`, `FloatingActionButton`) → child `Text` / `Semantics` label / `Icon.semanticsLabel` (unchanged relative order) → the message of a `Tooltip` enclosing the tapped widget on the hit-test path → `unknown`.

Per review feedback, the widget's own tooltip ranks directly below `RumUserActionAnnotation`: a button's tooltip describes the action being performed, while child content like an icon's semantic label describes the visuals. This also means an extended FAB with both a `label` and a `tooltip` reports the tooltip.

Two other deliberate design points:

* Tooltips are only read from the tapped widget's own `tooltip` property or from `Tooltip` widgets that *enclose* the tap position — never from an unbounded subtree search. This prevents a tap on a large `InkWell`/`GestureDetector` container from being misattributed to an embedded control's tooltip (e.g. a list row reporting `InkWell(Delete)` because of a trailing delete button, or `PopupMenuButton`'s default "Show menu" tooltip), and keeps the existing `searchForText: false` privacy behavior of generic containers intact. A regression test pins this.
* Hit-path capture of `RumUserActionAnnotation`/`Tooltip` is now scoped with save/restore instead of being cleared after the first completed subtree, so a value applies to *all* branches of its subtree even when the hit-test path crosses sibling render branches (e.g. a translucent tap layer inside a `Stack`). This also fixes a latent edge case for `RumUserActionAnnotation`, covered by new tests.

The `FloatingActionButton` recognition is a separate commit (it changes the reported element type for existing FAB actions) so it can be dropped independently if undesired.

`Tooltip.richMessage` is supported via `toPlainText()`, empty messages are ignored, and both Material 2 and Material 3 `IconButton` paths are covered. 22 new widget tests; `flutter analyze` and the full package unit suite pass.

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/CHANGELOG.md
+++ b/packages/datadog_flutter_plugin/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Use a widget's `Tooltip` message as a fallback description for actions detected by `RumUserActionDetector`. See [#1033](https://github.com/DataDog/dd-sdk-flutter/issues/1033).
+
 ## 3.4.1
 
 * Properly pin Android SDK version to 3.11.0.
--- a/packages/datadog_flutter_plugin/CHANGELOG.md
+++ b/packages/datadog_flutter_plugin/CHANGELOG.md
@@ -1,10 +1,5 @@
 # Changelog
 
-## Unreleased
-
-* Use a widget's `Tooltip` message as a fallback description for actions detected by `RumUserActionDetector`. See [#1033](https://github.com/DataDog/dd-sdk-flutter/issues/1033).
-* Report `FloatingActionButton` as its own element type in `RumUserActionDetector`, using its label text or tooltip as the description. Actions on these buttons were previously reported as their inner `InkWell`, so both the element type and description of existing actions may change.
-
 ## 3.4.1
 
 * Properly pin Android SDK version to 3.11.0.
--- a/packages/datadog_flutter_plugin/CHANGELOG.md
+++ b/packages/datadog_flutter_plugin/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Use a widget's `Tooltip` message as a fallback description for actions detected by `RumUserActionDetector`. See [#1033](https://github.com/DataDog/dd-sdk-flutter/issues/1033).
+* Report `FloatingActionButton` as its own element type in `RumUserActionDetector`, using its label text or tooltip as the description. Actions on these buttons were previously reported as their inner `InkWell`, so both the element type and description of existing actions may change.
 
 ## 3.4.1
 
--- a/packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
+++ b/packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
@@ -504,6 +504,395 @@ void main() {
     verifyNoMoreInteractions(mockRum);
   });
 
+  testWidgets('tap IconButton with tooltip reports tooltip message',
+      (tester) async {
+    final mockRum = MockDdRum();
+
+    final tooltip = randomString();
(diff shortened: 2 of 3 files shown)


fuzzybinary:
@codex review

packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
```diff
@@ -504,6 +504,482 @@ void main() {
...
+
+    final button = find.byType(IconButton);
+    await tester.tap(button);
+
+    verify(() => mockRum.addAction(RumActionType.tap, 'IconButton($tooltip)'));
+    verifyNoMoreInteractions(mockRum);
+  });
+
+  testWidgets(
+      'tap IconButton with semantic label and tooltip prefers semantic label',
```
fuzzybinary: Do we think this is correct behavior? What do you think you'd rather see, the semantic label on the icon or the tooltip of the button? I personally feel like the tooltip on the button would be more beneficial?

packages/datadog_flutter_plugin/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+* Use a widget's `Tooltip` message as a fallback description for actions detected by `RumUserActionDetector`. See [#1033](https://github.com/DataDog/dd-sdk-flutter/issues/1033).
+* Report `FloatingActionButton` as its own element type in `RumUserActionDetector`, using its label text or tooltip as the description. Actions on these buttons were previously reported as their inner `InkWell`, so both the element type and description of existing actions may change.
+
```
fuzzybinary: Don't add changes to the CHANGELOG manually. These are done by the conventional commits in our release process.

packages/datadog_flutter_plugin/test/rum/rum_user_action_detector_test.dart
brunovsiqueira: This is a really good point. I flipped it.  The chain now is: user annotation -> tooltip -> semantics label. https://github.com/DataDog/dd-sdk-flutter/pull/1099/commits/6b65d0349e29f723735c3629bab684aabe544833

packages/datadog_flutter_plugin/CHANGELOG.md
brunovsiqueira: done https://github.com/DataDog/dd-sdk-flutter/pull/1099/commits/1edfac564b4059e29968f57a491332735e493d53

fuzzybinary (commented):
This LGTM. Thanks for the contribution!

fuzzybinary (approved):

PR #1100, opened 2026-08-03
feat(sr): support Session Replay in hybrid iOS apps via per-engine slot routing

JuanNaranjoDD:
### What and why?

Session Replay assumed Flutter owned the whole screen, so records from an embedded engine had nowhere to go. Add a slot-based path that lets the native player composite Flutter records into the host's embedded_view placeholder.

### How?

Each FlutterViewController opts in by calling enableDatadogSessionReplay(), which registers the controller against its engine's binary messenger and mints a slot ID on its view via the embedding SDK's view.dd.sessionReplaySlotID. The native recorder picks that view up as an embedded_view placeholder wireframe; the Flutter side stamps its own records with the same slot ID, so the player knows where to draw them.

Routing is per-engine rather than per-process. Slot lookup goes through the engine's messenger, so several engines can record at once without stealing each other's context — the Dart side re-resolves its slot on didChangeMetrics and pushes it down via setSlotId, which also flushes any segments buffered before the embedding state was known.

Records then leave through the native message bus as FeatureMessage.embeddedContent instead of being written to a standalone SR segment. When no slot is present the plugin stays in standalone mode and behaves exactly as before, so pure-Flutter apps are unaffected.

The embedded_view placeholder, the slot ID association, and the embeddedContent message are Guille's work in dd-sdk-ios — this PR is the Flutter-side consumer of that API. datadog_session_replay's pubspec now requires datadog_flutter_plugin: ^3.5.0 so pub keeps the two in step and the core package can't ship without those changes.

### Review checklist


packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/DatadogSessionReplayPlugin.swift
```diff
@@ -29,18 +32,15 @@ public class DatadogSessionReplayPlugin: NSObject, FlutterPlugin {
...
             result(FlutterMethodNotImplemented)
         }
     }
-
-    public func detachFromEngine(for registrar: FlutterPluginRegistrar) {
-        // Null out the context callback only if this engine is the registered owner.
-        // Prevents a detaching secondary engine from clearing a live engine's callback,
-        // which would cause DLRT_GetFfiCallbackMetadata crashes on the next context update.
-        FlutterSessionReplay.detachFromEngine(messenger: registrar.messenger() as AnyObject)
-    }
```
fuzzybinary, line 45: Some version of this is actuallly probably necessary cleanup to prevent the crash on force close.

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/FlutterSessionReplayBridge.swift
```diff
@@ -43,19 +52,48 @@ public func __datadog_session_replay_keep_symbols() {
...
-    // Static properties so the callback and feature survive engine detach/re-attach cycles,
-    // mirroring the Android FlutterSessionReplayBridge singleton pattern.
-    internal static var contextCallback: ((FlutterRUMCoreContext?) -> Void)?
-    internal static var feature: FlutterSessionReplayFeature?
+    /// The coordinator shared with every other engine's bridge.
+    internal let manager: FlutterSessionReplayManager
+
+    /// Creates a bridge backed by the process-wide coordinator. This is the initializer Dart
+    /// calls over FFI, so it must stay `@objc` and argument-free.
+    @objc public override convenience init() {
```
fuzzybinary: The comment here is a bit misleading. We technically could change the Dart to use a different `alloc / init` pattern from Objective-C if there was a need, but we're choosing to leave it this way because using `FlutterSessionReplayManager.shared` makes sense as a sensible default and simplifies the binding.

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/FlutterSessionReplayBridge.swift
```diff
@@ -43,19 +52,48 @@ public func __datadog_session_replay_keep_symbols() {
...
+        self.init(manager: .shared)
+    }
+
+    /// Creates a bridge backed by `manager`, so tests can substitute the coordinator.
```
fuzzybinary: This comment has a very hight chance of being eventually wrong if we ever do decide, for some reason, to split out managers.

Better to put something like it's "primarily used in tests".

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/FlutterSessionReplayManager.swift
```diff
@@ -0,0 +1,249 @@
...
+    /// The shared feature, registered on the first engine to enable.
+    internal private(set) var feature: FlutterSessionReplayFeature?
+
+    /// Retained so embedded engines can post records to the native SR message bus.
+    private var core: DatadogCoreProtocol?
+
+    /// Weak set of every live engine bridge. The single `RUMContextReceiver` fans out to all
+    /// of them (see `broadcastContext`), so every engine — not just the last to enable —
+    /// receives live RUM context updates. Entries clear automatically when an engine's
+    /// bridge is released.
```
fuzzybinary, line 39: Just want to make sure this was tested and confirmed?

packages/datadog_session_replay/lib/src/datadog_session_replay.dart
```diff
@@ -82,6 +82,15 @@ class DatadogSessionReplay {
...
     _recorder.onContextChanged(context);
+
+    // Force the next capture tick to run. A RUM context change (notably a view
+    // change) requires a fresh snapshot for the new view, but captures are normally
+    // gated on `_newFrameBuilt`, which is only set when Flutter builds a frame. In
+    // hybrid add-to-app the native RUM view can change while an embedded Flutter panel
+    // is static (no new frame), so without this the new view's snapshot is deferred
+    // until the next interaction. `onContextChanged` only fires on actual context
+    // changes, so this doesn't add per-frame work.
+    // _newFrameBuilt = true;                       // This might not be necessary. Let's wait until the player ships to see whether we need to add it.
```
fuzzybinary: Let's remove this comment and just track it with a JIRA ticket if we can. I would rather the comment not accidentally stick around.

fuzzybinary (changes requested):
A few minor changes to address.

Do we want to make sure the hybrid sample application has Session Replay? It might also benefit from a slight redesign that actually looks like an actual app if you're up for it.

fuzzybinary (commented):
This LGTM, but all Unit tests are failing. Can we look into that and I'll re-review when that's updated?

fuzzybinary (approved):

PR #1105, opened 2026-08-04
feat!: Require a `service` name when configuring Datadog

fuzzybinary:
### What and why?

Service is required for Datadog, but we attempt to find a sensible default for most SDKs. Unfortunately, Browser does not really have an available "sensible default", and the default we're using on Desktop is the name of the executable, which isn't ideal.

To prevent any confusing issues, v4 is now going to require a service be set instead of trying to guess at a sensible value.

refs: RUM-17896

### Review checklist

### DIFF
--- a/examples/simple_example/lib/main.dart
+++ b/examples/simple_example/lib/main.dart
@@ -43,6 +43,7 @@ void main() async {
           clientToken: clientToken,
           env: env,
           site: siteConfig.datadogSite,
+          service: 'com.datadoghq.example.flutter',
           loggingConfiguration: DatadogLoggingConfiguration(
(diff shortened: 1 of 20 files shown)


rtrieu (dismissed):

kikoveiga (dismissed):
Claude noticed the `service` is missing in `packages/datadog_flags_flutter/README.md` code snippets, could you add it there? Might be similar for other `README` files.

fuzzybinary:
> Claude noticed the `service` is missing in `packages/datadog_flags_flutter/README.md` code snippets, could you add it there? Might be similar for other `README` files.

Yes good call Claude / @kikoveiga. I'll update them.

rtrieu (approved):

PR #1108, opened 2026-08-07
fix(flags): avoid Flutter Web intake preflights

leoromanovsky:
## Motivation

Flutter Web sends Flags intake requests with custom Datadog headers. Browsers send an OPTIONS preflight, and intake returns HTTP 403. Evaluation and exposure telemetry then fail.

## Changes and Decisions

This change matches the browser intake transport. Web requests use text/plain and put public intake metadata in query parameters. Evaluation uploads use newline-delimited JSON with Datadog context on each event. Native request headers and payloads remain unchanged. A Chrome integration test records cross-origin requests and rejects any OPTIONS request.

## Validation

Ran the new test in Chrome against the local recording server. The recorder received one POST to flagevaluation and no OPTIONS request.

### DIFF
--- a/packages/datadog_flags/lib/src/evaluation_aggregator.dart
+++ b/packages/datadog_flags/lib/src/evaluation_aggregator.dart
@@ -259,7 +259,9 @@ class EvaluationAggregator {
   Map<String, Object?> _webDatadogContext() {
     final applicationId = runtime.datadogConfig.applicationId;
     return _removeNullValues({
+      'env': runtime.datadogConfig.env,
       'service': runtime.datadogConfig.service,
+      'version': runtime.datadogConfig.version,
       'rum': applicationId == null
           ? null
           : {
--- a/packages/datadog_flags/lib/src/evaluation_aggregator.dart
+++ b/packages/datadog_flags/lib/src/evaluation_aggregator.dart
@@ -171,16 +171,15 @@ class EvaluationAggregator {
     List<_AggregatedEvaluation> evaluations, {
     required bool rescheduleOnFailure,
   }) async {
-    final nativeBody = jsonEncode({
-      'context': _datadogContext(),
-      'flagEvaluations': evaluations.map((e) => e.toJson()).toList(),
-    });
     final request = buildFlagsIntakeRequest(
       endpoint: _evaluationEndpoint(),
       clientToken: runtime.datadogConfig.clientToken,
       nativeContentType: 'application/json',
-      nativeBody: nativeBody,
-      webBody: evaluations.map(_webEvaluationJson).join('\n'),
+      nativeBodyBuilder: () => jsonEncode({
+        'context': _datadogContext(),
+        'flagEvaluations': evaluations.map((e) => e.toJson()).toList(),
+      }),
+      webBodyBuilder: () => evaluations.map(_webEvaluationJson).join('\n'),
     );
 
     try {
--- a/packages/datadog_flags/lib/src/evaluation_aggregator.dart
+++ b/packages/datadog_flags/lib/src/evaluation_aggregator.dart
@@ -7,13 +7,12 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:meta/meta.dart';
-import 'package:uuid/uuid.dart';
 
 import 'assignment.dart';
 import 'evaluation_context.dart';
 import 'flags_runtime.dart';
+import 'intake_request.dart';
 import 'json_value.dart';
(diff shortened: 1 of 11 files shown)


fuzzybinary:
@codex review

packages/datadog_flags/test/evaluation_aggregator_test.dart
```diff
@@ -80,6 +79,13 @@ void main() {
         'companyId': '1',
         'plan': 'pro',
       },
+      if (isWebFlagsIntake)
+        'dd': {
+          'service': 'shopping-cart',
+          'rum': {
+            'application': {'id': 'application-id'},
+          },
+        },
```
barboraplasovska, line 90: This expected dd maps only asserts `service` and `rum`, once `_webDatadogContext()` is fixed to include `env/version` it should verify those as well.

packages/datadog_flags/lib/src/evaluation_aggregator.dart
```diff
@@ -239,14 +238,34 @@ class EvaluationAggregator {
...
+      event['context'] = eventContext;
+    }
+    return jsonEncode(event);
+  }
+
+  Map<String, Object?> _webDatadogContext() {
+    final applicationId = runtime.datadogConfig.applicationId;
+    return _removeNullValues({
+      'service': runtime.datadogConfig.service,
+      'rum': applicationId == null
```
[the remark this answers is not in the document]
barboraplasovska: I agree with this!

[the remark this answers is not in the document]
leoromanovsky: Added env and application version to each Flutter Web evaluation event context.dd in commit 764cb4b4. The real Chrome integration test still records one POST and no OPTIONS.

packages/datadog_flags/test/evaluation_aggregator_test.dart
leoromanovsky: Updated both web payload assertions and the real Chrome integration test to verify env and version in commit 764cb4b4.

packages/datadog_flags/lib/src/evaluation_aggregator.dart
barboraplasovska: `_webDatadogContext()` and `_datadogContext()` are identical

packages/datadog_flags/lib/src/evaluation_aggregator.dart
```diff
@@ -172,24 +171,24 @@ class EvaluationAggregator {
...
+    final nativeBody = jsonEncode({
       'context': _datadogContext(),
       'flagEvaluations': evaluations.map((e) => e.toJson()).toList(),
     });
+    final request = buildFlagsIntakeRequest(
+      endpoint: _evaluationEndpoint(),
+      clientToken: runtime.datadogConfig.clientToken,
+      nativeContentType: 'application/json',
+      nativeBody: nativeBody,
+      webBody: evaluations.map(_webEvaluationJson).join('\n'),
```
barboraplasovska: `webBody` is built on every flush even on native where it's unused, consider only building it when the platform actually needs it

packages/datadog_flags/lib/src/intake_platform.dart
```diff
@@ -0,0 +1,7 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+export 'intake_platform_native.dart'
+    if (dart.library.js_interop) 'intake_platform_web.dart';
```
barboraplasovska: /question Why are we using `dart.library.js_interop` here instead of the `dart.library.io` check the rest of the repo uses (`ddlogs_method_channel.dart`, `datadog_sdk_method_channel.dart`)?

leoromanovsky: Changed the conditional export to match the repository pattern in commit da02c935. It now defaults to the web implementation and selects native when dart.library.io is available. Native and Chrome tests pass.

packages/datadog_flags/lib/src/evaluation_aggregator.dart
leoromanovsky: Changed request bodies to lazy builders in commit da02c935. The request builder now invokes only the selected platform callback. A new native and Chrome regression test verifies this behavior.

barboraplasovska (approved):

Issue #1062, opened 2026-06-15
[iOS] NSInternalInconsistencyException crash in DatadogRumPlugin — onSessionStart callback invokes methodChannel on torn-down Flutter engine


Anderzzon:
### Stack trace

Fatal Exception: NSInternalInconsistencyException
  0  CoreFoundation   __exceptionPreprocess
  1  libobjc.A.dylib  objc_exception_throw
  2  Foundation       _userInfoForFileAndLine
  3  Flutter          (Missing UUID 4c4c44de55553144a112d59d3471022b)
  4  Flutter          (Missing UUID 4c4c44de55553144a112d59d3471022b)
  5  Flutter          (Missing UUID 4c4c44de55553144a112d59d3471022b)
  6  Runner           closure #1 in closure #1 in DatadogRumPlugin.enable(arguments:call:) + 405 (DatadogRumPlugin.swift:405)
  7  Runner           <deduplicated_symbol>
  8  libdispatch      _dispatch_call_block_and_release
  9  libdispatch      _dispatch_client_callout
  10 libdispatch      _dispatch_main_queue_drain
  ...

### Volume

1 user so far

### Affected SDK versions

v3.x

### Does the crash manifest in the latest SDK version?

Yes

### Setup Type

Flutter Application

### Other relevant information

A fatal NSInternalInconsistencyException crash occurs on iOS inside DatadogRumPlugin.enable, specifically in the onSessionStart closure dispatched to the main queue. The crash is reported as fatal in Crashlytics.

fuzzybinary:
Hi @Anderzzon ,

Thanks for reporting and sorry for the inconvenience!

Can you give us a little bit more information about your setup? Are you in a Hybrid app or is this a normal Flutter application? Additionally, do you have any steps that might reproduce the issue?

I see something we can change that might fix your issue, but I'd love to try to have a reproducible case before I make any assumptions.

Anderzzon:
Hi,
Thank you for looking into this, it is a normal Flutter application.
I haven't been able to reproduce this myself yet, this is a crash reported from Crashlytics on an iPhone 15 running iOS 26.5.0. 
It looks like it happens at the same time as the app is put into background.

fuzzybinary:
Thanks @Anderzzon,

[...]

PR #1115, opened 2026-08-11
fix: Attempt to avoid crashes caused by forced app termination

fuzzybinary:
### What and why?

When Flutter is shutting down due to a force close (willTerminateNotification), it doesn't inform plugins that they need to detach from the engine. If a background thread is in the middle of calling a mapper or performing onContextChange, this can cause a crash.

To avoid this, we listen to `willTerminate` directly, and ensure we've torn down these functions so they can't be called accidentally. Since the whole application is going away, we don't have to worry about Flutter attempting to reattach an engine.

As part of this, refactor some the plugins to prevent similar issues, and make the setup for this detach more consistent.

This crash is incredibly difficult to reproduce, so we are guessing that this is a complete solution for preventing it.

refs: #1062

### Review checklist

### DIFF
--- a/packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogRumPlugin.swift
+++ b/packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogRumPlugin.swift
@@ -58,15 +58,16 @@ public extension RUM.Configuration {
 
 // swiftlint:disable:next type_body_length
 public class DatadogRumPlugin: NSObject, FlutterPlugin {
-    /// Assigned by `register(with:)` rather than by `init`, because `instance` is created before any
-    /// registrar exists. Internal so tests can seed and inspect it.
-    internal var methodChannel: FlutterMethodChannel?
+    /// No-op: registration now happens per-instance via `attachToEngine(registrar:)`, called by
+    /// `DatadogSdkPlugin`, which owns one `DatadogRumPlugin` instance per engine. This static method
+    /// only exists to satisfy `FlutterPlugin` conformance and is never actually invoked, since this
+    /// type is not listed in `GeneratedPluginRegistrant` (only `DatadogSdkPlugin` is).
+    public static func register(with registrar: FlutterPluginRegistrar) {}
 
-    public static let instance =  DatadogRumPlugin()
-    public static func register(with registrar: FlutterPluginRegistrar) {
+    func attachToEngine(registrar: FlutterPluginRegistrar) {
         let channel = FlutterMethodChannel(name: "datadog_sdk_flutter.rum", binaryMessenger: registrar.messenger())
-        instance.methodChannel = channel
-        registrar.addMethodCallDelegate(instance, channel: channel)
+        methodChannel = channel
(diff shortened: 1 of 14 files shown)


simaoseica-dd:
@codex review

packages/datadog_session_replay/ios/datadog_session_replay/Sources/Swift/DatadogSessionReplayPlugin.swift
```diff
@@ -55,7 +62,37 @@ public class DatadogSessionReplayPlugin: NSObject, FlutterPlugin {
...
+    private func observeEngineTeardown() {
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(engineWillTearDown),
+            name: UIApplication.willTerminateNotification,
+            object: nil
+        )
+    }
+
+    /// Internal rather than private so tests can drive it without posting the real notification,
```
simaoseica-dd: > tests can drive it without posting the real notification

You could take advantage of DI in these entities by injecting the notificationCenter, in the constructor. 
For tests it would help because you could inject a `NotificationCenterMock`

fuzzybinary:
@codex review

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogRumPlugin.swift
```diff
@@ -398,14 +431,17 @@ public class DatadogRumPlugin: NSObject, FlutterPlugin {
     private func enable(arguments: [String: Any?], call: FlutterMethodCall) {
         let configArg = arguments["configuration"] as? [String: Any?]
         if rum == nil {
-            if let configArg = configArg,
-               var config = RUM.Configuration(fromEncoded: configArg) {
+            if RUM._internal.isEnabled() {
+                // Another engine's instance already configured RUM; just attach to it.
+                rum = RUMMonitor.shared()
+                warnIfConfigMismatch(configArg)
```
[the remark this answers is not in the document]
fuzzybinary: This will have to be a known gap, at least until we see an actual instance of this being needed. Most uses cases should only have one engine initializing Datadog, even if multiple exist.

packages/datadog_flutter_plugin/ios/datadog_flutter_plugin/Sources/datadog_flutter_plugin/DatadogSdkPlugin.swift
```diff
@@ -113,20 +113,27 @@ public class DatadogSdkPlugin: NSObject, FlutterPlugin, DatadogFeature {
     var core: DatadogCoreProtocol?
     var oldConsolePrint: ((String, CoreLoggerLevel) -> Void)?
 
+    internal let rum = DatadogRumPlugin()
+    internal let logs = DatadogLogsPlugin()
+
     public init(channel: FlutterMethodChannel) {
```
simaoseica-dd: This entity could also receive the `notificationCenter` via constructor to help tests and not call `NotificationCenter.default` directly.

simaoseica-dd (dismissed):

simaoseica-dd (approved):

PR #1117, opened 2026-08-12
chore: Add package discovery + release plan foundation

fuzzybinary:
### What and why?

We're working on a new / simpler release process for all of the Flutter libraries, which will be necessary when we're close to 15+ with the new federation system. This is the first step to provide tooling that will be able to determine what changes occurred in each package and the nature of those changes are.

This portion replaces the hardcoded, federation-unaware package_list.dart approach with dynamic discovery (packages grouped by shared parent directory, ordered by publish dependency: platform interface -> impls -> app-facing) and a shared release_plan.dart that will let preview/prepare release tooling compute the same plan without duplicating logic.

Also started adding unit tests to better check our release tooling.

### Review checklist

### DIFF
--- a/tools/releaser/lib/package_discovery.dart
+++ b/tools/releaser/lib/package_discovery.dart
@@ -1,0 +1,181 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'dart:io';
+
+import 'package:glob/glob.dart';
+import 'package:glob/list_local_fs.dart';
+import 'package:path/path.dart' as p;
+import 'package:pubspec_parse/pubspec_parse.dart';
+
+/// Suffix a federated package's platform-interface uses. Within a group,
+/// this one always releases first.
+const platformInterfaceSuffix = '_platform_interface';
+
+/// Suffixes of federated per-platform implementation packages. Within a
+/// group these are independent of each other -- any relative order among
+/// them is fine.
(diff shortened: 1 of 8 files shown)


fuzzybinary:
@codex review

tools/releaser/lib/package_discovery.dart
```diff
@@ -0,0 +1,181 @@
...
+
+    if (pubspec.publishTo == 'none') continue;
+
+    final relativePath = p.dirname(
+      p.relative(pubspecFile.path, from: repoRoot),
+    );
+    packages.add(
+      DiscoveredPackage(
+        name: pubspec.name,
+        version: pubspec.version?.toString() ?? '0.0.0',
```
kikoveiga: I guess this default `0.0.0` should never happen, right?  Can we convert to an error when `pubspec.version` is `null` instead?

tools/releaser/test/support/fixture_repo.dart
```diff
@@ -0,0 +1,98 @@
...
+    );
+    repo._writePubspec(
+      '$flutterPluginBase/datadog_flutter_plugin_android',
+      name: 'datadog_flutter_plugin_android',
+      version: '1.0.0',
+    );
+    repo._writePubspec(
+      '$flutterPluginBase/datadog_flutter_plugin_ios',
+      name: 'datadog_flutter_plugin_ios',
+      version: '1.0.0',
```
kikoveiga, line 65: Can we also include `desktop` and `web` or not relevant for now?

kikoveiga (commented):
Worth to look at Codex comments as well, they seem valid.

fuzzybinary:
@codex review

kikoveiga (approved):

PR #1118, opened 2026-08-13
chore: Fix issues from Dart 3.13 upgrade

fuzzybinary:
### What and why?

Dart 3.13 added dot shorthand, which Flutter 3.47 uses extensively. But, because we're using older versions of `json_serializable` (which include older versions of `analyzer`), running json generation on files that include that doc shorthand causes the generation to crash.

Fix by only running the json_serializable on files that need it, and moving RUM enums out of a file that includes `flutter` libraries so that `ddrum_events.dart` can import them cleanly.

Thankfully, this doesn't affect consumers of the library as we ship pre-generated files, and they do not have to run the generation themselves.

Remove JSON generation from our build as well. This was done when we didn't commit the generated files and is no longer necessary.

Also fix datadog_dio build issues with Flutter 3.47.

### Review checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,8 +35,7 @@ Next, at the root of the repository, bootstrap melos:
 melos bootstrap
 ```
 
-Next, we need to generate some necessary files with `flutter pub run build_runner build`, 
-and generate `.env` files for the various apps in order to use them with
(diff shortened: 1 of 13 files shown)


fuzzybinary:
@codex review

maycmlee (approved):

abrooksv (approved):

PR #1123, opened 2026-08-18
chore: Fix multiple warnings that were language version dependent

fuzzybinary:
### What and why?

Fixed multiple analysis infos that required an update to Dart 3.8+. Some of these were older but we're getting to them now with the Dart 3.10 upgrade.

### Review checklist

### DIFF
--- a/tools/ci/bin/desktop_integration_tests.dart
+++ b/tools/ci/bin/desktop_integration_tests.dart
@@ -31,7 +31,7 @@ String currentDevice() {
 // lib/integration_scenarios/scenario_runner.dart.
 const _dataStorageService = 'com.datadoghq.flutter.integration';
 
-const _appExecutableName = 'datadog_integration_test_app';
+final packageName = Platform.environment['MELOS_PACKAGE_NAME'] ?? 'desktop';
 
 // Mirrors DesktopPlatform._storagePath in datadog_flutter_plugin_desktop.
 Directory _dataStorageDirectory() {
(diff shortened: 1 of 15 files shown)


kikoveiga (dismissed):

barboraplasovska (dismissed):
LGTM

fuzzybinary:
@kikoveiga Can I get a re-review on this? It now adds a build step to check that FFI is up to date before running builds. This is to prevent issues in integration testing we were seeing as part of this PR.

tools/ci/bin/desktop_integration_tests.dart
```diff
@@ -31,6 +31,8 @@ String currentDevice() {
 // lib/integration_scenarios/scenario_runner.dart.
 const _dataStorageService = 'com.datadoghq.flutter.integration';
 
+const _appExecutableName = 'datadog_integration_test_app';
```
kikoveiga: Where did we end up using this string?

kikoveiga (dismissed):

tools/ci/bin/desktop_integration_tests.dart
fuzzybinary: Sorry, from a previous implementation. I'll remove it.

fuzzybinary: Fixed and added back in the `tojunit` support that we removed for debugging purposes (and the retries that should no longer be necessary).

tools/ci/bin/desktop_integration_tests.dart
```diff
@@ -155,19 +197,25 @@ Future<int> _runTestWithJunit(List<String> args, String outputFile) async {
...
   // If tojunit exits early (e.g. it can't parse the test output) it closes its
   // stdin, and writing to it after that throws - swallow that here so a tojunit
   // failure can never mask the real flutter test exit code below.
   final pipeDone = testProcess.stdout
       .pipe(tojunitProcess.stdin)
       .catchError((_) {});
+  testProcess.stderr
+      .transform(utf8.decoder)
+      .transform(const LineSplitter())
+      .listen((line) => print('[${_timestamp()}] $line'));
```
kikoveiga: This is duplicated above, and I believe the reason why integration tests are failing:

> Bad state: Stream has already been listened to.

tools/ci/bin/verify_ffi_bindings.dart
```diff
@@ -0,0 +1,209 @@
...
+    '--no-pager',
+    'diff',
+    '--no-index',
+    '--',
+    backup.path,
+    bindings.path,
+  ], workingDirectory: root);
+  print(diff.stdout);
+
+  if (args.contains('--write')) {
```
kikoveiga, line 191: If this succeeds, should we return success immediately?

tools/ci/bin/desktop_integration_tests.dart
fuzzybinary: That'll teach me to try to do this quickly 😅

tools/ci/bin/verify_ffi_bindings.dart
fuzzybinary: If the generation succeeds and there's no mismatch, we exit immediately. If there is a mismatch and `--write` is set than we write out the diff and then exit with an error code.

I think this is set up correctly.

kikoveiga (approved):

tools/ci/bin/verify_ffi_bindings.dart
kikoveiga: I meant this mistmatch fixed by `--write` path, not the main branch with `if (after == before)` above.

fuzzybinary: Ah -- no even it `--write` succeeds, the tool is still emits a failure code, not success. The error code tells us that there's a mismatch. `--write` is just a convenience to leave the newly regenerated code in place.

Issue #944, opened 2026-02-09
App Startup TTID TTFD data missing in default RUM Performance Dashboard


vincevargadev:
### Question

I worked on integrating Datadog into a Flutter mobile app (iOS and Android). I see the logs, the sessions, the actions, crash free sessions, http requests, everything seems to be working well...

...but when I go to the default RUM performance dashboard (`https://app.datadoghq.com/dash/integration/rum_performance_mobile`), there is no data in the App Startup "widget", the TTID and TTFD graphs are all empty.

I couldn't figure out if I set up the project incorrectly / incompletely, or if it's not supported by the plugin yet.

<img width="925" height="846" alt="Image" src="https://github.com/user-attachments/assets/73ea5c09-491b-424d-9a8d-95f409f4022c" />

fuzzybinary:
Hi @vincevargadev,

We're still working on TTFD and TTID for Flutter. The most recent version the Flutter SDK (3.0.1) may give you TTID for Android devices, but neither platform currently supports TTFD.

If everything else is working, then you should be good!  We'll hopefully get TTID out very soon, with TTFD shortly after.

blackmenthor:
Hi @fuzzybinary I was working on DD RUM integration for our Flutter app, we're interested in both TTID and TTFD numbers for both Android and iOS.

- For TTID, in iOS it seems to be logging the data correctly and we can see `rum.measure.app.startup_to_initial_display` in the Dashboard. But We cannot seem to see any data for Android. Any idea why or if there's any guide on how to setup this?
- For TTFD, we cannot see any numbers for Android nor iOS in the Dasboard.

We use the 3.4.1 version of the Flutter SDK, and we rely 100% on the Flutter SDK code for the integration, so no Native code involved in our end.

Can you give some clarity here? Thanks

fuzzybinary:
Hi @blackmenthor,

TTID / TTFD still aren't officially supported. iOS is reporting it'd default time, but we're not sure that's the most useful number for Flutter applications. It's measuring the time from app boot to the first CADisplayLink frame, which may or may not be as useful for Flutter applications, I need to experiment a bit.

It's on my list to get to. Sorry for the delay!

blackmenthor:
Thanks for the clarifiation @fuzzybinary for now I think I would have to create my own metrics for TTID and TTFD.

[...]

PR #1124, opened 2026-08-18
feat: Add reportAppFullyDisplayed to RUM for app launch time to full display (TTFD)

brunovsiqueira:
### What and why?

Bridges `reportAppFullyDisplayed()` from the native SDKs to Dart.

Both native SDKs have shipped this API for a while — `RUMMonitorProtocol.reportAppFullyDisplayed()` on iOS and `RumMonitor.reportAppFullyDisplayed()` on Android — but it was never exposed through the Flutter plugin. `addViewLoadingTime` was bridged in #655; the app-launch equivalent was not. Without this method there is no way for a Flutter app to mark the moment its UI is fully displayed, so `rum.measure.app.startup_to_full_display` never populates for Flutter apps at all.

#944 is the standing request for TTFD support on Flutter.

### How?

Purely additive: one new no-argument method threaded through the existing layers — `DatadogRum` → `DdRumPlatform` → method channel → `DatadogRumPlugin` on iOS and Android. Web is a no-op, as the Browser SDK has no equivalent (same treatment as `addViewLoadingTime`).

No existing behaviour changes. `addViewLoadingTime` is deliberately untouched — that is per-view loading time, a different metric.

### Review checklist

Tests mirror the `addViewLoadingTime` ones: a method-channel test in `ddrum_method_channel_test.dart`, plugin tests plus contract entries on both `DatadogRumPluginTest.kt` and `DatadogRumPluginTests.swift`, and the call is exercised in the manual RUM integration scenario. The existing native mocks already implemented `reportAppFullyDisplayed`, so they needed no changes.

### DIFF
--- a/packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
+++ b/packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
@@ -187,6 +187,28 @@ void main() {
     expect(view1.vitalStepEvents[2].vitalOperationKey, isNull);
     expect(view1.vitalStepEvents[2].vitalFailureReason, 'error');
 
+    // `reportAppFullyDisplayed` reports TTFD as an app launch vital rather than
+    // as a property of the view it was called from, so look for it across the
+    // whole session. It is not supported by the Browser SDK.
+    if (!kIsWeb) {
+      final ttfdVitals = rumLog
+          .where((e) =>
+              e.eventType == 'vital' &&
+              RumVitalAppLaunchEventDecoder.isAppLaunchVital(e.rumEvent))
+          .map((e) => RumVitalAppLaunchEventDecoder(e.rumEvent))
+          .where((e) => e.appLaunchMetric == 'ttfd')
+          .toList();
+
+      // Only the first call to `reportAppFullyDisplayed` is reported.
+      expect(ttfdVitals.length, 1);
+      // TTFD is measured from the launch of the app, so it should be at least
+      // as long as the fake loading the scenario performs before reporting it.
+      expect(ttfdVitals[0].duration,
+          greaterThanOrEqualTo(const Duration(milliseconds: 50).inNanoseconds));
+      expect(ttfdVitals[0].duration,
+          lessThan(const Duration(seconds: 60).inNanoseconds));
+    }
+
     // Verify user in all events, except for the first view event
     for (final viewEvent in view1.viewEvents.sublist(1)) {
       verifyUser(viewEvent);
--- a/packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
+++ b/packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
@@ -189,8 +189,13 @@ void main() {
 
     // `reportAppFullyDisplayed` reports TTFD as an app launch vital rather than
     // as a property of the view it was called from, so look for it across the
-    // whole session. It is not supported by the Browser SDK.
-    if (!kIsWeb) {
+    // whole session.
+    //
+    // This is only checked on iOS. The Browser SDK has no equivalent API, and
+    // the Android SDK only sends TTFD once it has computed TTID for the startup
+    // scenario, which does not happen in this app -- it sends no app launch
+    // vitals at all, so there is nothing to assert on there yet.
+    if (!kIsWeb && Platform.isIOS) {
       final ttfdVitals = rumLog
           .where((e) =>
               e.eventType == 'vital' &&
--- a/packages/datadog_flutter_plugin/integration_test_app/lib/integration_scenarios/rum_manual_instrumentation_scenario.dart
+++ b/packages/datadog_flutter_plugin/integration_test_app/lib/integration_scenarios/rum_manual_instrumentation_scenario.dart
@@ -105,6 +105,7 @@ class _RumManualInstrumentationScenarioState
     await Future<void>.delayed(const Duration(milliseconds: 50));
     DatadogSdk.instance.rum?.addTiming('content-ready');
     DatadogSdk.instance.rum?.addViewLoadingTime();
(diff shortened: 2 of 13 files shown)


packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
```diff
@@ -202,6 +202,14 @@ class DdRumMethodChannel extends DdRumPlatform {
     });
   }
 
+  @override
+  Future<void> reportAppFullyDisplayed() {
+    return methodChannel.invokeMethod(
+      'reportAppFullyDisplayed',
+      <String, Object?>{},
```
fuzzybinary, line 209: nit: We tend to add the type annotations on these maps, especially if they're empty.

packages/datadog_flutter_plugin/integration_test_app/lib/integration_scenarios/rum_manual_instrumentation_scenario.dart
```diff
@@ -105,6 +105,7 @@ class _RumManualInstrumentationScenarioState
     await Future<void>.delayed(const Duration(milliseconds: 50));
     DatadogSdk.instance.rum?.addTiming('content-ready');
     DatadogSdk.instance.rum?.addViewLoadingTime();
+    DatadogSdk.instance.rum?.reportAppFullyDisplayed();
```
fuzzybinary, line 109: If we're going to add this to the scenario, we should verify in the test that the value was set to some reasonable value in the test.

fuzzybinary (changes requested):
Looks good, just one very minor request - if we're going to add TTFD to the test scenario we should verify it sent the value.

packages/datadog_flutter_plugin/integration_test_app/lib/integration_scenarios/rum_manual_instrumentation_scenario.dart
brunovsiqueira: Done in fae6387d.

TTFD isn't a view property, it's a separate app launch vital. Those share the `vital` event type with operation steps, and `RumSessionDecoder` was decoding every vital as an operation step, so the TTFD one landed in `vitalStepEvents` and broke the existing count (3 became 4). Decoder now splits on `vital.type`.

Test asserts one `ttfd` vital with a sane duration. Ran it on the iOS simulator, passes. Removing the `reportAppFullyDisplayed()` call makes it fail with `Expected: <1> Actual: <0>`, so it's not vacuous. Only ran iOS locally, CI will cover Android.

packages/datadog_flutter_plugin/lib/src/rum/ddrum_method_channel.dart
brunovsiqueira: The map is already `<String, Object?>{}`, same as `stopSession` below. Did you mean `invokeMethod<void>`?

fuzzybinary: Sorry, I missed a "not" in there. We tend to *not* add the type annotations on these maps. It slipped in on `stopSession`, and not a big deal so I'm not going to worry about it.

packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
```diff
@@ -187,6 +187,28 @@ void main() {
...
+    // as a property of the view it was called from, so look for it across the
+    // whole session. It is not supported by the Browser SDK.
+    if (!kIsWeb) {
+      final ttfdVitals = rumLog
+          .where((e) =>
+              e.eventType == 'vital' &&
+              RumVitalAppLaunchEventDecoder.isAppLaunchVital(e.rumEvent))
+          .map((e) => RumVitalAppLaunchEventDecoder(e.rumEvent))
+          .where((e) => e.appLaunchMetric == 'ttfd')
+          .toList();
```
fuzzybinary, line 200: I'll likely refactor this to have all vitals accessible on the `session` but for now this is fine.

packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
```diff
@@ -187,6 +187,28 @@ void main() {
...
+
+      // Only the first call to `reportAppFullyDisplayed` is reported.
+      expect(ttfdVitals.length, 1);
+      // TTFD is measured from the launch of the app, so it should be at least
+      // as long as the fake loading the scenario performs before reporting it.
+      expect(ttfdVitals[0].duration,
+          greaterThanOrEqualTo(const Duration(milliseconds: 50).inNanoseconds));
+      expect(ttfdVitals[0].duration,
+          lessThan(const Duration(seconds: 60).inNanoseconds));
```
fuzzybinary, line 209: We'll see if this is enough for CI or if it flakes 😅.

fuzzybinary (dismissed):

fuzzybinary:
@brunovsiqueira You need to run `melos dartfmt`

packages/datadog_flutter_plugin/integration_test_app/integration_test/rum_manual_test.dart
```diff
@@ -187,6 +187,33 @@ void main() {
...
 
+    // `reportAppFullyDisplayed` reports TTFD as an app launch vital rather than
+    // as a property of the view it was called from, so look for it across the
+    // whole session.
+    //
+    // This is only checked on iOS. The Browser SDK has no equivalent API, and
+    // the Android SDK only sends TTFD once it has computed TTID for the startup
+    // scenario, which does not happen in this app -- it sends no app launch
+    // vitals at all, so there is nothing to assert on there yet.
+    if (!kIsWeb && Platform.isIOS) {
```
fuzzybinary, line 198: If that's the case, we may want to hold off on this change. I know we're in the process of looking into TTID for Android. Let me look into the state of it and I'll get back to you.

PR #1125, opened 2026-08-20
feat(flags): send the split serial id on exposure events

danyal002:
### What and why?

EX-3420. This change adds a `serial_id` field to the exposure event.

The serial id identifies the split that the subject was assigned to. The intake uses it to find the holdout that the allocation comes from. A holdout becomes a usual allocation before the SDK receives it, so the exposure event does not record the holdout in another way.

The precompute response already sends the value in the `serialId` field of each flag. The SDK discarded it.

### How?

| File | Change |
|---|---|
| `lib/src/assignment.dart` | Add an optional `serialId` to `FlagAssignment`. |
| `lib/src/assignment.g.dart` | Generated again. |
| `lib/src/exposure_logger.dart` | Set `serial_id` on the event, and add the value to the exposure cache. |

The server sends an absent key or an explicit null when a flag has no serial id. Both give a null `serialId`, and the SDK then sends no field.

The SDK does not validate the value. The flag configuration layer validates it.

Three new tests. They pin a serial id of `0`, which is a usual value because serial ids start at zero for each organization, a serial id of `340132`, an explicit null, and an absent key.

### Risks

Low. The field is optional. An exposure event without a serial id does not change.

The exposure cache now compares the serial id in addition to the allocation key and the variation key. A subject therefore reports one more exposure when its serial id appears or changes. This is intended: a configuration refresh can add a serial id while both keys stay the same, and the intake must receive that value.

### Review checklist

*This description was generated by Claude.*

### DIFF
--- a/packages/datadog_flags/lib/src/exposure_logger.dart
+++ b/packages/datadog_flags/lib/src/exposure_logger.dart
@@ -47,6 +47,7 @@ class ExposureLogger {
     final cacheValue = _ExposureCacheValue(
       allocationKey: assignment.allocationKey,
       variationKey: assignment.variationKey,
+      serialId: assignment.serialId,
     );
     if (_loggedAssignments[cacheKey] == cacheValue) {
       return;
@@ -183,11 +184,13 @@ class ExposureLogger {
       'id': evaluationContext.targetingKey,
       'attributes': sanitizeJsonValue(evaluationContext.attributes),
     });
+    final serialId = assignment.serialId;
     return {
       'timestamp': runtime.configuration.dateProvider().millisecondsSinceEpoch,
       'allocation': {'key': assignment.allocationKey},
       'flag': {'key': flagKey},
       'variant': {'key': assignment.variationKey},
+      if (serialId != null) 'serial_id': serialId,
       'subject': subject,
     };
   }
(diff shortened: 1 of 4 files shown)


typotter (approved):

fuzzybinary (approved):

packages/datadog_flags/lib/src/exposure_logger.dart
```diff
@@ -183,11 +184,13 @@ class ExposureLogger {
       'id': evaluationContext.targetingKey,
       'attributes': sanitizeJsonValue(evaluationContext.attributes),
     });
+    final serialId = assignment.serialId;
     return {
       'timestamp': runtime.configuration.dateProvider().millisecondsSinceEpoch,
       'allocation': {'key': assignment.allocationKey},
       'flag': {'key': flagKey},
       'variant': {'key': assignment.variationKey},
+      if (serialId != null) 'serial_id': serialId,
```
fuzzybinary, line 193: nit: What's your minimum Dart version for this package? Above 3.8 you can use null aware elements:

```dart
'serial_id': ?serialId
```

danyal002: The minimum is `3.6.0`, so this package cannot use null-aware elements yet. The analyzer gives:

```
error - This requires the 'null-aware-elements' language feature to be enabled.
Try updating your pubspec.yaml to set the minimum SDK constraint to 3.8.0 or higher
```

I kept the `if` element. A change to the minimum SDK constraint of a published package is a separate decision, and I do not want to make it in this pull request.

PR #168, opened 2022-07-14
📝 Add error manual error tracking documentation

fuzzybinary:
### What and why?

We've had some clients ask about the state of Native Crash Reporting and desymbolication. I'd like to add these docs and somewhere to point them to while we work on the simplified / unified solution.

### Review checklist

- [ ] This pull request has appropriate unit and / or integration tests 
- [ ] This pull request references a Github or JIRA issue

### CI Configuration (optional)
- [ ] Run unit tests
- [ ] Run integration tests

PR #1134, opened 2026-08-26
[FFL-3028] Add Datadog OpenFeature provider

leoromanovsky:
## Motivation

Dart and Flutter applications need a canonical OpenFeature integration without a second Datadog package. The integration must keep Dart 3.6 compatibility.

This PR is stacked on [#1155](https://github.com/DataDog/dd-sdk-flutter/pull/1155), which adds the core initialization timeout.

[OpenFeature Dart PR #168](https://github.com/open-feature/dart-server-sdk/pull/168) lowers the client SDK minimum from Dart 3.12.2 to Dart 3.6.0.

## Changes and Decisions

```mermaid
flowchart TD
    subgraph Customer["Customer application"]
        Dart["Dart application"] --> API["OpenFeature client API"]
        Flutter["Flutter application"] --> API
    end

    SDK["openfeature_dart_client_sdk<br/>Dart 3.6+"] --> API
    API --> Provider["DatadogOpenFeatureProvider<br/>exported by datadog_flags"]
    Provider --> Core["datadog_flags core"]
    Core --> Init["Initialize evaluation context"]
    Init --> Work["Load cache → request assignments → decode JSON → store state"]
    Work --> Deadline{"Finished before timeout?"}
    Deadline -->|Yes| Ready["Provider ready"]
    Deadline -->|No| Defaults["Return customer defaults"]
    Defaults -->|Late success| Ready
    Core --> Telemetry["Datadog exposure and evaluation telemetry"]

    Flutter -.->|Optional RUM, Logs, and Traces| FlutterSDK["Datadog Flutter SDK"]
    FlutterSDK --> Datadog["Datadog"]
    Telemetry --> Datadog
```

- Export `DatadogOpenFeatureProvider` from `datadog_flags`. Do not add a separate Datadog provider package.
- Release the provider API as `datadog_flags` `1.2.0-beta.1` while the OpenFeature client remains beta.
- Make OpenFeature the recommended integration in the package documentation and Flutter example.
- Keep the existing Datadog Flags API for low-level lifecycle control and existing users.
- Use committed dependency overrides for the OpenFeature PR branch. The public package keeps a hosted dependency.
- Replace the temporary overrides with the next compatible OpenFeature release before publication.
- Apply the core timeout to cache loading, the network request, JSON decoding, storage, and state publication.
- Keep timed-out initialization work alive. Emit a ready event if the first assignment request succeeds later.
- Keep the previous context active until context reconciliation succeeds.
- Keep the Datadog Flutter SDK separate. It continues to provide RUM, Logs, and Traces.


packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -111,9 +120,13 @@ class FlagsRepository {
         return;
       }
       _state = data;
+      _setStatus(DatadogFlagsClientStatus.ready);
```
aarsilv, line 123: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P0]** Please publish `_state` and the `ready` status before awaiting `_writeCached`.

A store write that never settles turns a **successful** fetch into a permanent provider error. The fetch returns 200 and decodes, then publication never runs. So `statusChanges` never emits `ready`, and the late-success path has nothing to fire on. I also tried bounding the write instead of reordering, and it is not safe — details below.

<details>
<summary><b>Executed evidence</b> — a never-settling store write against an HTTP 200</summary>

A `DatadogFlagsStore` whose `write` returns a future that never completes, a 50 ms `initializationTimeout`, and a mock returning `200` with a `show-paywall` assignment of `true`. Run in `dart:3.6.2`:

| Tree | Result |
| --- | --- |
| PR head `c4bb07b` | `ProviderStatus.error`, value `false` |
| PR head + the fix below | `ProviderStatus.ready`, value `true` |

At head the single provider event is:

```
error: "Datadog assignments are not available before the initialization deadline."
```

The deadline did expire, because storage is inside the budget. What the message hides is that the assignments were fetched in time and only publication was blocked, so it points at the budget rather than at the store. Recovery is impossible, not merely delayed: the late-success path is driven by `statusChanges`, and `_setStatus(ready)` sits behind the stuck write.

On provenance: `origin/develop` publishes before persisting, so the ordering is new relative to `develop` — but it arrived with the base branch, not this PR. What this PR adds is the `ready` transition behind the write.

This PR's own description sets the contract I am measuring against:

> Keep timed-out initialization work alive. Emit a ready event if the first assignment request succeeds later.

</details>

<details>
<summary><b>What this costs</b> — one test asserts the current ordering</summary>

With the fix applied, the full suite is 80/81. The one failure is `initialization_timeout_test.dart:219` in `includes assignment storage before publishing ready state`:

```
Expected: FlagEvaluationError:<FlagEvaluationError.providerNotReady>
  Actual: <null>
```

That test pins the ordering as intended behaviour, so it needs to change with it. #1155 has since renamed it to `includes assignment storage in the total initialization budget`, which reads as being about the budget rather than the order — but that is my inference, not a statement the test makes.

Your description says the budget covers `storage, and state publication`. The fix leaves `await _writeCached(data)` within the same operation, though the await now sits after publication rather than before it. It stops storage gating the `ready` transition, and I would not claim more than that. A deadline expiring neither cancels nor completes a stuck write, so "inside the budget" was never a bound on the store itself.

</details>

<details>
<summary><b>Adversarial checks on the reorder</b> — four cases, same final state on both trees</summary>

A previous attempt at this ordering elsewhere let a slow write land after `reset()` deleted the cache. I tested for that specifically. Output was byte-identical at head and with the fix in all four:

| Case | Result |
| --- | --- |
| `reset()` during a pending slow write | store empty after reset, 1 write / 1 delete, no resurrection |
| `shutdown()` during a pending slow write | no error, no hang |
| second `initialize()` superseding the first mid-write | second context wins, `ready`, second value |
| store write that throws `StateError` | `ready`, publication survives |

One limit this fix does not address, measured on both trees: a write that *never* settles still wedges `_cacheOperation`, so a later `reset()` never runs its delete. With a 50 ms budget and a stuck write, `reset()` had not completed after 400 ms and `delete` was never called — identical at head and with the reorder. That is the existing cache queue, not something the reorder changes, but it means publication is fixed and cache teardown is not.

</details>

<details>
<summary><b>Suggested fix</b> — publish, then persist</summary>

```dart
      final publicationDeadlineYield = _yieldAfterExpiredDeadline(deadline);
      if (publicationDeadlineYield != null) {
        await publicationDeadlineYield;
      }
      if (token.isCanceled) {
        return;
      }
      _state = data;
      _setStatus(DatadogFlagsClientStatus.ready);
      await _writeCached(data);
```

`dart analyze` clean. Two honest costs. Publication now happens before the write starts, so a context superseded mid-write has already made its assignments visible and they stay visible until the newer initialization publishes. Head, publishing after the write, never makes them visible at all: measured `ctx=null`, `notReady` at head against `ctx=user-A`, `ready` with the fix, both settling on the newer context. If the newer initialization never completes, the superseded value stays. `origin/develop` has the same window, so this matches `origin/develop` rather than inventing a behaviour.

It also removes the `token.isCanceled` check that sat between the write and the yield, which is now redundant rather than load-bearing — the state it guarded is published earlier. Re-adding a post-write check to clear `_state` would be wrong, because by then `_state` can belong to the newer initialization.

</details>

<details>
<summary><b>Do not bound the write inside the cache operation</b> — I tried it, and it resurrects deleted data</summary>

Bounding the write looked better: `_readCached` already has `storeReadTimeout`, the write has none, and adding a `storeWriteTimeout` leaves your ordering and its test untouched. It passes `dart analyze`, keeps the suite at 81/81, and on the default budget gives a clean `ready` with no error event.

It is still wrong. `Future.timeout` bounds the *wait*, not the underlying write, and `DatadogFlagsStore.write` returns a plain `Future<void>` with no cancellation. So the abandoned write lands later and can repopulate data that `reset()` already deleted. Measured with a store whose write takes 300 ms against a 100 ms bound:

| Point | Head, unbounded | Bounded write |
| --- | --- | --- |
| after `initialize()` | store `[default]`, write completed | store `[]`, write **not** completed |
| right after `reset()` | store `[]`, 1 delete | store `[]`, 1 delete |
| +500 ms | store `[]` | store **`[default]`** — resurrected |

The reorder does not have this problem, because it never abandons the write, so `_enqueueCacheOperation` ordering is preserved. That is what the adversarial `reset()` case above measures.

To be precise about what I ruled out: I timed out the write *inside* the serialized cache operation. Bounding only the caller's wait while leaving `_cacheOperation` chained to the underlying write is a different shape that I did not test, and it would not have this failure mode.

</details>

packages/datadog_flags/lib/src/datadog_openfeature_provider.dart
```diff
@@ -0,0 +1,508 @@
...
+      case datadog.DatadogFlagsClientStatus.ready:
+      case datadog.DatadogFlagsClientStatus.stale:
+        await _activateRuntime(
+          runtime,
+          status,
+          revision: revision,
+          isInitialization: isInitialization,
+        );
+      case datadog.DatadogFlagsClientStatus.notReady:
+        _emitLoadError(
```
aarsilv, line 296: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P1]** Please reject a disabled or oversized `initializationTimeout` here.

`DatadogFlagsConfiguration.initializationTimeout` documents `null`, zero and negative as ways to disable it. Under this provider, disabling it lets OpenFeature's own lifecycle timeout fire instead. OpenFeature then **permanently quarantines the provider instance**: late success never recovers and the instance cannot be reused. A budget longer than that bound hits the same wall whenever initialization actually runs past OpenFeature's timeout.

<details>
<summary><b>Executed evidence</b> — the instance becomes permanently unusable</summary>

`initializationTimeout: null`, a fetch that never lands, `lifecycleTimeout` shortened to 300 ms so the test is quick. Run in `dart:3.6.2`:

```
first registration : OpenFeatureException: Provider initialization did not
                     complete within 300 ms. (general)
after the fetch succeeds 200 : providerStatus = notReady, value = false
re-registering the SAME instance : OpenFeatureException: A provider instance
                     that timed out cannot be reused. Create a replacement
                     provider instance.
```

Raising that bound means leaving the stable API. In `openfeature_dart_client_sdk`, `lifecycleTimeout` is `final` and defaults to 30 s (`api.dart:23,36`); the only way to set it is `createIsolatedOpenFeatureAPI`, exported solely from `openfeature_dart_client_sdk_experimental.dart`. `OpenFeatureAPI.instance` — what your own `examples/simple_example/lib/main.dart` uses — is fixed at 30 s.

The bound is applied at `api.dart:582` (initialize), `:646` (reconcile) and `:936` (dispose), and each expiry routes to quarantine at `:601`, `:674` and `:949`. `_setProvider` then rejects the instance at `:389-393`.

</details>

<details>
<summary><b>Why not a second timer</b></summary>

#1155 rules that out explicitly:

> The dashed OpenFeature path is future work. It must use the core timeout instead of creating another timer.

So the fix validates the budget instead of bounding it again.

</details>

<details>
<summary><b>Suggested fix</b> — the disabled-budget case executed; the oversized one not</summary>

Alongside the existing metadata constants:

```dart
  /// Largest Datadog initialization budget this provider accepts.
  ///
  /// OpenFeature bounds every provider lifecycle call with its own timeout and
  /// permanently quarantines a provider instance that exceeds it. The Datadog
  /// budget must expire first, so a disabled or larger budget is rejected.
  static const maxInitializationTimeout = Duration(seconds: 20);
```

Then in `_loadContext`, right after the existing `datadogConfig == null` guard:

```dart
    final budget = configuration.initializationTimeout;
    if (budget == null ||
        budget <= Duration.zero ||
        budget > maxInitializationTimeout) {
      _emitLoadError(
        'DatadogFlagsConfiguration.initializationTimeout must be positive and '
        'no longer than ${maxInitializationTimeout.inSeconds}s. OpenFeature '
        'quarantines a provider whose initialization exceeds its own timeout.',
      );
      return;
    }
```

Measured with this applied: the same scenario now reports `Provider initialization ended with status error` immediately, the instance is never quarantined, and **re-registering it succeeds** instead of throwing. `dart analyze` clean and the package suite green.

Three caveats, so you can judge the policy rather than inherit mine. 20 s is arbitrary — it is a ceiling on what the provider accepts, chosen to leave headroom under OpenFeature's 30 s default, and it guarantees nothing: an app using `createIsolatedOpenFeatureAPI` with a shorter `lifecycleTimeout` would still be under-protected. I executed only the disabled-budget case, not an oversized one. And an oversized budget only bites when initialization genuinely runs long, so it is a latent hazard rather than a certain failure.

</details>

packages/datadog_flags/lib/src/datadog_openfeature_provider.dart
```diff
@@ -0,0 +1,508 @@
...
+        type: isInitialization
+            ? openfeature.ProviderEventType.ready
+            : openfeature.ProviderEventType.contextChanged,
+      ),
+    );
+    if (status == datadog.DatadogFlagsClientStatus.stale) {
+      _events.add(
+        openfeature.ProviderEvent(
+          type: openfeature.ProviderEventType.stale,
+          message: 'Using stored Datadog assignments because refresh failed.',
```
aarsilv, line 381: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P1]** This message states a cause that may not have happened.

The refresh has not necessarily failed when this fires. With matching cached assignments and a budget that expires mid-fetch, the provider emits this `stale` message and the fetch then **succeeds**. The enum doc at `flags_client.dart:19` has the same problem: `_setStatus(stale)` runs at `flags_repository.dart:98` before any fetch is attempted.

<details>
<summary><b>Executed evidence</b> — the refresh was still running, then succeeded</summary>

A store holding matching assignments for the context, a 50 ms `initializationTimeout`, and a fetch held open. Run in `dart:3.6.2`:

```
providerStatus = ProviderStatus.stale
events = [ready, stale: "Using stored Datadog assignments because refresh failed."]
cached value visible = true
```

then completing the fetch with `200`:

```
providerStatus = ProviderStatus.ready
value = false   (the freshly fetched value)
```

So the app was told the refresh failed while it was still running, and it then succeeded.

The enum comment reads, verbatim:

```dart
  /// Cached assignments are available because the latest refresh failed.
  stale,
```

Clause one is true wherever this state is set. Clause two is not, and the assignments' origin varies too: the guard at `flags_repository.dart:96` skips `_state = matchingCached` when current state already matches the context, so the `stale` at `:129` can describe assignments an earlier successful fetch produced. The replacement wording below drops the causal clause entirely. It states what is true at every point this status is set — during initialization, during reconciliation, and on a later refresh — and claims nothing about why or where the assignments came from. Clause two is not true at `flags_repository.dart:96-98`, which is reached before `fetcher.fetch` is called.

`README.md:296` repeats it, so there are three places to change, not two:

> - `stale` when matching stored assignments remain usable after refresh fails.

</details>

<details>
<summary><b>Suggested fix</b> — state what is true, drop the cause</summary>

Here:

```dart
          message: 'Using existing Datadog assignments. Fresh assignments '
              'are not available.',
```

In `flags_client.dart`:

```dart
  /// Assignments are available for evaluation, but fresh assignments are not.
  stale,
```

And the matching README bullet:

```
- `stale` when assignments are available for evaluation but fresh assignments
  are not.
```

No logic changes, but the provider message is observable at runtime, so this one does change what an app sees. I applied both Dart edits: `dart analyze` clean and the package suite green over 6 runs.

</details>

packages/datadog_flags/lib/src/datadog_openfeature_provider.dart
```diff
@@ -0,0 +1,508 @@
...
+    if (configuration.datadogConfig == null) {
+      _emitLoadError('Datadog Flags configuration is required.');
+      return;
+    }
+
+    final owner = datadog.DatadogFlags();
+    _ProviderRuntime? runtime;
+    var initializationReturned = false;
+    try {
+      await owner.enable(configuration: configuration);
```
aarsilv, line 222: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** `shutdown()` is not a fence here, so a shut-down provider can still write to the app's store.

A `_loadContext` suspended on this `await` resumes after `shutdown()` bumped `_contextRevision`. It then builds a full runtime and completes its fetch. The guard at lines 209-211 sits above the `await`, so it cannot see the shutdown. Only direct lifecycle calls reach this; `OpenFeatureAPI` serialises them, hence `[P2]`.

<details>
<summary><b>Executed evidence</b> — measured from outside the library</summary>

No instrumentation of `lib/`: a `Zone` with a `createPeriodicTimer` hook counts the `EvaluationAggregator` timers, and a counting `DatadogFlagsStore` observes writes. `provider.initialize(...)` is started unawaited, `provider.shutdown()` is awaited `k` microtasks later, and the fetch is completed only afterwards. `initializationTimeout: null`, `dart:3.6.2`:

| k | timers live when `shutdown()` returned | store writes after shutdown |
| --- | --- | --- |
| 0 | 0 | 0 |
| **1** | **1** | **1** |
| 2-6 | 0 | 0 |

The timer half self-heals once `_isCurrent` rejects the stale runtime. The store write does not — it has already happened. So the consequence is a write-after-shutdown into an app-supplied store, not a permanent leak.

</details>

<details>
<summary><b>Reachability</b> — not through OpenFeature</summary>

`setProviderAndWait` (`api.dart:90-98`), `setEvaluationContextAndWait` (`:134-148`) and `shutdown()` (`:319-323`) all route through `_enqueueBindingMutations` then `_enqueueProviderMutations` (`:512-520`), which chains every mutation onto a per-provider-identity queue.

Measured with a decorator counting concurrent lifecycle calls: 12 fire-and-forget public-API calls gave `max_concurrent = 1, overlaps = 0`, against a positive control of two direct unawaited `onContextChanged` calls that measured 2.

Also worth recording, since Copilot raised the other half of this: I could not reproduce the pending-overwrite it describes. Sweeping the microtask offset across 7 interleavings of two direct unawaited context changes, the newest context won every time and no runtime was left undisabled. That is non-reproduction, not proof of impossibility — when both loads see `_pendingRuntime == null` they pay the same awaits, so I cannot argue the ordering is structurally guaranteed.

</details>

<details>
<summary><b>Suggested fix</b> — re-check the revision after enabling</summary>

```dart
      await owner.enable(configuration: configuration);
      if (revision != _contextRevision) {
        await _disableOwnerQuietly(owner);
        return;
      }
      final client = owner.sharedClient(
```

Measured with this applied: `k=1` becomes 0 timers live, 0 precompute requests, 0 store writes. `dart analyze` clean, full suite green.

</details>

packages/datadog_flags/pubspec.yaml
```diff
@@ -16,6 +16,7 @@ dependencies:
   http: ^1.0.0
   json_annotation: ^4.9.0
   meta: ^1.0.0
+  openfeature_dart_client_sdk: ^0.0.1-beta.1
```
aarsilv, line 19: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** The new `publish_dry_run:flags` gate passes even though this constraint cannot resolve on the SDK floor you declare.

The PR description already says the overrides are temporary and the public package keeps a hosted dependency, so this is about enforcement. The version is now `1.2.0-beta.1`, the gate you added to `build-flutter` exits `0`, and `packages/datadog_flags/pubspec.lock` is untracked, so CI resolves the fork branch tip each run.

<details>
<summary><b>Executed evidence</b> — the gate passes; the same resolve with the overrides removed does not, on Dart 3.6</summary>

In `dart:3.6.2`, in the package exactly as shipped, the gate's command:

```
Package has 0 warnings and 2 hints.
exit=0
```

One hint is the relevant one, verbatim:

> Non-dev dependencies are overridden in pubspec_overrides.yaml.

Then with both `pubspec_overrides.yaml` files removed and `.dart_tool` and `pubspec.lock` cleared, the same command:

```
Because datadog_flags depends on openfeature_dart_client_sdk any which
requires SDK version ^3.12.2, version solving failed.
exit=1
```

Stated precisely: the package is publishable, and it resolves on a newer SDK. What it cannot do is resolve on the `sdk: ">=3.6.0 <4.0.0"` floor this package declares and the PR opens by requiring — a floor `datadog_dio` and `datadog_common_test` share as `^3.6.0`. `pub.dev` lists exactly one version of `openfeature_dart_client_sdk`, `0.0.1-beta.1`, with `environment.sdk: ^3.12.2`, and `open-feature/dart-server-sdk#168` is still open.

Since posting this I installed Flutter 3.47.4, so I could re-run the same thing on today's stable Dart, 3.13.3, with both `pubspec_overrides.yaml` files removed and no lockfile:

```
dart pub get                    + openfeature_dart_client_sdk 0.0.1-beta.1   exit=0
dart analyze .                  No issues found!                             exit=0
dart test                       +81: All tests passed!                       exit=0
dart test --platform chrome     +10: All tests passed!                       exit=0
  ... --compiler dart2wasm      +10: All tests passed!                       exit=0
dart compile js ...smoke.dart   Compiled ... to 11,662 characters JavaScript  exit=0
dart pub publish --dry-run      Package has 0 warnings and 1 hint.           exit=0
```

That covers every gate this PR adds — `unit_test:openfeature-web` both ways and `compile:openfeature-web` — with the fork out of the picture.

That remaining hint is the version one, not the override one — `The previous version is 1.0.1. It seems you are not publishing an incremental update.`

So the fork is not needed to compile or test this package on a current SDK; it is needed only to hold the `>=3.6.0` floor. To be exact about the gate: it is green because the committed override is active, which is true on any SDK — that is the 3.6.2 run above. What the 3.13.3 run adds is that the *unoverridden* pubspec resolves there too, so dropping the override would not strand a developer on current stable. CI is always in the first situation, not the second: the override is committed, so CI never resolves the published constraint at all. `.gitlab-ci.yml:37` runs `flutter upgrade --force` and nothing in the repo pins a Flutter or Dart version, so the SDK version decides only what *would* happen once the override comes out. Either way the declared floor is the part nothing tests.

</details>

<details>
<summary><b>Pinning is asymmetric</b></summary>

`examples/simple_example/pubspec.lock` is tracked and this PR adds `resolved-ref: 3377bc0725f93cabfdc196f1a02711d6779fc656`, so the example is pinned to a commit.

`packages/datadog_flags/pubspec.lock` is not tracked — `.gitignore:11` is a bare `pubspec.lock`, and `git ls-files` lists six lockfiles, none of them that one. So `unit_test:openfeature-web`, `compile:openfeature-web` and `publish_dry_run:flags` all resolve whatever `fix/client-sdk-minimum-dart-version` points at, with no diff here when it moves.

</details>

<details>
<summary><b>Suggested gate</b> — fails while an override exists; still does not test the floor</summary>

```yaml
  publish_dry_run:flags:
    exec:
      concurrency: 1
    run: |
      log="$(dart pub publish --dry-run 2>&1)"; rc=$?
      echo "$log"
      if printf '%s' "$log" | grep -q 'overridden in pubspec_overrides.yaml'; then
        echo "ERROR: a non-dev dependency is overridden, so the published pubspec was never resolved." >&2
        echo "ERROR: remove pubspec_overrides.yaml and depend on a resolvable published version before publishing." >&2
        exit 1
      fi
      exit $rc
    packageFilters:
      scope: 'datadog_flags'
```

Verified as a shell script, `bash -n` clean, three cases: exit `1` with the committed override present; exit `1` with the override removed on Dart 3.6, because the resolve itself then fails; and exit `65` when I corrupted `name:` in `pubspec.yaml`, which confirms a genuine publish failure is no longer swallowed. My first draft of this used `|| true` and did swallow it.

Two caveats. This gate stays red until the committed override is removed — an upstream release alone will not turn it green, since it keys on the override hint rather than on resolution. And it does not test the declared floor; if you want that enforced, the check has to resolve against Dart 3.6 specifically, which this does not do.

There is more than one way out, and the error message deliberately does not pick one: an upstream release compatible with 3.6, or raising this package's floor to match the dependency. Only the first keeps the floor where it is.

</details>

packages/datadog_flags/test/datadog_openfeature_provider_test.dart
```diff
@@ -0,0 +1,538 @@
...
+        EvaluationContext(targetingKey: 'user-second'),
+      ),
+      throwsA(isA<OpenFeatureException>()),
+    );
+
+    expect(client.providerStatus, ProviderStatus.error);
+    expect(client.getBooleanValue('show-paywall', false), isTrue);
+  });
+
+  test('uses matching stored assignments with stale provider status', () async {
```
aarsilv, line 189: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** Cached timeout has no provider-level test, shutdown has no assertion on what it releases, and `_isCurrent`'s rejection path is untested.

Context replacement and late success are genuinely covered here; I killed each with a mutation. Uncached timeout is asserted by the same test as late success, though my mutation only proves the late-success half. Cached timeout is covered only in the core. Shutdown does run at this layer, and its telemetry flush is asserted, but nothing asserts the state it should release. Separately, replacing `_isCurrent`'s body with `return true;` survives **11 of 11 runs**.

<details>
<summary><b>Why this test is not the cached-timeout case</b></summary>

#1155's own wording, verbatim:

> Test `cached timeout, uncached timeout, late success, context replacement, and shutdown behavior` in the provider workstream.

The test name here is accurate as written:

```dart
  test('uses matching stored assignments with stale provider status', () async {
```

But it pairs the default 5 s budget with a mock returning an immediate `503`, so the deadline is inert. Two mutations confirm the `stale` status is over-determined: deleting `_setStatus(stale)` at `flags_repository.dart:98` leaves it green, and so does deleting the catch-branch `_setStatus(stale)` at `:129`. Re-running it with `initializationTimeout: null` and with `1 ms` both pass.

No provider test combines a warm store with an expiring deadline. The core does have one — `initialization_timeout_test.dart` `uses matching stored assignments when initialization times out` — so the gap is what the *provider* emits on that path, not the path itself.

</details>

<details>
<summary><b>Mutations that survive</b></summary>

All in `dart:3.6.2`, reverted between runs. Positive control first: making `resolveBooleanValue` always return `_notReady` kills 7 of 10, so the harness does detect breakage.

The two cases I call covered, with the mutation that kills each:

| Case | Mutation that kills it | Test |
| --- | --- | --- |
| context replacement | delete the `reconciling` event in `onContextChanged` | `keeps the previous identity active until reconciliation succeeds` |
| late success | make `_handleLateStatus` return immediately | `recovers when initial assignments arrive after the deadline` |

That same test also asserts the uncached-timeout half — `ProviderStatus.error` and the leading `error` event — but my mutation kills only the late-success half, so I have not demonstrated the timeout half is protected. Deleting the `stale` emission in `_activateRuntime` also kills a test, but it is `uses matching stored assignments with stale provider status` — the cached refresh failure, not any of the five.

| Mutation | Result |
| --- | --- |
| `_isCurrent` → `return true;` | **survives** 6/6 provider-file and 5/5 full-suite runs |
| `shutdown()` drops `_contextRevision += 1` | survives |
| `shutdown()` drops `_activeRuntime = null` | survives |
| `shutdown()` drops `_resolvedClientName = null` | survives |
| `shutdown()` drops the pending-runtime disable | survives |

`shutdown()` emptied entirely does go red, via the exposure-upload counts in `preserves Datadog evaluation telemetry without claiming tracking` at lines 334-343. Those assertions do depend on shutdown running, so shutdown is not unobserved — but nothing asserts the state it is supposed to release, which is why every mutant above survives.

For `_isCurrent`, the scenario nothing catches is three rapid contexts A→B→C where B's response lands after C's request is in flight: with the guard broken, B activates over C and evaluations serve B's assignments under C's identity. I did not execute that scenario, and reaching it needs direct unawaited provider calls — `OpenFeatureAPI` serialises its own lifecycle calls — so treat it as the untested risk the mutant exposes rather than a demonstrated bug.

</details>

<details>
<summary><b>Suggested test</b> — green on head, red when `_activeRuntime = null` is deleted</summary>

```dart
  test('shutdown releases the active runtime', () async {
    final provider = DatadogOpenFeatureProvider(
      configuration: _configuration(
        MockClient(
          (request) async => _responseFor(request, _assignmentsResponse()),
        ),
      ),
    );

    await api.setEvaluationContextAndWait(
      EvaluationContext(targetingKey: 'user-123'),
    );
    await api.setProviderAndWait(provider);
    expect(api.getClient().getBooleanValue('show-paywall', false), isTrue);

    await provider.shutdown();

    final details = provider.resolveBooleanValue(
      'show-paywall',
      false,
      EvaluationContext.empty,
    );
    expect(details.value, isFalse);
    expect(details.errorCode, ErrorCode.providerNotReady);
    expect(
      details.errorMessage,
      'The Datadog provider has not loaded assignments.',
    );
  });
```

Green on the PR head (11/11 with it added). Red when the `_activeRuntime = null;` assignment is deleted from `shutdown()`:

```
Expected: 'The Datadog provider has not loaded assignments.'
  Actual: 'Datadog flag evaluation failed with PROVIDER_NOT_READY.'
```

The `errorMessage` assertion is what makes it bite — I ran the version without it and the mutant survives 11/11. That does pin wording rather than semantics; if you would rather not couple to the string, asserting that `provider.resolveBooleanValue` no longer reaches the torn-down client some other way would be better, and I do not have a tested alternative. A cached-timeout case and a three-context case for `_isCurrent` are still worth adding; I did not write those.

</details>

packages/datadog_flags/test/datadog_openfeature_provider_test.dart
```diff
@@ -0,0 +1,538 @@
...
+    assignmentResponse.complete(
+      http.Response(jsonEncode(_assignmentsResponse()), 200),
+    );
+    await _waitUntil(() => client.providerStatus == ProviderStatus.ready);
+
+    expect(client.getBooleanValue('show-paywall', false), isTrue);
+    expect(providerEvents, [ProviderEventType.error, ProviderEventType.ready]);
+  });
+
+  test(
```
aarsilv, line 261: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** Please make this test independent of run order.

Running the file in its declared order passes 20 of 20 runs. Randomising the order fails 3 of 10 seeds, and every failure is this test, scheduled first. Alone it fails 20 of 20, at line 287 — the initial `setProviderAndWait`, before the reconciliation the test is about. The 5 ms `initializationTimeout` at line 280 has to cover the whole asynchronous initialization path, not just the mock's response.

<details>
<summary><b>Executed evidence</b> — reproducible in isolation, 20 of 20</summary>

PR head, no mutations:

| Run mode | Result |
| --- | --- |
| whole file, declared order, 20 runs | 20 pass / 0 fail |
| whole file, `--test-randomize-ordering-seed` 1-10 | 7 pass / **3 fail** |
| this test alone (`--plain-name`), 20 runs | **0 pass / 20 fail** |
| `recovers when initial assignments arrive after the deadline` alone | 10 pass / 0 fail |

Each of the three randomised failures is this test and only this test, and in each it had been scheduled first — `00:00 +0 -1: does not apply late assignments from a failed reconciliation [E]` on seeds 1, 4 and 7. That is what makes this run order rather than a slow machine: the test needs something to have run before it.

The table above is Dart 3.13.3, after installing Flutter 3.47.4. The first measurement was on `dart:3.6.2` and agreed on the two modes it covered — one whole-file run of its 10 tests passing, and 0 of 20 isolated — so this is not an artifact of the older SDK. The randomised-order runs are new here.

The isolated failure:

```
OpenFeatureException: Provider initialization ended with status error. (general)
package:openfeature_dart_client_sdk/src/api.dart 592:9  OpenFeatureAPI._initialize
===== asynchronous gap ===========================
package:openfeature_dart_client_sdk/src/api.dart 419:9  OpenFeatureAPI._setProvider
===== asynchronous gap ===========================
test/datadog_openfeature_provider_test.dart 287:7       main.<fn>
```

The other 5 ms test is fine because it *wants* the deadline to expire.

Related, and not this PR's to fix: the full suite failed 3 times in 50 counted runs on the PR head, and four more times during unrelated work here. I captured the name on only some of those, and every captured one was `initialization_timeout_test.dart: uses matching stored assignments when initialization times out`, also a 5 ms-budget test. I did not capture its failure location, so I am claiming neither the same mechanism nor that it is the only test affected. Worth a look in #1155.

</details>

<details>
<summary><b>Suggested fix</b> — verified, and still load-bearing</summary>

Two edits in this test. The budget:

```dart
          initializationTimeout: const Duration(milliseconds: 200),
```

and the settle at line 304:

```dart
      await pumpEventQueue();
```

The reconciliation still misses its deadline, because `nextContextResponse` is only completed after the assertion. 200 ms was enough in every run here; it is not a derived bound.

Measured with both edits applied together: **25 isolated runs, 25 pass / 0 fail** where it was 0 of 20; one whole-file run with all 10 of its tests passing; `dart analyze` clean. I did not measure the budget change on its own. And it still catches what it exists to catch — with the `if (!isInitialization)` early return deleted from `_handleLateStatus`, the fixed test is red:

```
Expected: ProviderStatus:<ProviderStatus.error>
  Actual: ProviderStatus:<ProviderStatus.ready>
```

</details>

packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -94,13 +107,25 @@ class DefaultDatadogFlagsClient implements DatadogFlagsClient {
...
+  }
+
   @override
   Future<void> shutdown() async {
     await Future.wait([
       _evaluationAggregator.shutdown(),
       _exposureLogger.shutdown(),
     ]);
-    await _repository.clearMemory();
+    await _repository.dispose();
```
aarsilv, line 128: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** This makes `statusChanges` go silent on a reused client, and it is worth settling whether reuse is supported at all.

`dispose()` closes the repository's broadcast controller, and `shutdown()` leaves the client in `DatadogFlags._clients`. So `sharedClient()` returns that same instance with a closed stream, and a later successful `initialize()` sets `status` to `ready` while the stream emits nothing. `_setStatus` guards on `isClosed`, so there is no error — just silence.

<details>
<summary><b>Executed evidence</b> — `status` keeps working, the stream does not</summary>

`dart:3.6.2`, PR head:

```
statuses seen before shutdown            : [ready]
sharedClient() identical after shutdown  : true
that instance's statusChanges closed     : true
after a successful re-initialize         : statuses seen = []   status = ready
evaluation after re-initialize           : works
```

Reverting `shutdown()` to `_repository.clearMemory()` gives `closed = false` and the re-initialize emits `[ready]`; `dart analyze` clean and the package suite green over 6 consecutive runs.

`datadog_flags_flutter` resolves its delegate with `_flags.sharedClient(name: name)` (`datadog_flags_plugin.dart:96`) and clears `_delegate` on shutdown, so it re-resolves that same instance. I can measure that half now. Driving `DatadogFlutterFlagsClient` with a delegate that closes its status controller in `shutdown()`, which is what `FlagsRepository.dispose()` does:

```
app listener before shutdown : [DatadogFlagsClientStatus.ready]
core stream closed           : true
app listener after reuse     : []
client.status after reuse    : DatadogFlagsClientStatus.ready
```

The wrapper does re-subscribe on re-resolve, but to a stream that is already closed, so the silence reaches Flutter app listeners unchanged.

</details>

<details>
<summary><b>Reverting one line is only a partial fix</b></summary>

Reuse is already broken in a second way that predates this PR. `EvaluationAggregator` creates its periodic flush timer once, in the constructor (`evaluation_aggregator.dart:37`), and `shutdown()` cancels and nulls it (`:96-97`) with nothing recreating it. Measured with `trackEvaluations: true` and a 1 s interval:

```
flagevaluation posts before shutdown, 1.3 s wait : 1
same client instance returned after shutdown     : true
flagevaluation posts after reuse, 1.5 s wait     : 0
```

So a reused client evaluates flags and reports `status` correctly while its periodic evaluation flush never runs again. Recorded evaluations are not lost — a later explicit `flush()` or `shutdown()` would still send them — but nothing sends them on the interval the configuration asks for. That is not something this PR introduced, but it means "revert to `clearMemory()`" restores status events without making reuse healthy.

</details>

<details>
<summary><b>Do not fix this by closing the other two controllers</b></summary>

I checked what happens if the provider is re-registered, because closing `_events` in `shutdown()` looks like the symmetric change. Reuse works today:

```
register first -> register second (API disposes first) -> re-register first
  => works, providerStatus = ready, first's value returns
```

Adding `await _events.close();` to `provider.shutdown()` breaks exactly that:

```
Bad state: Cannot add new events after calling close
```

and the package suite still passed, so nothing would have caught it. The same argument applies to `DatadogFlutterFlagsClient._statusChanges` at `datadog_flags_plugin.dart:172`, which I have now measured rather than argued. Adding `await _statusChanges.close();` to that class's `shutdown()`:

```
Bad state: Cannot add new events after calling close
```

raised on the re-subscribe after reuse — and that package's suite still reported `00:00 +11: All tests passed!`, so nothing there would have caught it either. Provider reuse works because each registration builds a fresh `DatadogFlags` owner; core-client reuse is the case that does not. Whether provider reuse is *intended* is your call — I only measured that it currently works.

</details>

<details>
<summary><b>Two ways to settle it</b></summary>

**Make `shutdown()` terminal and say so.** Keep `dispose()`, document on `DatadogFlagsClient.shutdown()` that the client is spent and callers must go through `DatadogFlags.disable()` / `enable()`, and have `DatadogFlags` drop the client from `_clients` so `sharedClient()` returns a fresh one. That also fixes the aggregator case. It needs a back-reference or a `DatadogFlags`-side method, so I did not write it.

**Or keep reuse working for status**, which is the one-line change I did test:

```dart
    await _repository.clearMemory();
```

`FlagsRepository.dispose()` then has no callers, so delete it:

```dart
  Future<void> dispose() async {
    await clearMemory();
    if (!_statusChanges.isClosed) {
      await _statusChanges.close();
    }
  }
```

I would take the first — the second leaves periodic evaluation flushing stopped on reuse, so it fixes the symptom I can see rather than the contract.

</details>

packages/datadog_flags/lib/src/datadog_openfeature_provider.dart
```diff
@@ -0,0 +1,508 @@
...
+    if (identical(_activeRuntime, runtime)) {
+      if (status == datadog.DatadogFlagsClientStatus.ready) {
+        _events.add(
+          openfeature.ProviderEvent(type: openfeature.ProviderEventType.ready),
+        );
+      }
+      return;
+    }
+
+    if (!isInitialization) {
```
aarsilv, line 329: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P3]** Please say in a comment why reconciliation drops a late success when initialization keeps it.

This early return discards a context change whose assignments arrive after the deadline. The provider stays in `error`, serving the previous subject's assignments, with no recovery in the window I observed. Your test `does not apply late assignments from a failed reconciliation` pins that, so I read it as deliberate. The asymmetry with `initialize` is what a future reader will try to "fix".

<details>
<summary><b>Executed evidence</b> — what the asymmetry produces</summary>

`initializationTimeout: 50 ms`, first context `200`, second context held past the deadline and then completed with `200`. `dart:3.6.2`:

```
reconcile threw            : yes
providerStatus right after : error
events                     : [ready, reconciling, error]
after the second context's assignments arrive 200:
  providerStatus : error      (unchanged)
  value          : true       (= user-first)
  precompute requests : 2
```

Deleting this early return makes the same scenario reach `providerStatus = ready`, which confirms it is the deciding line rather than something downstream. That is one observed late completion over roughly 300 ms, so I am not claiming how or whether the provider recovers afterwards — only that it did not here.

For contrast, the initialization path does recover: `recovers when initial assignments arrive after the deadline` asserts `[error, ready]`.

One caveat on the pinning test: it fails 20/20 when run in isolation on the PR head, which I raise separately, so treat it as pinning the intent rather than as reliable protection.

</details>

<details>
<summary><b>Suggested comment</b></summary>

I deliberately have not written the rationale for you, because I could not derive it from the code and guessing it here would put words in your mouth. What the comment needs to state is the behaviour plus the reason:

```dart
    // Reconciliation does not activate a late success: after this load has
    // already reported an error event, the candidate is discarded rather than
    // swapped in. The initialization path does activate late. Reason: <why>.
    if (!isInitialization) {
```

Two things I checked while trying to infer the reason, in case they help. `isInitialization` does not imply `_activeRuntime == null`. And the switch below hardcodes `isInitialization: true` at line 344, so if the branch were removed a late reconciliation success would emit `ready` rather than `contextChanged` (lines 370-375) — which may itself be part of why the branch exists. If a caller is expected to recover by setting the context again, saying so in the class doc would help too.

</details>

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -152,24 +165,41 @@ class FlagsRepository {
...
+    }
   }
 
   Future<void> reset() async {
     await clearMemory();
     await _deleteCached();
   }
 
+  void _setStatus(DatadogFlagsClientStatus status) {
+    if (_status == status) {
```
aarsilv, line 184: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P3]** Nothing pins this dedup, and `statusChanges` became public API in this PR.

Deleting the early return keeps the full suite green. It does change what the core's public stream emits: the cached-plus-failed-refresh path calls `_setStatus(stale)` twice, at line 98 and again at line 129. The README section this PR adds describes those emissions, so a duplicate would read as a bug to anyone following it.

<details>
<summary><b>Executed evidence, including the part that corrects me</b></summary>

`dart:3.6.2`, cached assignments plus a `503` refresh:

| Tree | core `statusChanges` | provider events |
| --- | --- | --- |
| PR head | `[stale]` | `[ready, stale]` |
| dedup deleted | `[stale, stale]` | `[ready, stale]` |

I had assumed the duplicate would reach OpenFeature consumers. It does not, and that is worth stating plainly: during `client.initialize()` the provider's `initializationReturned` flag is still false, so its subscription ignores both emissions and it reads `lifecycle.status` once at completion. I could not construct any path where consecutive duplicate statuses reach provider events. So this is about the core stream only, which is why it is `[P3]`. `README.md:299` says `statusChanges` emits later transitions; it does not promise exact-once, so treat this as tightening a documented behaviour rather than restoring a stated guarantee.

Deleting the early return survives the full suite over 3 runs.

</details>

<details>
<summary><b>Suggested test</b> — verified green, and red without the dedup</summary>

```dart
  test('reports stored assignments as stale exactly once when refresh fails',
      () async {
    const context = FlagsEvaluationContext(targetingKey: 'user-123');
    final store = InMemoryDatadogFlagsStore();
    final datadogFlags = DatadogFlags();
    addTearDown(datadogFlags.disable);
    await datadogFlags.enable(
      configuration: DatadogFlagsConfiguration(
        datadogConfig: _datadogConfig(),
        httpClient: _clientWithResponse([], _assignmentsResponse()),
        store: store,
      ),
    );
    await datadogFlags.sharedClient().initialize(context);
    await datadogFlags.disable();

    await datadogFlags.enable(
      configuration: DatadogFlagsConfiguration(
        datadogConfig: _datadogConfig(),
        httpClient: MockClient((_) async => http.Response('unavailable', 503)),
        store: store,
      ),
    );
    final restored = datadogFlags.sharedClient();
    final lifecycle = restored as DatadogFlagsClientLifecycle;
    final statusChanges = <DatadogFlagsClientStatus>[];
    final subscription = lifecycle.statusChanges.listen(statusChanges.add);
    addTearDown(subscription.cancel);

    await restored.initialize(context);

    expect(statusChanges, [DatadogFlagsClientStatus.stale]);
    expect(
      restored
          .getBooleanDetails(key: 'show-paywall', defaultValue: false)
          .value,
      isTrue,
    );
  });
```

Added to `test/datadog_flags_test.dart`, which already has `_datadogConfig`, `_clientWithResponse` and `_assignmentsResponse`. Green on the PR head; red with the dedup deleted:

```
Expected: [DatadogFlagsClientStatus:DatadogFlagsClientStatus.stale]
  Actual: [
            DatadogFlagsClientStatus:DatadogFlagsClientStatus.stale,
            DatadogFlagsClientStatus:DatadogFlagsClientStatus.stale
          ]
   Which: at location [1] ... which longer than expected
```

`dart analyze` clean, full suite green with it added.

</details>

packages/datadog_flags/lib/src/default_flags_client.dart
```diff
@@ -200,6 +227,11 @@ class DefaultDatadogFlagsClient implements DatadogFlagsClient {
       value: resolvedValue as T,
       variant: assignment.variationKey,
       reason: assignment.reason,
+      flagMetadata: {
+        'datadog.allocation_key': assignment.allocationKey,
```
aarsilv, line 231: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P3]** These keys are literals here but public constants on the provider, and they diverge from Android.

`DatadogOpenFeatureProvider` declares `allocationKeyMetadata` and `serialIdMetadata` for exactly these two strings, so the values are written twice. Android also names the allocation key differently, so a customer reading `flagMetadata` gets a different key per platform. The provider name differs too, though that is `ProviderMetadata`, not flag metadata.

<details>
<summary><b>What Android uses</b> — checkout dated 2026-09-11</summary>

| | Flutter, this PR | Android |
| --- | --- | --- |
| provider name | `'Datadog'` | `"Datadog Feature Flags Provider"` (`DatadogFlagsProvider.kt:247`) |
| allocation key | `datadog.allocation_key` | `allocationKey`, unprefixed (`DatadogFlagsClient.kt:413`) |
| extra logging | not forwarded | each `precomputedFlag.extraLogging` entry whose value is a `String`, `Number` or `Boolean` is copied into metadata (`DatadogFlagsClient.kt:406-411`) |

One precision, because it would be easy to overstate: Android does have `serialId` — it is a model field and it goes into exposure events (`ExposureEventsProcessor.kt:27,41,68`, `PrecomputedFlagConstants.kt:27`). It is simply not placed in `flagMetadata`. So `datadog.serial_id` is a Flutter addition rather than a Flutter-only concept.

No `.swift` file in my `dd-sdk-ios` checkout mentions OpenFeature, so I had only Android to compare against.

</details>

<details>
<summary><b>Suggested fix</b> — verified for the duplication half</summary>

The literals are the part I can fix without a cross-SDK decision. In `flags_client.dart`, next to `FlagDetails`:

```dart
/// Flag metadata key carrying the Datadog allocation key.
const datadogAllocationKeyMetadata = 'datadog.allocation_key';

/// Flag metadata key carrying the Datadog assignment serial ID.
const datadogSerialIdMetadata = 'datadog.serial_id';
```

Here:

```dart
      flagMetadata: {
        datadogAllocationKeyMetadata: assignment.allocationKey,
        if (assignment.serialId case final serialId?)
          datadogSerialIdMetadata: serialId,
      },
```

And the provider's public constants become aliases:

```dart
  /// OpenFeature flag metadata key containing the Datadog allocation key.
  static const allocationKeyMetadata = datadog.datadogAllocationKeyMetadata;

  /// OpenFeature flag metadata key containing the Datadog assignment serial ID.
  static const serialIdMetadata = datadog.datadogSerialIdMetadata;
```

`default_flags_client.dart` already imports `flags_client.dart`, so this adds no dependency edge. I applied all three edits: `dart analyze` clean and the `datadog_flags` package suite green.

The names and the `extraLogging` gap are a cross-SDK call, not mine — but they are cheaper to settle now than after the provider ships.

</details>

packages/datadog_flags/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Changelog
 
-## 1.1.0
+## 1.2.0-beta.1
```
aarsilv, line 3: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P3]** This file is the only one that conflicts with the base branch.

You agreed on #1155 to leave changelog entries to the releaser, and `fef293e` removed the `## 1.1.0` section there. This branch diverged before that commit, so it still carries that section's entry, renames the heading to `## 1.2.0-beta.1`, and adds three more. Raising it only because the mechanical consequence is concrete.

<details>
<summary><b>The conflict, and exactly what differs</b></summary>

`git merge-tree --write-tree` against base head `5fdccc9` reports one conflict:

```
CONFLICT (content): Merge conflict in packages/datadog_flags/CHANGELOG.md
```

`flags_repository.dart`, `initialization_timeout_test.dart` and both READMEs auto-merge cleanly.

The three sides:

| Tree | Top section |
| --- | --- |
| merge base `9b60dc4` | `## 1.1.0` + the timeout entry |
| this PR `c4bb07b` | `## 1.2.0-beta.1` + that entry and three new ones |
| base head `5fdccc9` | section deleted by `fef293e` |

So the heading is part of the conflict, not just the bullets. Given the releaser-owns-the-changelog policy, dropping the whole section is the resolution I would recommend; git will accept other resolutions, so this is policy rather than mechanics.

To be clear about what this does and does not unblock: it clears the only conflicting file. The PR is still a draft, which is its own gate on merging.

</details>

<details>
<summary><b>One thing worth keeping separate</b></summary>

The `version: 1.2.0-beta.1` bump in `pubspec.yaml` is a different decision from the changelog text, and it is the part that makes this a publishable state — see my other comment on the publish gate. Dropping the changelog section does not address that.

For reference, #1155's thread:

> Let releaser add changelog items (I promise we'll get a git hook in for this soon 😅)

and your reply:

> Understood; I'll add that note to the contribution or agents file in the other PR to avoid this.

</details>

examples/simple_example/lib/flags/flags_example_config.dart
```diff
@@ -180,6 +181,18 @@ DatadogSite _datadogSiteForName(String? siteName) {
   }
 }
 
+DatadogFlagsSite? _datadogFlagsSiteFor(DatadogSite site) {
```
aarsilv, line 184: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P3]** This duplicates the public `datadogFlagsSiteFor`, but the reuse path may not be one you want.

`datadog_flags_flutter` exports that exact function, and the body here is character-identical. The catch: the mapping takes `DatadogSite` from `datadog_flutter_plugin`, so it cannot move into pure-Dart `datadog_flags`. Reusing it means keeping a `datadog_flags_flutter` dependency in the example — the one this change drops.

<details>
<summary><b>Why it cannot simply move</b></summary>

`datadog_flags_flutter/lib/datadog_flags_flutter.dart` ends with an unfiltered `export 'src/datadog_flags_plugin.dart';`, and the function sits at `datadog_flags_plugin.dart:342`, so it is public API today.

But `packages/datadog_flags/pubspec.yaml` depends only on `http`, `json_annotation`, `meta`, `openfeature_dart_client_sdk` and `uuid`. It is pure Dart by design and has no way to name `DatadogSite`. Any `DatadogSite → DatadogFlagsSite` mapping has to live somewhere that depends on both. `datadog_flags_flutter` is the one published package that does; this example is the other place, which is exactly why the copy ended up here (it imports both at `flags_example_config.dart:8-9`).

</details>

<details>
<summary><b>Two options, and which I would take</b></summary>

**Keep the copy, and say it is a copy.** Two comment lines, and the example stays free of `datadog_flags_flutter`:

```dart
// Mirrors datadogFlagsSiteFor in datadog_flags_flutter. Kept local so this
// example does not depend on datadog_flags_flutter. Update both together.
DatadogFlagsSite? _datadogFlagsSiteFor(DatadogSite site) {
```

**Or skip `DatadogSite` on the flags path**, mapping the configured site name straight to `DatadogFlagsSite` so there is no second enum translation to keep in sync at all.

I would take the first — it is honest about the duplication and preserves the point of the change. I have now applied it with Flutter 3.47.4 installed: in `examples/simple_example`, `dart analyze .` gives `No issues found!` and `dart format --output=none --set-exit-if-changed .` exits `0`. The example also builds: `flutter build web` gives `✓ Built build/web`. Analyze needs a `.env` present or it reports `The asset file '.env' doesn't exist`; the README points developers at `../../generate_env.sh` for that, and I wrote one by hand to the same shape.

That matters because this example sits on a developer footing rather than a CI one: `melos.yaml`'s `packages:` globs — `packages/*`, `packages/*/example`, `packages/*/integration_test_app`, `packages/*/e2e_test_app` — do not match `examples/simple_example`, so no melos script reaches it, `generate_env` included. Nothing will catch the two copies drifting apart, either: a changed mapping stays compilable in both places and the example does not assert on it. That is the argument for the comment, or for the second option above, which removes the second copy rather than annotating it.

</details>

<details>
<summary><b>Dropping the dependency also drops the only RUM sample</b></summary>

At this head, `git grep` finds no reference to `datadog_flags_flutter` or `DatadogFlagsPlugin` anywhere under `examples/simple_example`. At the base branch it matches `main.dart:77`, `flags_example_config.dart:8`, `flags_screen.dart:9`, `pubspec.yaml` and `README.md:9,26`.

What the plugin uniquely provides is the automatic bridge from an evaluation to RUM, and it spans both classes: `DatadogFlagsPlugin._addRumFeatureFlagEvaluation` calls `instance.rum?.addFeatureFlagEvaluation(key, value)` at `:162-164`, the plugin injects that callback into the client, and `DatadogFlutterFlagsClient._trackRumEvaluation` invokes it at `:321`. The RUM call itself is public (`ddrum.dart:560`), so the example could still do it by hand — as written it does neither. This is the only Flutter sample for either package; `packages/datadog_flags/example` is a pure-Dart CLI under `bin/` and has no RUM to reach.

That may be exactly right for a sample whose point is OpenFeature. I am flagging it as a choice to make deliberately rather than as a defect.

</details>

packages/datadog_flags_flutter/lib/src/datadog_flags_plugin.dart
```diff
@@ -282,6 +308,9 @@ class DatadogFlutterFlagsClient implements DatadogFlagsClient {
       return existing;
     }
     final resolved = await _resolveDelegate();
+    if (resolved case final DatadogFlagsClientLifecycle lifecycle) {
+      _statusSubscription = lifecycle.statusChanges.listen(_statusChanges.add);
```
aarsilv, line 312: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** This subscribe can run twice, or after `shutdown()`, because nothing serialises the resolve.

`_delegate` is assigned only after the `await` at `:310`, so two callers both clear the `_delegate == null` check at `:306-309`. `initialize()` (`:213`) and `reset()` (`:289`) are both public and both route here. Two subscriptions then feed one controller, and `:300` cancels only the latest. A `shutdown()` arriving during a resolve is worse: it returns at `:297` having done nothing, and the resolve subscribes after it.

<details>
<summary><b>Executed evidence</b> — two named tests, red here, green on the fix</summary>

Flutter 3.47.4, both tests below added to `datadog_flags_plugin_test.dart`, delegate resolution held on a completer so the calls overlap:

```
00:00 +6 -1: resolves the delegate once when lifecycle calls overlap [E]
  Expected: <1>
    Actual: <2>
00:00 +6 -2: shutdown during an in-flight resolve leaves nothing subscribed [E]
  Expected: empty
    Actual: [DatadogFlagsClientStatus:DatadogFlagsClientStatus.ready]
00:00 +11 -2: Some tests failed.
```

Printing rather than asserting, on the first case, shows what an app listener receives — one upstream status delivered twice, and a third delivered after the client was shut down:

```
RESOLVE_COUNT=2
EVENTS=[ready, ready]
EVENTS_AFTER_SHUTDOWN=[ready, ready, stale]
```

And on the second case, `verify(() => delegate.shutdown())` reports `No matching calls` on this head: a `shutdown()` that lands during a resolve does not reach the delegate either.

With the fix below: `00:00 +13: All tests passed!` (the 11 already here plus these two), `dart analyze .` `No issues found!`, and `dart format --output=none --set-exit-if-changed .` exits `0`. The 11 existing tests pass either way, which is why nothing catches this today.

To be fair about attribution: the missing serialisation predates this PR. At the base the method body was just `final resolved = await _resolveDelegate(); _delegate = resolved;` (`datadog_flags_plugin.dart:279-287` there), so a double resolve assigned the same instance twice and left no per-resolve state behind — the concurrent `initialize`/`reset` on the delegate was already possible, but nothing accumulated. The subscription added at `:312` is what turns it into duplicate delivery, and the cancel added at `:300` is what makes `shutdown()` incomplete rather than absent.

This is a different defect from the `statusChanges` point I raised on `default_flags_client.dart:128`: that one is the core client going silent on reuse, this one is the wrapper delivering twice and outliving its own `shutdown()`.

</details>

<details>
<summary><b>Suggested fix</b> — memoize the resolve, and make `shutdown()` drain it</summary>

A field beside `_statusSubscription` (`:176`):

```dart
  StreamSubscription<DatadogFlagsClientStatus>? _statusSubscription;
  Future<DatadogFlagsClient>? _pendingResolve;
```

The resolve itself becomes the thing that is shared:

```dart
  Future<DatadogFlagsClient> _delegateOrResolve() {
    final existing = _delegate;
    if (existing != null) {
      return Future.value(existing);
    }
    return _pendingResolve ??= _resolveDelegate().then((resolved) {
      if (resolved case final DatadogFlagsClientLifecycle lifecycle) {
        _statusSubscription =
            lifecycle.statusChanges.listen(_statusChanges.add);
      }
      _delegate = resolved;
      return resolved;
    }).whenComplete(() => _pendingResolve = null);
  }
```

And `shutdown()` has to wait for it, or it still returns early and lets the late resolve subscribe behind its back:

```dart
  Future<void> shutdown() async {
    final pending = _pendingResolve;
    if (pending != null) {
      try {
        await pending;
      } catch (_) {
        // A resolve that failed left no delegate and no subscription.
      }
    }
    final delegate = _delegate;
    if (delegate == null) {
      return;
    }
    await delegate.shutdown();
    await _statusSubscription?.cancel();
    _statusSubscription = null;
    _delegate = null;
  }
```

Three things worth knowing, each of which I got wrong first and found by running it:

- Memoizing alone is **not** enough. With only the `_pendingResolve` change, the second test still fails identically — `[ready]` after shutdown, and `delegate.shutdown()` never called. The `shutdown()` change is the load-bearing half.
- `whenComplete` clears the slot on failure too, so a delegate that throws does not wedge every later call on a rejected future — hence the `try`/`catch` in `shutdown()` rather than a bare `await`.
- **This makes `shutdown()` wait on `_resolveDelegate()`, and I cannot tell you that wait is bounded.** My first probe deadlocked for the full 30 s timeout because it awaited `shutdown()` before completing the resolve. The real callback is `() async { await ready; return _flags.sharedClient(name: name); }` (`:94-97`), so `shutdown()` would now block on the plugin's `ready` future first. A *failing* resolve is fine — the `catch` handles it, and I measured that: with the resolve completed as an error, `initialize()` threw `Bad state: ready failed` and `shutdown()` still returned. The case that hangs is a `ready` future that never settles at all. That is the cost of this version, and whether to accept it is your call.

  I am not proposing the obvious alternative, because I could not make it correct in one line. A `_shuttingDown` flag checked before subscribing would stop the subscription but would still leave the late delegate running and unshut, and it would not say what the in-flight `initialize()` should do once `shutdown()` has overtaken it. That is a contract decision rather than a patch, so I would rather raise it than guess at it.

The wrap on the `_statusSubscription =` line is what `dart format` produces once the statement sits inside the `.then` closure — the extra indent pushes it past 80 columns. It is unwrapped at `:312` today because it is one level shallower.

</details>

<details>
<summary><b>The two tests</b></summary>

`dart:async`, `MockDatadogFlagsLifecycleClient` and `DatadogFlags.defaultClientName` are all already in this file, so they need no new imports:

```dart
  test('resolves the delegate once when lifecycle calls overlap', () async {
    const context = FlagsEvaluationContext(targetingKey: 'user-1');
    final delegate = MockDatadogFlagsLifecycleClient();
    final delegateStatusChanges =
        StreamController<DatadogFlagsClientStatus>.broadcast(sync: true);
    addTearDown(delegateStatusChanges.close);
    when(() => delegate.name).thenReturn(DatadogFlags.defaultClientName);
    when(() => delegate.initialize(context)).thenAnswer((_) async {});
    when(() => delegate.reset()).thenAnswer((_) async {});
    when(() => delegate.shutdown()).thenAnswer((_) async {});
    when(() => delegate.statusChanges).thenAnswer(
      (_) => delegateStatusChanges.stream,
    );

    final resolveGate = Completer<void>();
    var resolveCount = 0;
    final client = DatadogFlutterFlagsClient(
      name: 'default',
      resolveDelegate: () async {
        resolveCount++;
        await resolveGate.future;
        return delegate;
      },
      addRumFeatureFlagEvaluation: null,
    );
    final statusChanges = <DatadogFlagsClientStatus>[];
    final statusSubscription = client.statusChanges.listen(statusChanges.add);
    addTearDown(statusSubscription.cancel);

    final initializing = client.initialize(context);
    final resetting = client.reset();
    resolveGate.complete();
    await Future.wait([initializing, resetting]);

    delegateStatusChanges.add(DatadogFlagsClientStatus.ready);
    await pumpEventQueue();

    expect(resolveCount, 1);
    expect(statusChanges, [DatadogFlagsClientStatus.ready]);

    await client.shutdown();
    delegateStatusChanges.add(DatadogFlagsClientStatus.stale);
    await pumpEventQueue();

    expect(statusChanges, [DatadogFlagsClientStatus.ready]);
  });

  test('shutdown during an in-flight resolve leaves nothing subscribed',
      () async {
    const context = FlagsEvaluationContext(targetingKey: 'user-1');
    final delegate = MockDatadogFlagsLifecycleClient();
    final delegateStatusChanges =
        StreamController<DatadogFlagsClientStatus>.broadcast(sync: true);
    addTearDown(delegateStatusChanges.close);
    when(() => delegate.name).thenReturn(DatadogFlags.defaultClientName);
    when(() => delegate.initialize(context)).thenAnswer((_) async {});
    when(() => delegate.shutdown()).thenAnswer((_) async {});
    when(() => delegate.statusChanges).thenAnswer(
      (_) => delegateStatusChanges.stream,
    );

    final resolveGate = Completer<void>();
    final client = DatadogFlutterFlagsClient(
      name: 'default',
      resolveDelegate: () async {
        await resolveGate.future;
        return delegate;
      },
      addRumFeatureFlagEvaluation: null,
    );
    final statusChanges = <DatadogFlagsClientStatus>[];
    final statusSubscription = client.statusChanges.listen(statusChanges.add);
    addTearDown(statusSubscription.cancel);

    final initializing = client.initialize(context);
    final shuttingDown = client.shutdown();
    resolveGate.complete();
    await Future.wait([initializing, shuttingDown]);

    delegateStatusChanges.add(DatadogFlagsClientStatus.ready);
    await pumpEventQueue();

    expect(statusChanges, isEmpty);
    verify(() => delegate.shutdown()).called(1);
  });
```

</details>

PR #1136, opened 2026-08-28
chore(android): Pin `dd-sdk-android` to 3.13.1

gonzalezreal:
### What and why?

Pin all Android SDK dependencies to `3.13.1`. This is required for hybrid Session Replay support in Android. See #1129

### How?

Updates all six `ext.datadog_version` declarations, including the app-level declarations not covered by the pinning helper.

### Review checklist

### DIFF
--- a/packages/datadog_inappwebview_tracking/android/build.gradle
+++ b/packages/datadog_inappwebview_tracking/android/build.gradle
@@ -6,7 +6,7 @@ version = "1.0-SNAPSHOT"
 
 buildscript {
     ext.kotlin_version = "2.2.20"
-    ext.datadog_version = "3.0.0"
+    ext.datadog_version = "3.13.1"
 
     repositories {
         google()
--- a/packages/datadog_inappwebview_tracking/android/build.gradle
+++ b/packages/datadog_inappwebview_tracking/android/build.gradle
@@ -6,7 +6,7 @@ version = "1.0-SNAPSHOT"
 
 buildscript {
     ext.kotlin_version = "2.2.20"
-    ext.datadog_version = "3.13.1"
+    ext.datadog_version = "3.0.0"
 
     repositories {
         google()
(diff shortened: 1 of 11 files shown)


kikoveiga (dismissed):

packages/datadog_inappwebview_tracking/android/build.gradle
```diff
@@ -6,7 +6,7 @@ version = "1.0-SNAPSHOT"
 
 buildscript {
     ext.kotlin_version = "2.2.20"
-    ext.datadog_version = "3.0.0"
+    ext.datadog_version = "3.13.1"
```
fuzzybinary: We shouldn't update the version specified in some of the side packages. If the script did this, I should probably adjust it so it doesn't do this.

packages/datadog_webview_tracking/android/build.gradle
```diff
@@ -6,7 +6,7 @@ version '1.0-SNAPSHOT'
 
 buildscript {
     ext.kotlin_version = '2.2.20'
-    ext.datadog_version = "3.0.0"
+    ext.datadog_version = "3.13.1"
```
fuzzybinary: This is the other package that shouldn't be updated.

fuzzybinary (changes requested):

packages/datadog_inappwebview_tracking/android/build.gradle
gonzalezreal: The pinning script updated both WebView packages. The manual changes only covered the app-level examples. I’ve reverted both WebView packages to 3.0.0.

fuzzybinary (dismissed):

packages/datadog_inappwebview_tracking/android/build.gradle
fuzzybinary: 👍 I'll fix the pinning script at some point.

kikoveiga (dismissed):

kikoveiga (approved):

fuzzybinary (approved):

Issue #1133, opened 2026-08-26
[Android] SIGSEGV during RUM and Logs event mapper registration


dfdgsdfg:
### Stack trace

Production Android releases crash during `DatadogSdk.initialize` while registering RUM or Logs event mappers through the generated JNI bridge.

Native stack sampled independently by Google Play:

    signal 11 (SIGSEGV), SEGV_MAPERR, null-address access
    libart.so  art::InvokeVirtualOrInterfaceWithVarArgs
    libart.so  art::JNI<false>::CallVoidMethodV
    libdartjni.so  globalEnv_CallVoidMethod+224
    libapp.so  stub CallNativeThroughSafepoint

Symbolized Dart stacks from Crashlytics continue from `CallNativeThroughSafepoint` into one of these paths:

    DatadogRumPlugin$Companion._setRumEventMapper.#ffiClosure10
    DatadogRumPlugin$Companion.setRumEventMapper
    AndroidRumEventMapper
    DdRumMethodChannel.enable
    DatadogRum.enable
    DatadogSdk.initialize

or:

    DatadogLogsPlugin$Companion._setLogsEventMapper.#ffiClosure10
    DatadogLogsPlugin$Companion.setLogsEventMapper
    AndroidLogEventMapper
    DdLogsMethodChannel.enable
    DatadogLogging.enable
    DatadogSdk.initialize

This is related in native shape to #942, but our setup does not use Session Replay. The recovered Dart frames consistently select mapper registration during SDK initialization rather than widget-tree capture.

### Reproduction steps

No deterministic local reproduction yet. Production setup is a normal Flutter application initialized once during bootstrap:

    final configuration = DatadogConfiguration(
      clientToken: clientToken,
      env: environment,
      service: service,
      nativeCrashReportEnabled: true,
      loggingConfiguration: DatadogLoggingConfiguration(),
      rumConfiguration: DatadogRumConfiguration(
        applicationId: applicationId,
        traceSampleRate: traceSampleRate,
        sessionSamplingRate: sessionSamplingRate,
        telemetrySampleRate: telemetrySampleRate,
        actionEventMapper: (_) => null,
      ),
      site: DatadogSite.us5,
      firstPartyHosts: firstPartyHosts,
    );
    await DatadogSdk.instance.initialize(
      configuration,
      TrackingConsent.granted,
    );

Session Replay is not enabled and `datadog_session_replay` is not in the dependency graph.

[...]

PR #1140, opened 2026-09-01
fix(android): Prevent retaining a potentially GC'd Companion object

fuzzybinary:
### What and why?

Using `Companion` in `setLogsEventMapper` could potentially hold on to the Companion object in Dart in such a way that it could be GC'd, which would cause a SIGSEV crash.

Use the `@JvmStatic` call instead, which won't retain the Companion object, preventing the crash.

refs: #1133

### Review checklist


fuzzybinary:
@codex review

kikoveiga (approved):

PR #1144, opened 2026-09-02
V4 regenerate jni bindings

dballance:
### What and why?

Current v4 Bindings are generated using 0.16.x JNIGen. 

There is a class of crashes caused by JNI bindings being reaped when `reference.pointer` is used directly. Basically, it doesn't keep around the Dart `JObject` wrapper and it can be reaped by the GC causing crashes. 

We've experienced similar crashes caused by `dart-http/ok_http` (https://pub.dev/packages/ok_http) and are trying to get it updated. This plugin on v4 uses `jni: ^1.0.0` but still had bindings generated from 0.16.x of JNIGen which suffers from the same issue. This was resolved in 0.17.x of JNIGen.

Basically, you need to do this:

```dart
// Before - will crash if the underlying native resource is reaped by GC
_nativeCall(reference.pointer, ...);

// Fixed - _$$selfRef will remain live for the entire call expression, so the JNI reference can't be finalized and reaped before the method call is complete. There are other methods constructors / static methods.
final _$$selfRef = reference;
_nativeCall(_$$selfRef.pointer, ...);
```

See: 
https://github.com/dart-lang/native/issues/3342
https://github.com/dart-lang/native/pull/3350

### How?

Ran the below to regenerate and validate.

```bash
dart run jnigen --config jnigen.yaml
dart format lib/src/datadog_android_bridge.dart

flutter analyze
flutter test -r failures-only

cd example
flutter pub get
flutter build apk --release
```

There were some pubspec files updated, but I put those in a separate commit. Can probably just drop `8da977bfdd9d3a5a603b2672f7c9ab59b3b83d3b` if it's not desired here.

### Review checklist

- [X] This pull request has appropriate unit and / or integration tests 

Don't have an issue for this - happy to file one if it is needed.

Thanks!


dballance:
Also, if it wasn't clear, I opened this on the V4 branch. 

JNI was upgraded in V4 to 1.0.0 and JNIGen to 0.17.x, which is the correct version. Just wanted to make sure this regeneration fix landed in a v4 release.

fuzzybinary:
Hi @dballance ! Thanks for pointing this out!

I am going to close this though in favor of #1147 which is going to add CI verification as well, to ensure this doesn't happen again. I'd added this for the new desktop platform, but neglected iOS and Android. The new PR closes that gap.

Thanks again!

PR #1146, opened 2026-09-03
chore: pin iOS SDK to 3.16.0

gonzalezreal:
### What and why?

The example apps currently fetch `dd-sdk-ios` from `develop`, so unrelated native changes can break Flutter CI.
This PR pins the iOS SDK to the released version `3.16.0`.

### How?

Updates the CocoaPods overrides and Swift Package Manager manifests for the core and Session Replay packages.
The integration packages remain unchanged.

### Review checklist


fuzzybinary (approved):
LGTM - are we sure none of the other examples require modifying the `Datadog Pod Overrides`?

PR #1148, opened 2026-09-04
chore: upgrade jnigen

Tiska:
### What and why?

Upgrade jnigen to 1.0.0 to be compatible with latest flutter releases and ecosystem packages.


fuzzybinary:
Hi @Tiska 

Sorry, we can't do this in v3 - the requirement for a higher version of Dart / Flutter (3.9 / 3.35) we consider a breaking change.

v4 preview should be out soon which upgrades both `objective-c` and `jni` to latest.

Sorry for the inconvenience!

Issue #1127, opened 2026-08-23
Wasm compilation fail targeting web


scognito:
### Describe the bug

`DdRumWeb.initialize` in `lib/src/rum/web/ddrum_web.dart` converts a plain Dart closure (`String` → `bool`) to a JS function via `Function.toJS`. Since the
  closure's parameter/return types are not JS-interop types, this trips `dart2wasm`'s `--wasm-dry-run` interop check. Instead of emitting a normal           
  diagnostic, the compiler crashes with an unhandled `RangeError` while trying to report the location of the offending call.                                 
                                                                                                                                                             
  This happens with `flutter build web` / `flutter run -d chrome` even when **not** targeting `--wasm`, because Flutter runs the wasm dry-run as an          
  informational check by default.                                                                                                                            
                                                                                                                                                             
  ### Offending code                                                                                                                                         
                                                                                                                                                             
  `lib/src/rum/web/ddrum_web.dart` (unchanged from 3.3.0 through the latest 3.5.0):   

  ```dart                                                                                                                                                    
  allowedTracingUrls: [                                                                                                                                      
    for (final host in sanitizedFirstPartyHosts)                                                                                                             
      _TracingUrl(                                                                                                                                           
        match: ((String check) {                                                                                                                             
          final uri = Uri.parse(check);                                                                                                                      
          return host.regExp.hasMatch(uri.host);                                                                                                             
        }).toJS,                                                                                                                                             
        propagatorTypes: host.headerTypes                                                                                                                    
            .map(_headerTypeToPropagatorType)                                                                                                                
            .toList()                                                                                                                                        
            .toJS,                                                                                                                                           
      ),                                                                                        
  ].toJS,
  
  match is typed as external JSFunction match; on _TracingUrl (an @anonymous extension type over JSObject). Converting a bool Function(String) Dart closure
  directly via .toJS is not a valid interop conversion — Function.toJS requires the signature to use JS types (e.g. JSBoolean Function(JSString)).

### Reproduction steps

1. Add datadog_flutter_plugin (any version 3.3.0–3.5.0) with RUM enabled and at least one firstPartyHost configured, to a Flutter web app.
2. Run flutter build web or flutter run -d chrome (default dry-run behavior, no special flags needed).                                                     
3. Observe the compiler crash below instead of a normal build.

### SDK logs

Unexpected wasm dry run failure (255):                                                                                                                     
                                                                                                                                                             
  Unhandled exception:                                                                                                                                       
[...]

PR #1154, opened 2026-09-10
fix(web): avoid createJSInteropWrapper crash in wasm dry-run

p-divita-salto:
## Summary

Fixes the wasm dry-run crash reported in #1127. The 3.5.1 fix (referenced in that issue) targeted the `allowedTracingUrls` `match:` closure in `ddrum_web.dart`, but bisection shows that was never the trigger — the crash persists even with that closure fully removed.

The actual trigger is:

```dart
createJSInteropWrapper<RumWebPluginImpl>(_webPlugin!)
```

`createJSInteropWrapper` (backed by `@JSExport` codegen in `dart:js_interop`) generates a `Function.toJS` conversion with no real source offset. `dart2wasm`'s dry-run diagnostic reporter (`_CollectingDiagnosticReporter.report` → `Component.getLocation`) isn't guarded against that sentinel offset (`-1`) and throws `RangeError` instead of reporting normally — aborting `flutter build web` even without `--wasm`, since Flutter runs this dry-run by default.

This likely also explains why an actual `--wasm` compile is unaffected (as noted in the issue thread): only the dry-run-only `DryRunSummarizer` path seems to eagerly resolve a location for every node.

## Change

Replace the `@JSExport`/`createJSInteropWrapper` wrapper for `RumWebPluginImpl` with a hand-written extension-type wrapper (`createRumWebPluginJs`) that tears off the real, user-authored methods directly (`plugin.name`, `plugin.getConfigurationTelemetry.toJS`, `plugin.onRumStart.toJS`). Same runtime behavior, but every node now has a proper source location, so the dry-run reporter no longer trips.

## Test plan

- [x] `flutter build web` (non-wasm target) in a downstream app with RUM + first-party hosts configured, previously crashing on 3.5.1 — now completes successfully.
- [x] The dry-run subsequently runs to completion and correctly reports an unrelated, genuine incompatibility elsewhere in the downstream app's dependency tree (`universal_html`'s `dart:html` usage), confirming the reporter itself works again rather than just having the crash suppressed.
- [x] `dart analyze` on the two changed files.

refs: #1127

### DIFF
--- a/packages/datadog_flutter_plugin/lib/src/rum/web/rum_web_plugin.dart
+++ b/packages/datadog_flutter_plugin/lib/src/rum/web/rum_web_plugin.dart
@@ -12,7 +12,6 @@ import 'raw_events.dart';
 /// https://github.com/DataDog/browser-sdk/blob/bd6074ff1f33cb0b94acf7b6b7eae95180271475/packages/rum-core/src/domain/plugins.ts#L30
 ///
 /// Keep this class updated to match the original interface.
-@JSExport()
 abstract interface class RumWebPlugin {
   String get name;
 
@@ -28,7 +27,6 @@ abstract interface class RumWebPlugin {
   void onRumStart(OnRumStartOptions options) {}
 }
 
-@JSExport()
 class RumWebPluginImpl extends RumWebPlugin {
   @override
   String get name => 'DatadogFlutterWeb';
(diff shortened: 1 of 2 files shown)


fuzzybinary:
@p-divita-salto Retarget this on `develop` -- we aren't currently using `main`

p-divita-salto:
@fuzzybinary done!

fuzzybinary (approved):

domalessi (dismissed):

packages/datadog_flutter_plugin/lib/src/rum/web/rum_web_plugin.dart
```diff
@@ -60,3 +58,23 @@ class RumWebPluginImpl extends RumWebPlugin {
 extension type OnRumStartOptions._(JSObject _) implements JSObject {
   external JSFunction addEvent;
 }
+
+// Hand-written in place of `createJSInteropWrapper<RumWebPluginImpl>`:
+// the SDK's @JSExport codegen emits a Function.toJS conversion with no
+// source location, which crashes dart2wasm's dry-run diagnostic reporter
+// (RangeError in Source.getLocation, offset -1). Tearing off real,
+// user-authored methods keeps every node in this file properly located.
```
fuzzybinary: This is a bit overkill of an explication. This is what we do for every other object sso it's snot ssuper weird we're doing it here too. This comment can be removed.

fuzzybinary (changes requested):
Sorry @p-divita-salto  - I need you to [Sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)

Also since I need you to go back anyway, can you remove the one comment?

p-divita-salto:
@fuzzybinary done!

fuzzybinary (approved):

PR #1155, opened 2026-09-11
Add core Flags initialization timeout

leoromanovsky:
## Motivation

Flutter Flags initialization can wait indefinitely for its first evaluation context. The Flutter SDK needs bounded initialization behavior at the core layer.

## Changes and Decisions

- Add a five-second default initialization timeout to `datadog_flags`.
- Start one timer before the complete first initialization operation.
- Include cache loading, network work, JSON decoding, state publication, and persistence in the same budget.
- Throw `FlagsInitializationTimeoutException` when the timer wins.
- Keep the assignment operation active after timeout so a late result can publish assignments.
- Keep matching cached assignments available and ensure a newer context always wins.
- Publish assignments before persistence so a stalled store does not block evaluations.
- Forward the core configuration and exception through `datadog_flags_flutter` without another timer.

```mermaid
flowchart TD
    App["Flutter app"] --> Wrapper["datadog_flags_flutter"]
    Wrapper --> Core["datadog_flags core"]
    OpenFeature["Future OpenFeature app"] -.-> Provider["Datadog OpenFeature provider"]
    Provider -.-> Core

    Core --> Init["First initialize()"]
    Init --> Timer["Start one timeout timer"]
    Init --> Cache["Load matching cache"]
    Cache --> Fetch["Encode request and fetch response"]
    Fetch --> Decode["Decode JSON"]
    Decode --> Publish["Publish assignments"]
    Publish --> Persist["Persist assignments"]
    Persist --> Operation["Initialization operation completes"]
    Timer --> Race{"Which completes first?"}
    Operation --> Race
    Race -->|Timer| Timeout["Throw typed timeout; operation continues"]
    Race -->|Operation| Ready["Return with assignments"]
    Timeout -.-> Publish
```

### Customer Experience

- Customers can catch a typed timeout after the configured startup budget.
- Matching cached assignments remain usable after the exception.
- Without assignments, evaluations return customer defaults with `providerNotReady`.
- Late assignments affect later evaluations. They do not rerender the Flutter application automatically.
- The future OpenFeature provider can map the exception to `ProviderError` without starting another timer.

### Next Steps

- Add observable lifecycle state for late recovery.
- Map late recovery to OpenFeature `ProviderReady`.

### DIFF
--- a/packages/datadog_flags/CHANGELOG.md
+++ b/packages/datadog_flags/CHANGELOG.md
@@ -1,9 +1,5 @@
 # Changelog
 
-## 1.1.0
-
-* Add a configurable timeout for the complete first context initialization.
-
 ## 1.0.1
 
 * Fix Flutter Web flag evaluation events failing with HTTP 403 errors caused by CORS preflight requests.
--- a/packages/datadog_flags/CHANGELOG.md
+++ b/packages/datadog_flags/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.1.0
+
+* Add a configurable timeout for the complete first context initialization.
+
 ## 1.0.1
 
 * Fix Flutter Web flag evaluation events failing with HTTP 403 errors caused by CORS preflight requests.
--- a/packages/datadog_flags/lib/src/flags_client.dart
+++ b/packages/datadog_flags/lib/src/flags_client.dart
@@ -20,7 +20,8 @@ abstract interface class DatadogFlagsClient {
   /// The first call completes when initialization finishes or the configured
   /// initialization timeout expires. A timeout does not cancel the assignment
   /// operation. A late successful response still makes assignments available.
-  /// Later calls do not use the initialization timeout.
+  /// If a later call supersedes the first call, the first call remains bounded
+  /// by its original deadline. The later call does not use this timeout.
   ///
   /// Evaluations made before initialization completes return their provided
   /// default value with a `providerNotReady` error.
--- a/packages/datadog_flags/lib/src/flags_client.dart
+++ b/packages/datadog_flags/lib/src/flags_client.dart
@@ -17,6 +17,11 @@ abstract interface class DatadogFlagsClient {
 
   /// Fetches assignments for [context] and makes them available to evaluations.
   ///
+  /// The first call completes when initialization finishes or the configured
+  /// initialization timeout expires. A timeout does not cancel the assignment
+  /// operation. A late successful response still makes assignments available.
+  /// Later calls do not use the initialization timeout.
+  ///
   /// Evaluations made before initialization completes return their provided
   /// default value with a `providerNotReady` error.
   Future<void> initialize(
--- a/packages/datadog_flags/lib/src/flags_configuration.dart
+++ b/packages/datadog_flags/lib/src/flags_configuration.dart
@@ -31,7 +31,7 @@ final class DatadogFlagsConfiguration {
   /// Additional headers sent with precompute assignment requests.
   final Map<String, String>? customFlagsHeaders;
 
-  /// Maximum time to wait for the first evaluation context to become ready.
+  /// Maximum time to wait for initialization of the first evaluation context.
   ///
   /// This timeout covers the complete initialization operation. It includes
   /// loading stored assignments, encoding the request, fetching assignments,
@@ -42,7 +42,8 @@ final class DatadogFlagsConfiguration {
   ///
   /// The timeout applies only to the first
   /// [DatadogFlagsClient.initialize] call for each client. A `null`, zero, or
-  /// negative value disables the timeout.
+  /// negative value disables the timeout. If a later call supersedes the first
+  /// call, only the first call remains bounded by this timeout.
   final Duration? initializationTimeout;
 
   /// Overrides the exposure intake endpoint.
--- a/packages/datadog_flags/lib/src/flags_configuration.dart
+++ b/packages/datadog_flags/lib/src/flags_configuration.dart
@@ -31,14 +31,15 @@ final class DatadogFlagsConfiguration {
   /// Additional headers sent with precompute assignment requests.
   final Map<String, String>? customFlagsHeaders;
 
-  /// Maximum time to wait for initialization of the first evaluation context.
+  /// Maximum wall-clock time to wait for the first evaluation context.
   ///
-  /// This timeout covers the complete initialization operation. It includes
-  /// loading stored assignments, encoding the request, fetching assignments,
-  /// reading the response body, decoding JSON, storing assignments, and making
-  /// the assignments available for evaluation. It does not change the HTTP
-  /// client's timeout. The assignment operation continues after this timeout
-  /// and can make assignments available when it completes.
+  /// This value is one budget for the complete initialization operation. The
+  /// SDK does not restart it for each stage. It includes loading stored
+  /// assignments, encoding the request, fetching assignments, reading the
+  /// response body, decoding JSON, storing assignments, and making the
+  /// assignments available for evaluation. It does not change the HTTP client's
+  /// timeout. The assignment operation continues after this timeout and can
+  /// make assignments available when it completes.
   ///
   /// The timeout applies only to the first
   /// [DatadogFlagsClient.initialize] call for each client. A `null`, zero, or
--- a/packages/datadog_flags/lib/src/flags_configuration.dart
+++ b/packages/datadog_flags/lib/src/flags_configuration.dart
@@ -7,11 +7,15 @@ import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
 
 import 'datadog_flags_config.dart';
+import 'flags_client.dart';
 import 'flags_store.dart';
 
 /// Runtime configuration for Datadog feature flag clients.
 @immutable
 final class DatadogFlagsConfiguration {
+  /// Default maximum time to wait for the first evaluation context.
+  static const defaultInitializationTimeout = Duration(seconds: 5);
+
   /// Default interval for aggregating and sending flag evaluation telemetry.
   static const defaultEvaluationFlushInterval = Duration(seconds: 10);
 
@@ -26,6 +30,20 @@ final class DatadogFlagsConfiguration {
 
   /// Additional headers sent with precompute assignment requests.
   final Map<String, String>? customFlagsHeaders;
+
+  /// Maximum time to wait for the first evaluation context to become ready.
+  ///
+  /// This timeout covers the complete initialization operation. It includes
+  /// loading stored assignments, encoding the request, fetching assignments,
+  /// reading the response body, decoding JSON, storing assignments, and making
+  /// the assignments available for evaluation. It does not change the HTTP
+  /// client's timeout. The assignment operation continues after this timeout
+  /// and can make assignments available when it completes.
+  ///
+  /// The timeout applies only to the first
+  /// [DatadogFlagsClient.initialize] call for each client. A `null`, zero, or
+  /// negative value disables the timeout.
+  final Duration? initializationTimeout;
 
   /// Overrides the exposure intake endpoint.
   final Uri? customExposureEndpoint;
@@ -59,6 +77,7 @@ final class DatadogFlagsConfiguration {
   const DatadogFlagsConfiguration({
     this.customFlagsEndpoint,
     this.customFlagsHeaders,
+    this.initializationTimeout = defaultInitializationTimeout,
     this.customExposureEndpoint,
     this.trackExposures = true,
     this.customEvaluationEndpoint,
--- a/packages/datadog_flags/lib/src/flags_repository.dart
+++ b/packages/datadog_flags/lib/src/flags_repository.dart
@@ -203,6 +203,9 @@ class FlagsRepository {
 }
 
 class _InitializationDeadline {
+  // Dart timers cannot run while synchronous encoding or decoding blocks the
+  // isolate. The stopwatch checks enforce the wall-clock deadline before the
+  // operation can publish assignments or win the completion race.
   final Duration timeout;
   final Completer<void> _completion = Completer<void>();
   final Stopwatch _stopwatch = Stopwatch()..start();
--- a/packages/datadog_flags/lib/src/flags_repository.dart
+++ b/packages/datadog_flags/lib/src/flags_repository.dart
@@ -23,29 +23,60 @@ class FlagsRepository {
   final FlagAssignmentsFetcher fetcher;
   final DatadogFlagsStore? store;
   final DateTime Function() dateProvider;
+  final Duration? initializationTimeout;
+
+  @visibleForTesting
+  final Timer Function(Duration, void Function()) scheduleInitializationTimeout;
 
   FlagsData? _state;
   _CancelToken? _currentToken;
   Future<void> _cacheOperation = Future<void>.value();
+  bool _didStartInitialization = false;
 
   FlagsRepository({
     required this.clientName,
     required this.fetcher,
     this.store,
     required this.dateProvider,
+    this.initializationTimeout,
     this.storeReadTimeout = defaultStoreReadTimeout,
+    this.scheduleInitializationTimeout = _scheduleInitializationTimeout,
   });
 
   FlagsEvaluationContext? get context => _state?.context;
 
   FlagAssignment? flagAssignment(String key) => _state?.flags[key];
 
-  Future<void> initialize(FlagsEvaluationContext context) async {
+  Future<void> initialize(FlagsEvaluationContext context) {
     _currentToken?.cancel();
     final token = _CancelToken();
     _currentToken = token;
 
+    final timeout = _takeInitializationTimeout();
+    final deadline = timeout == null
+        ? null
+        : _InitializationDeadline(timeout, scheduleInitializationTimeout);
+    final operation = _initialize(context, token, deadline);
+    if (deadline == null) {
+      return operation;
+    }
+    deadline.observe(operation);
+    return deadline.future;
+  }
+
+  Future<void> _initialize(
+    FlagsEvaluationContext context,
+    _CancelToken token,
+    _InitializationDeadline? deadline,
+  ) async {
     final cached = store == null ? null : await _readCached();
+    if (token.isCanceled) {
+      return;
+    }
+    final cacheDeadlineYield = _yieldAfterExpiredDeadline(deadline);
+    if (cacheDeadlineYield != null) {
+      await cacheDeadlineYield;
+    }
     if (token.isCanceled) {
       return;
     }
@@ -68,13 +99,48 @@ class FlagsRepository {
         context: context,
         date: dateProvider(),
       );
-      _state = data;
       await _writeCached(data);
+      if (token.isCanceled) {
+        return;
+      }
+      final publicationDeadlineYield = _yieldAfterExpiredDeadline(deadline);
+      if (publicationDeadlineYield != null) {
+        await publicationDeadlineYield;
+      }
+      if (token.isCanceled) {
+        return;
+      }
+      _state = data;
     } catch (_) {
       if (!token.isCanceled && matchingCached == null) {
         _state = null;
       }
     }
+  }
+
+  Duration? _takeInitializationTimeout() {
+    if (_didStartInitialization) {
+      return null;
+    }
+    _didStartInitialization = true;
+    final timeout = initializationTimeout;
+    return timeout != null && timeout > Duration.zero ? timeout : null;
+  }
+
+  static Timer _scheduleInitializationTimeout(
+    Duration timeout,
+    void Function() action,
+  ) {
+    return Timer(timeout, action);
+  }
+
+  static Future<void>? _yieldAfterExpiredDeadline(
+    _InitializationDeadline? deadline,
+  ) {
+    if (deadline?.expireIfNeeded() ?? false) {
+      return Future<void>.delayed(Duration.zero);
+    }
+    return null;
   }
 
   bool _hasCurrentStateForContext(FlagsEvaluationContext context) {
@@ -133,6 +199,69 @@ class FlagsRepository {
     final next = _cacheOperation.then((_) => operation());
     _cacheOperation = next.catchError((_) {});
     return next;
+  }
+}
+
+class _InitializationDeadline {
+  final Duration timeout;
+  final Completer<void> _completion = Completer<void>();
+  final Stopwatch _stopwatch = Stopwatch()..start();
+  Timer? _timer;
+  var _expired = false;
+
+  _InitializationDeadline(
+    this.timeout,
+    Timer Function(Duration, void Function()) schedule,
+  ) {
+    _timer = schedule(timeout, _expire);
+    if (_expired) {
+      _timer?.cancel();
+    }
+  }
+
+  Future<void> get future => _completion.future;
+
+  bool expireIfNeeded() {
+    if (!_expired && _stopwatch.elapsed >= timeout) {
+      _expire();
+    }
+    return _expired;
+  }
+
+  void observe(Future<void> operation) {
+    expireIfNeeded();
+    operation.then<void>(
+      (_) => _completeOperation(),
+      onError: _completeOperationWithError,
+    );
+  }
+
+  void _expire() {
+    if (_completion.isCompleted) {
+      return;
+    }
+    _expired = true;
+    _stopwatch.stop();
+    _timer?.cancel();
+    _completion.complete();
+  }
+
+  void _completeOperation() {
+    if (expireIfNeeded()) {
+      return;
+    }
+    _stopwatch.stop();
+    _timer?.cancel();
+    _completion.complete();
+  }
+
+  void _completeOperationWithError(Object error, StackTrace stackTrace) {
+    if (expireIfNeeded()) {
+      return;
+    }
+    _stopwatch.stop();
+    _timer?.cancel();
+    _completion.completeError(error, stackTrace);
   }
 }
 
--- a/packages/datadog_flags/lib/src/flags_repository.dart
+++ b/packages/datadog_flags/lib/src/flags_repository.dart
@@ -94,6 +94,13 @@ class FlagsRepository {
       if (token.isCanceled) {
         return;
       }
+      final fetchDeadlineYield = _yieldAfterExpiredDeadline(deadline);
+      if (fetchDeadlineYield != null) {
+        await fetchDeadlineYield;
+      }
+      if (token.isCanceled) {
+        return;
+      }
       final data = FlagsData(
         flags: assignments.flags,
         context: context,
--- a/packages/datadog_flags/test/initialization_timeout_test.dart
+++ b/packages/datadog_flags/test/initialization_timeout_test.dart
@@ -381,29 +381,27 @@ void main() {
     },
   );
 
-  test('uses one total budget across storage, network, and JSON decoding',
-      () async {
-    void Function()? timeoutAction;
-    var scheduleCount = 0;
-    final store = _DelayedReadStore();
-    final httpClient = _ControlledResponseBodyClient();
+  test('returns after slow JSON decoding before assignment storage', () async {
+    final store = _DelayedWriteStore();
     addTearDown(() async {
-      if (!store.allowRead.isCompleted) {
-        store.allowRead.complete();
+      if (!store.allowWrite.isCompleted) {
+        store.allowWrite.complete();
       }
-      httpClient.close();
+    });
+    final httpClient = MockClient((_) async {
+      return http.Response('{}', 200);
     });
     final configuration = _configuration(
       httpClient: httpClient,
       store: store,
-      initializationTimeout: const Duration(seconds: 5),
+      initializationTimeout: const Duration(milliseconds: 1),
     );
     final fetcher = FlagAssignmentsFetcher(
       datadogConfig: _datadogConfig,
       configuration: configuration,
       httpClient: httpClient,
       responseDecoder: (_) {
-        timeoutAction!();
+        _blockFor(const Duration(milliseconds: 10));
         return PrecomputedAssignments(
           flags: {'show-paywall': _assignment(value: true)},
         );
@@ -415,24 +413,17 @@ void main() {
       store: store,
       dateProvider: DateTime.now,
       initializationTimeout: configuration.initializationTimeout,
-      scheduleInitializationTimeout: (_, action) {
-        scheduleCount += 1;
-        timeoutAction = action;
-        return _TestTimer();
-      },
+      scheduleInitializationTimeout: (_, __) => _TestTimer(),
     );
 
     final initialization = repository.initialize(_context);
-    await store.readStarted.future;
-    expect(scheduleCount, 1);
-
-    store.allowRead.complete();
-    await httpClient.requestStarted.future;
-    final bodyCompletion = httpClient.completeBody(_assignmentsResponse());
     await initialization.timeout(const Duration(seconds: 1));
 
+    expect(store.writeStarted.isCompleted, isFalse);
     expect(repository.flagAssignment('show-paywall'), isNull);
-    await bodyCompletion;
+
+    await store.writeStarted.future.timeout(const Duration(seconds: 1));
+    store.allowWrite.complete();
     await _waitUntil(() => repository.flagAssignment('show-paywall') != null);
   });
 
@@ -542,6 +533,11 @@ Future<void> _waitUntil(
   }
 }
 
+void _blockFor(Duration duration) {
+  final stopwatch = Stopwatch()..start();
+  while (stopwatch.elapsed < duration) {}
+}
+
 class _ControlledResponseBodyClient extends http.BaseClient {
   final StreamController<List<int>> _body = StreamController<List<int>>();
   final Completer<void> requestStarted = Completer<void>();
@@ -634,8 +630,7 @@ class _SlowIterable extends Iterable<Object?> {
 
   @override
   Iterator<Object?> get iterator {
-    final stopwatch = Stopwatch()..start();
-    while (stopwatch.elapsed < delay) {}
+    _blockFor(delay);
     return <Object?>[true].iterator;
   }
 }
--- a/packages/datadog_flags/test/initialization_timeout_test.dart
+++ b/packages/datadog_flags/test/initialization_timeout_test.dart
@@ -1,0 +1,472 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0. This product includes software
+// developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:datadog_flags/datadog_flags.dart';
+import 'package:datadog_flags/src/assignment.dart';
+import 'package:datadog_flags/src/flag_assignments_fetcher.dart';
+import 'package:datadog_flags/src/flags_repository.dart';
+import 'package:datadog_flags/src/flags_store.dart';
+import 'package:http/http.dart' as http;
+import 'package:http/testing.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('uses a five-second initialization timeout by default', () {
+    expect(
+      const DatadogFlagsConfiguration().initializationTimeout,
+      const Duration(seconds: 5),
+    );
+    expect(
+      const DatadogFlagsConfiguration(
+        initializationTimeout: Duration(milliseconds: 2500),
+      ).initializationTimeout,
+      const Duration(milliseconds: 2500),
+    );
+  });
+
+  test('accepts disabled initialization timeout values', () {
+    expect(
+      const DatadogFlagsConfiguration(
+        initializationTimeout: null,
+      ).initializationTimeout,
+      isNull,
+    );
+    expect(
+      const DatadogFlagsConfiguration(
+        initializationTimeout: Duration.zero,
+      ).initializationTimeout,
+      Duration.zero,
+    );
+    expect(
+      const DatadogFlagsConfiguration(
+        initializationTimeout: Duration(milliseconds: -1),
+      ).initializationTimeout,
+      const Duration(milliseconds: -1),
+    );
+  });
+
+  test('does not schedule a timer for disabled timeout values', () async {
+    for (final timeout in <Duration?>[
+      null,
+      Duration.zero,
+      const Duration(milliseconds: -1),
+    ]) {
+      var scheduleCount = 0;
+      final httpClient = MockClient((_) async {
+        return http.Response(jsonEncode(_assignmentsResponse()), 200);
+      });
+      final configuration = _configuration(
+        httpClient: httpClient,
+        initializationTimeout: timeout,
+      );
+      final repository = FlagsRepository(
+        clientName: DatadogFlags.defaultClientName,
+        fetcher: FlagAssignmentsFetcher(
+          datadogConfig: _datadogConfig,
+          configuration: configuration,
+          httpClient: httpClient,
+        ),
+        dateProvider: DateTime.now,
+        initializationTimeout: timeout,
+        scheduleInitializationTimeout: (_, __) {
+          scheduleCount += 1;
+          return _TestTimer();
+        },
+      );
+
+      await repository.initialize(_context);
+
+      expect(scheduleCount, 0, reason: 'Unexpected timer for $timeout');
+      expect(repository.flagAssignment('show-paywall'), isNotNull);
+    }
+  });
+
+  test(
+    'times out while downloading the response body and recovers later',
+    () async {
+      final httpClient = _ControlledResponseBodyClient();
+      final datadogFlags = DatadogFlags();
+      addTearDown(() async {
+        httpClient.close();
+        await datadogFlags.disable();
+      });
+      await datadogFlags.enable(
+        configuration: _configuration(
+          httpClient: httpClient,
+          initializationTimeout: const Duration(milliseconds: 5),
+        ),
+      );
+      final client = datadogFlags.sharedClient();
+
+      final initialization = client.initialize(_context);
+      await httpClient.requestStarted.future;
+      await initialization.timeout(const Duration(seconds: 1));
+
+      expect(httpClient.bodyCompleted, isFalse);
+      expect(
+        client
+            .getBooleanDetails(key: 'show-paywall', defaultValue: false)
+            .error,
+        FlagEvaluationError.providerNotReady,
+      );
+
+      await httpClient.completeBody(_assignmentsResponse());
+      await _waitUntil(() {
+        return client
+                .getBooleanDetails(key: 'show-paywall', defaultValue: false)
+                .error ==
+            null;
+      });
+      expect(
+        client
+            .getBooleanDetails(key: 'show-paywall', defaultValue: false)
+            .value,
+        isTrue,
+      );
+    },
+  );
+
+  test(
+    'uses matching stored assignments when initialization times out',
+    () async {
+      final store = InMemoryDatadogFlagsStore();
+      await store.write(
+        DatadogFlags.defaultClientName,
+        _storedAssignments(_context, value: true),
+      );
+      final httpClient = _ControlledResponseBodyClient();
+      final datadogFlags = DatadogFlags();
+      addTearDown(() async {
+        httpClient.close();
+        await datadogFlags.disable();
+      });
+      await datadogFlags.enable(
+        configuration: _configuration(
+          httpClient: httpClient,
+          store: store,
+          initializationTimeout: const Duration(milliseconds: 5),
+        ),
+      );
+      final client = datadogFlags.sharedClient();
+
+      await client.initialize(_context).timeout(const Duration(seconds: 1));
+
+      expect(
+        client
+            .getBooleanDetails(key: 'show-paywall', defaultValue: false)
+            .value,
+        isTrue,
+      );
+
+      await httpClient.completeBody(_assignmentsResponse(booleanValue: false));
+      await _waitUntil(() {
+        return client
+                .getBooleanDetails(key: 'show-paywall', defaultValue: true)
+                .value ==
+            false;
+      });
+    },
+  );
+
+  test('includes assignment storage before publishing ready state', () async {
+    final store = _DelayedWriteStore();
+    final datadogFlags = DatadogFlags();
+    addTearDown(() async {
+      if (!store.allowWrite.isCompleted) {
+        store.allowWrite.complete();
+      }
+      await datadogFlags.disable();
+    });
+    await datadogFlags.enable(
+      configuration: _configuration(
+        httpClient: MockClient((_) async {
+          return http.Response(jsonEncode(_assignmentsResponse()), 200);
+        }),
+        store: store,
+        initializationTimeout: const Duration(milliseconds: 5),
+      ),
+    );
+    final client = datadogFlags.sharedClient();
+
+    final initialization = client.initialize(_context);
+    await store.writeStarted.future;
+    await initialization.timeout(const Duration(seconds: 1));
+
+    expect(
+      client.getBooleanDetails(key: 'show-paywall', defaultValue: false).error,
+      FlagEvaluationError.providerNotReady,
+    );
+
+    store.allowWrite.complete();
+    await _waitUntil(() {
+      return client
+              .getBooleanDetails(key: 'show-paywall', defaultValue: false)
+              .error ==
+          null;
+    });
+  });
+
+  test(
+    'applies the initialization timeout only to the first context',
+    () async {
+      final responses = <Completer<http.Response>>[];
+      final datadogFlags = DatadogFlags();
+      addTearDown(datadogFlags.disable);
+      await datadogFlags.enable(
+        configuration: _configuration(
+          httpClient: MockClient((_) {
+            final response = Completer<http.Response>();
+            responses.add(response);
+            return response.future;
+          }),
+          initializationTimeout: const Duration(milliseconds: 5),
+        ),
+      );
+      final client = datadogFlags.sharedClient();
+
+      await client.initialize(_context).timeout(const Duration(seconds: 1));
+
+      var secondCompleted = false;
+      final second = client
+          .initialize(const FlagsEvaluationContext(targetingKey: 'user-second'))
+          .whenComplete(() => secondCompleted = true);
+      await _waitUntil(() => responses.length == 2);
+      await Future<void>.delayed(const Duration(milliseconds: 20));
+      expect(secondCompleted, isFalse);
+
+      responses[1].complete(
+        http.Response(
+          jsonEncode(_assignmentsResponse(booleanValue: false)),
+          200,
+        ),
+      );
+      await second.timeout(const Duration(seconds: 1));
+      responses[0].complete(
+        http.Response(jsonEncode(_assignmentsResponse()), 200),
+      );
+      await Future<void>.delayed(Duration.zero);
+
+      expect(
+        client.getBooleanDetails(key: 'show-paywall', defaultValue: true).value,
+        isFalse,
+      );
+    },
+  );
+
+  test('keeps the deadline active through response JSON decoding', () async {
+    void Function()? timeoutAction;
+    final fetcher = FlagAssignmentsFetcher(
+      datadogConfig: _datadogConfig,
+      configuration: _configuration(
+        httpClient: MockClient((_) async {
+          return http.Response('{}', 200);
+        }),
+      ),
+      httpClient: MockClient((_) async {
+        return http.Response('{}', 200);
+      }),
+      responseDecoder: (_) {
+        timeoutAction!();
+        return PrecomputedAssignments(
+          flags: {'show-paywall': _assignment(value: true)},
+        );
+      },
+    );
+    final repository = FlagsRepository(
+      clientName: DatadogFlags.defaultClientName,
+      fetcher: fetcher,
+      dateProvider: DateTime.now,
+      initializationTimeout: const Duration(seconds: 5),
+      scheduleInitializationTimeout: (_, action) {
+        timeoutAction = action;
+        return _TestTimer();
+      },
+    );
+
+    await repository.initialize(_context);
+
+    expect(repository.flagAssignment('show-paywall'), isNull);
+    await _waitUntil(() => repository.flagAssignment('show-paywall') != null);
+  });
+
+  test('counts synchronous request encoding against the deadline', () async {
+    final response = Completer<http.Response>();
+    final httpClient = MockClient((_) => response.future);
+    final configuration = _configuration(
+      httpClient: httpClient,
+      initializationTimeout: const Duration(milliseconds: 1),
+    );
+    final repository = FlagsRepository(
+      clientName: DatadogFlags.defaultClientName,
+      fetcher: FlagAssignmentsFetcher(
+        datadogConfig: _datadogConfig,
+        configuration: configuration,
+        httpClient: httpClient,
+      ),
+      dateProvider: DateTime.now,
+      initializationTimeout: configuration.initializationTimeout,
+      scheduleInitializationTimeout: (_, __) => _TestTimer(),
+    );
+    final context = FlagsEvaluationContext(
+      targetingKey: 'user-123',
+      attributes: {'slow': _SlowIterable(const Duration(milliseconds: 10))},
+    );
+
+    await repository.initialize(context).timeout(const Duration(seconds: 1));
+
+    expect(repository.flagAssignment('show-paywall'), isNull);
+    response.complete(http.Response(jsonEncode(_assignmentsResponse()), 200));
+    await _waitUntil(() => repository.flagAssignment('show-paywall') != null);
+  });
+}
+
+const _context = FlagsEvaluationContext(targetingKey: 'user-123');
+
+const _datadogConfig = DatadogFlagsConfig(
+  clientToken: 'client-token',
+  env: 'staging',
+  site: DatadogFlagsSite.us1,
+);
+
+DatadogFlagsConfiguration _configuration({
+  required http.Client httpClient,
+  Duration? initializationTimeout =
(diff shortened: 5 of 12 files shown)


drichards-87 (dismissed):

packages/datadog_flags/lib/src/flags_client.dart
```diff
@@ -17,6 +17,11 @@ abstract interface class DatadogFlagsClient {
 
   /// Fetches assignments for [context] and makes them available to evaluations.
   ///
+  /// The first call completes when initialization finishes or the configured
+  /// initialization timeout expires. A timeout does not cancel the assignment
+  /// operation. A late successful response still makes assignments available.
+  /// Later calls do not use the initialization timeout.
```
typotter: what if a later call happens before the first call's timeout expires?

packages/datadog_flags/lib/src/flags_configuration.dart
```diff
@@ -27,6 +31,20 @@ final class DatadogFlagsConfiguration {
   /// Additional headers sent with precompute assignment requests.
   final Map<String, String>? customFlagsHeaders;
 
+  /// Maximum time to wait for the first evaluation context to become ready.
```
typotter:
```suggestion
  /// Maximum time to wait for the first evaluation context to produce flag assignments.
```

typotter (dismissed):

drichards-87 (dismissed):

typotter (dismissed):

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -23,32 +23,63 @@ class FlagsRepository {
...
+    final timeout = _takeInitializationTimeout();
+    final deadline = timeout == null
+        ? null
+        : _InitializationDeadline(timeout, scheduleInitializationTimeout);
+    final operation = _initialize(context, token, deadline);
+    if (deadline == null) {
+      return operation;
+    }
+    deadline.observe(operation);
+    return deadline.future;
```
fuzzybinary: This seems weirdly overly complicated to me. If we want to use this `_InitializationDeadline` class, since `_initialize` already has the deadline and is the operation itself, why can't it perform `.observe` and hand back the proper future? This could reduce to:
```dart
return _initialize(context, token, timeout);
```

I guess I also don't understand why there are two functions in the first place?

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -68,15 +99,50 @@ class FlagsRepository {
...
+      if (token.isCanceled) {
+        return;
+      }
+      final publicationDeadlineYield = _yieldAfterExpiredDeadline(deadline);
+      if (publicationDeadlineYield != null) {
+        await publicationDeadlineYield;
+      }
+      if (token.isCanceled) {
+        return;
+      }
```
fuzzybinary: Why do we need to check cancelation both before an after this expiration?

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -68,15 +99,50 @@ class FlagsRepository {
...
       if (!token.isCanceled && matchingCached == null) {
         _state = null;
       }
     }
   }
 
+  Duration? _takeInitializationTimeout() {
+    if (_didStartInitialization) {
+      return null;
+    }
```
fuzzybinary: What is this preventing?

It looks like if I call initialize twice, the second one won't have a timeout... is that intended?

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -68,15 +99,50 @@ class FlagsRepository {
...
+    void Function() action,
+  ) {
+    return Timer(timeout, action);
+  }
+
+  static Future<void>? _yieldAfterExpiredDeadline(
+    _InitializationDeadline? deadline,
+  ) {
+    if (deadline?.expireIfNeeded() ?? false) {
+      return Future<void>.delayed(Duration.zero);
```
fuzzybinary: I'm curious why this is needed? Does this need to be pushed back onto the microtask queue for some reason or would this work:
```
return Future<void>.value();
```

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -136,6 +202,69 @@ class FlagsRepository {
   }
 }
 
+class _InitializationDeadline {
+  final Duration timeout;
+  final Completer<void> _completion = Completer<void>();
+  final Stopwatch _stopwatch = Stopwatch()..start();
+  Timer? _timer;
+  var _expired = false;
```
fuzzybinary: From what I can tell, this uses both a `Timer` and a `Stopwatch` to be able to timeout faster than the microtask queue can process the `Timer` is that accurate? E.g. you want to be able to timeout as close to 5 seconds as possible, but if you're in the middle of processing something then the Timer won't fire until the microtask queue gets a chance to run, which might be delayed.

If that's the case, can we comment that?

packages/datadog_flags/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.1.0
+
+* Add a configurable timeout for the complete first context initialization.
+
```
fuzzybinary: Let releaser add changelog items (I promise we'll get a git hook in for this soon 😅)

fuzzybinary (changes requested):
Okay, this struct me as WAY overly complex for what we're trying to do, and so I did feed it to Claude to back me up. *IF* it's understanding your intent correclty, it recommends using `Future.any` over your custom `_IntializationDeadline` system, and it explains it better than I can:

Future.any gets you the actual behavior you want, directly:

```dart
  Future<void> initialize(FlagsEvaluationContext context) {
    _currentToken?.cancel();
    final token = _CancelToken();
    _currentToken = token;
    
    final operation = _initialize(context, token);
    final timeout = _takeInitializationTimeout();
    if (timeout == null) {
      return operation;
    }
    return Future.any([operation, Future<void>.delayed(timeout)]);
  } 
```

  - _initialize goes back to having zero awareness of deadlines — delete the deadline parameter and every _yieldAfterExpiredDeadline call. Cache read, fetch, and publish logic is no longer interleaved with timeout bookkeeping.
  - Future.any doesn't cancel the loser — the losing future (whichever isn't first) keeps running to completion in the background, which is exactly "operation keeps going, deadline future just returns early."
  - It's safe against errors here specifically because _initialize already swallows everything internally (catch (_) { ... }) and never actually completes with an error — so there's no risk of Future.any surfacing an unexpected error from the
    loser.
  - _InitializationDeadline, its Stopwatch, Completer, and the polling expireIfNeeded() calls all disappear.

  Where this needs a decision, not just a mechanical swap: the scheduleInitializationTimeout hook and the "counts synchronous request encoding against the deadline" test. That test only passes today because production code polls a Stopwatch
  manually — its injected fake Timer never invokes the callback at all. Under Future.any with a real Future<void>.delayed(timeout), a real Timer that gets starved by synchronous work still fires as soon as the isolate yields, so production behavior
  is unaffected — but that specific test needs to change from "inject a dead fake timer" to "use a real short duration and observe it actually fires," since there's no longer a manual escape hatch for a timer that's rigged to never fire.

  If you still want deterministic, non-wall-clock-dependent tests for the timeout path, swap the hook's shape from "returns a Timer" to "returns a Future":

fuzzybinary:
This is of course *IF* Claude's read is correct - you want `initialize` to return immediately if you reach `timeout`, but continue with initialization until it's complete.

packages/datadog_flags/CHANGELOG.md
leoromanovsky: Understood; I'll add that note to the contribution or agents file in the other PR to avoid this.

leoromanovsky:
The product goal is for a customer should set the maximum time their app can wait for the first flag initialization.

This value must be one wall-clock budget. It covers persistent storage, request encoding, network I/O, JSON decoding, assignment storage, and state publication. Customers should not need to understand these stages.

When the budget expires, `initialize()` should return. Compatible cached assignments remain usable. The assignment operation continues, and a late successful result can make the client ready. A newer context must always win over an older operation.

The timeout can complete only when the isolate yields. At that point, the deadline must win before more work starts or assignments become visible. This constraint is why `Future.any` alone does not meet the product goal. A microtask can continue initialization before a delayed timer runs. The stopwatch checks enforce the wall-clock budget at these boundaries.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -63,20 +94,62 @@ class FlagsRepository {
...
+        return;
+      }
+      final publicationDeadlineYield = _yieldAfterExpiredDeadline(deadline);
+      if (publicationDeadlineYield != null) {
+        await publicationDeadlineYield;
+      }
+      if (token.isCanceled) {
+        return;
+      }
+      _state = data;
```
aarsilv, line 120: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P0]** Please publish `_state` before awaiting `_writeCached`.

`e9c1d79` moved this line below `await _writeCached(data)`. `_writeCached` has no timeout, unlike `_readCached` with its `storeReadTimeout`. So if an app-supplied store returns a write future that never settles, a successful fetch is decoded and then never published. `origin/develop` published first, so this is a regression. iOS and Android both publish in memory before calling persistence.

<details>
<summary><b>Executed evidence</b> — a never-settling store write, run against three trees</summary>

I added a `DatadogFlagsStore` whose `write` returns a future that never completes, then asserted the fetched assignment becomes visible. Run first in `dart:3.6.2`, then re-run on Dart 3.13.3 — identical results on both, same `Expected: not null / Actual: <null>` on head:

| Tree | Result |
| --- | --- |
| `origin/develop` | pass |
| PR head `5fdccc9` | **fail** — `Expected: not null / Actual: <null>` |
| PR head + the fix below | pass |

All 72 tests pass on PR head, on both SDKs. One test does touch the ordering — `includes assignment storage in the total initialization budget` at lines 222-259 — but it asserts the current behaviour, so nothing would have caught the loss of the old guarantee.

Scope, stated precisely: this covers a fresh repository with no publishable state. Matching cached assignments are published earlier, at lines 84-90, and survive. Where there is no matching cache, `_state` stays null for the life of that repository instance — a retry re-fetches and then queues a second write behind the stuck one on `_cacheOperation`, so it cannot publish either.

</details>

<details>
<summary><b>What I did and did not establish about how a write stalls</b></summary>

What the test establishes: given a write future that never settles, publication never happens at head and does happen on `develop`.

What I did **not** establish, and am not claiming: which real device conditions produce such a future, or how often. `packages/datadog_flags/lib` ships only `InMemoryDatadogFlagsStore`, so a production store is app-supplied and its failure modes are the app's. A future that never settles is a different case from one that throws, and the `catch` inside `_writeCached` only covers the latter — that asymmetry is the part I am confident about.

</details>

<details>
<summary><b>What iOS and Android do, and where the analogy stops</b></summary>

iOS `DatadogFlags/Sources/Client/FlagsRepository.swift:323` assigns `self.flagsData`, then `:329` calls `writeState()`, which reaches `featureScope.dataStore.setValue(...)` at `FlagsDataStore.swift:19` with no completion handler and no `await`.

Android `internal/repository/DefaultFlagsRepository.kt:50` calls `atomicState.set(newState)`, then `:53` calls `persistenceManager.saveFlagsState(..., DataStoreWriteCallback)`, whose `onFailure` only logs.

So on both platforms a *pending* persistence write cannot delay evaluation or readiness, because neither awaits it. Two honest caveats: both platforms invoke persistence before their `ready` transition, so a *synchronously blocking* persistence call could still delay readiness there; and their persistence APIs are callback- or void-shaped, not an app-supplied Dart `Future<void>`. The analogy is about ordering, not about equivalent contracts.

</details>

<details>
<summary><b>Suggested fix</b> — publish, then persist</summary>

Replacing lines 104-120:

```dart
      final data = FlagsData(
        flags: assignments.flags,
        context: context,
        date: dateProvider(),
      );
      final publicationDeadlineYield = _yieldAfterExpiredDeadline(deadline);
      if (publicationDeadlineYield != null) {
        await publicationDeadlineYield;
      }
      if (token.isCanceled) {
        return;
      }
      _state = data;
      await _writeCached(data);
```

The operation still awaits the write, so the write is still part of the initialization operation. Note this drops the post-write `token.isCanceled` check at lines 110-112; the meaningful guard moves ahead of publication, which is where it matters.

This does invalidate `includes assignment storage in the total initialization budget` at lines 222-259, which asserts `providerNotReady` while the write is pending. Replacing it is a policy call, not just a mechanical edit: the PR's docs describe storage and publication as parts of one operation, and that test encodes the stronger reading. Two replacements, which also keep repository-level rather than public-client coverage:

```dart
  test('assignment storage does not gate assignment publication', () async {
    final store = _DelayedWriteStore();
    addTearDown(() {
      if (!store.allowWrite.isCompleted) {
        store.allowWrite.complete();
      }
    });
    final httpClient = MockClient((_) async {
      return http.Response(jsonEncode(_assignmentsResponse()), 200);
    });
    final configuration = _configuration(
      httpClient: httpClient,
      store: store,
      initializationTimeout: const Duration(milliseconds: 5),
    );
    final repository = FlagsRepository(
      clientName: DatadogFlags.defaultClientName,
      fetcher: FlagAssignmentsFetcher(
        datadogConfig: _datadogConfig,
        configuration: configuration,
        httpClient: httpClient,
      ),
      store: store,
      dateProvider: DateTime.now,
      initializationTimeout: configuration.initializationTimeout,
    );

    await repository.initialize(_context).timeout(const Duration(seconds: 1));
    await store.writeStarted.future.timeout(const Duration(seconds: 1));

    expect(repository.flagAssignment('show-paywall')?.variationValue, isTrue);
  });

  test('assignment storage stays inside the initialization operation',
      () async {
    final store = _DelayedWriteStore();
    addTearDown(() {
      if (!store.allowWrite.isCompleted) {
        store.allowWrite.complete();
      }
    });
    final httpClient = MockClient((_) async {
      return http.Response(jsonEncode(_assignmentsResponse()), 200);
    });
    final configuration = _configuration(
      httpClient: httpClient,
      store: store,
      initializationTimeout: null,
    );
    final repository = FlagsRepository(
      clientName: DatadogFlags.defaultClientName,
      fetcher: FlagAssignmentsFetcher(
        datadogConfig: _datadogConfig,
        configuration: configuration,
        httpClient: httpClient,
      ),
      store: store,
      dateProvider: DateTime.now,
      initializationTimeout: configuration.initializationTimeout,
    );

    var completed = false;
    final initialization =
        repository.initialize(_context).whenComplete(() => completed = true);
    await store.writeStarted.future.timeout(const Duration(seconds: 1));
    await Future<void>.delayed(const Duration(milliseconds: 50));

    expect(completed, isFalse);

    store.allowWrite.complete();
    await initialization.timeout(const Duration(seconds: 1));
    expect(await store.read(DatadogFlags.defaultClientName), isNotNull);
  });
```

Measured, and the two are not equivalent: **the first is red on PR head and green with the fix. The second passes on both orderings** — with the timeout disabled, both await the write. It is there to pin that the write stays inside the operation, not to test the fix. With exactly the snippet above and both replacement tests, and no other change: `dart analyze` clean, `dart test` 73/73. That discrimination reproduces on `dart:3.6.2` and on Dart 3.13.3.

</details>

<details>
<summary><b>An alternative I tried and withdrew</b></summary>

My first patch kept the ordering and bounded each queued cache operation with `operation().timeout(storeOperationTimeout, onTimeout: () {})`. It analysed clean, kept the existing 72 green, and turned both never-settling-write tests green — publication and the `reset()` wedge. Then an attack test showed it lets a slow-but-working write land after `reset()` deleted the cache, resurrecting stale assignments (`Expected: null / Actual: <Instance of 'FlagsData'>`). The ordering fix above passes that test. It also fails the first replacement test above, because a bounded write still publishes only after the write. I got this wrong before testing it.

Also worth separating: `reset()` hangs behind a stalled write on `develop`, at head, and with this fix. `git diff origin/develop...HEAD` changes no lines in `_enqueueCacheOperation`, `_writeCached` or `_deleteCached`, so that one is pre-existing and not this PR's to fix.

</details>

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -23,32 +23,63 @@ class FlagsRepository {
...
+  Future<void> _initialize(
+    FlagsEvaluationContext context,
+    _CancelToken token,
+    _InitializationDeadline? deadline,
+  ) async {
     final cached = store == null ? null : await _readCached();
     if (token.isCanceled) {
       return;
     }
+    final cacheDeadlineYield = _yieldAfterExpiredDeadline(deadline);
```
aarsilv, line 76: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P1]** Please move this yield below the `_state = matchingCached` block.

The poll sits above the block that publishes matching cached assignments. Suppose the budget expires while the stored-assignment load blocks the isolate. The caller then resumes with `_state` still null and reads `providerNotReady` defaults, although the read had already returned usable assignments. `_state = matchingCached` runs on the next turn. The PR description promises "matching cached assignments remain usable" at timeout.

<details>
<summary><b>Executed evidence</b> — a store read that blocks the isolate past the budget</summary>

A store holding assignments for this exact context. Its `read` blocks the isolate for 300 ms against a 50 ms budget. The read returns, and the default 100 ms `storeReadTimeout` does not discard it. `.timeout()` is attached only after `store.read` returns, so the blocking happens before that timer exists (`flags_repository.dart:176-179`):

```
caller sees matching cached assignments when the budget expires
during a synchronous stored-assignment load
  Expected: true
    Actual: <null>
  cached assignments were loaded but not visible to the caller
```

Green after the move below. Red on head and green with the move on `dart:3.6.2` and on Dart 3.13.3, with the same message both times. A correction to what I posted: I said this needed `storeReadTimeout` raised. It does not. At the default the test is red on head and green with the move, exactly as above. The gap closes when the `Future<void>.delayed(Duration.zero)` continuation at `flags_repository.dart:148` runs. That is a later event-loop turn, normally the next one, and a loaded isolate can make it later. So the exposed caller is one that evaluates before then, which is the documented usage pattern. That is why I have this at `[P1]` rather than higher.

</details>

<details>
<summary><b>Which case this needs, and which it does not</b></summary>

With a fast store read the deadline has not expired, so the poll returns `null` and nothing changes. With a read still pending at the deadline there are no cached assignments to publish yet. The nearest existing coverage is `initialization_timeout_test.dart:134-176`. That test uses an empty delayed store, so it checks the null-at-timeout half, not a pending read that later returns matching cache.

The gap needs a read that *returns* but blocks the isolate while doing so — a `wasm`/JS-interop `localStorage` read, or a store doing synchronous work before its first `await`. `Future.timeout` cannot interrupt that, so `storeReadTimeout` does not cover it. `flags_store.dart:11-23` does not prohibit synchronous work in a store. I am not claiming these three cases are exhaustive; an async continuation can also be delayed by unrelated isolate starvation.

Both siblings treat "we have matching cache" as observable at timeout time: iOS picks `.stale` over `.error` on `flagsData?.context == context` (`FlagsRepository.swift:183`), and Android checks `matchingCachedAssignments` plus `hasLoadedFlagsForContext` (`EvaluationsManager.kt:194-201`). Their storage and concurrency models differ from a blocked Dart isolate, so that is the policy precedent, not a proof.

</details>

<details>
<summary><b>Suggested fix</b> — publish the cached snapshot, then honour the deadline</summary>

Replacing lines 76-90:

```dart
    final matchingCached =
        cached != null && _contextsMatch(cached.context, context)
            ? cached
            : null;
    if (matchingCached != null && !_hasCurrentStateForContext(context)) {
      _state = matchingCached;
    }

    final cacheDeadlineYield = _yieldAfterExpiredDeadline(deadline);
    if (cacheDeadlineYield != null) {
      await cacheDeadlineYield;
    }
    if (token.isCanceled) {
      return;
    }
```

The `token.isCanceled` check at lines 73-75 has no `await` between it and `_state = matchingCached`, so no new *asynchronous* cancellation window opens before the assignment. One caveat, which head shares: `_contextsMatch` walks the caller's `attributes` map (lines 288-300, through `sanitizeJsonValue` and `_jsonValuesMatch`), so a custom `Map` could run app code synchronously between that check and the assignment. Head has no token check between its own comparison and assignment either, so the move neither creates nor closes that window. It does mean the cached publication now happens before the yield and before the second cancellation check at lines 80-82. A superseding context still wins, but the protection is the token cancellation at lines 73-75, not `_hasCurrentStateForContext` — that helper only recognises the *same* context (lines 153-156), so it does not guard against a newer one.

The move also has a cost, though not in the case measured above. There the deadline is already complete before any comparison runs. `_initialize` does its blocking read before its first suspension, so `observe()` expires the deadline at lines 241-247. What the move buys there is publishing the cache before the zero-delay yield, and so before the caller's continuation. Where the block instead falls after `observe()`, the poll's own `expireIfNeeded()` completes the deadline. With the move, the context comparison runs before that. It recurses through the caller's attributes — `sanitizeJsonValue` (`json_value.dart:12-35`) and `_jsonValuesMatch` (lines 303-334) — so a large context can push completion out. I did not measure that.

With this move and nothing else on head, on both SDKs: `dart analyze` clean, the existing suite still 72/72, and the regression above green. Verified with the `[P0]` hunk applied together as well, on both SDKs: analyze clean and 73/73. That count is the existing suite with the `[P0]` comment's two replacement tests, and it does not include the cache regression, which I ran separately.

</details>

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -136,6 +209,72 @@ class FlagsRepository {
...
+
+  void observe(Future<void> operation) {
+    expireIfNeeded();
+    operation.then<void>(
+      (_) => _completeOperation(),
+      onError: _completeOperationWithError,
+    );
+  }
+
+  void _expire() {
```
aarsilv, line 249: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P1]** When this deadline expires, please surface it.

`initialize()` completes normally, so the app cannot tell the budget expired. iOS fails the completion with `FlagsError.initializationTimedOut`. Android calls `onFailure(FlagsInitializationTimeoutException(timeoutMs))`. Both also move the client to `stale` or `error` on an observable. This package has none of the three for the timeout path: no timeout-specific error, no diagnostic log, no observable lifecycle state. An app reading defaults cannot tell a timeout from a slow network.

<details>
<summary><b>What both siblings do on timeout</b></summary>

iOS, `DatadogFlags/Sources/Models/FlagsError.swift:34-35`:

```swift
/// The first evaluation context did not become ready before the configured initialization timeout.
case initializationTimedOut
```

Android, `FlagsInitializationTimeoutException.kt:9-15`:

```kotlin
/**
 * Indicates that the first evaluation context did not complete before the initialization timeout.
 *
 * The request continues and can make the client ready later.
 */
class FlagsInitializationTimeoutException internal constructor(timeoutMs: Long) :
    RuntimeException("Flags initialization timed out after ${timeoutMs}ms")
```

Both then set client state — `stale` if matching cached assignments exist, `error` otherwise, and only if the state is not already ready or stale — and both publish it through `client.state`.

Neither sibling logs or emits telemetry on the timeout path; I checked both. So on those platforms the error plus the state transition are the whole signal, and this port currently has neither half.

</details>

<details>
<summary><b>What this package can do today</b></summary>

`packages/datadog_flags` is pure Dart and has no general-purpose diagnostic logger. It does log: `ExposureLogger.logExposure` (`exposure_logger.dart:17`, called from `default_flags_client.dart:185`) records evaluation exposures, and there is evaluation telemetry. Neither is a channel an internal timeout could write a diagnostic to. `grep -rnE 'internalLogger|developer\.log|debugPrint|telemetry' packages/datadog_flags/lib/` returns 14 matches, all of them doc-comment prose. `InternalLogger` lives in `datadog_flutter_plugin`, which this package does not depend on by design — it is pure Dart so it can be used without Flutter.

There is also no client-state concept. Every export in `lib/datadog_flags.dart` carries a `show` clause, and between them they name exactly ten symbols: `DatadogFlagsConfig`, `DatadogFlagsSite`, `DatadogFlags`, `DatadogFlagsClient`, `FlagDetails`, `DatadogFlagsConfiguration`, `FlagEvaluationError`, `DatadogFlagsStore`, `FlagsData` and `FlagsEvaluationContext`. No state enum and no observable — and `FlagsException` and `FlagsErrorType` are not among them, which matters for option 2 below.

On the OpenFeature question: there is no OpenFeature provider in this repository, so no provider event can fire today. Being precise about the connection, because I overstated it in an earlier draft — a timeout error signals expiry, not late success. This PR's Next Steps item "map late success to OpenFeature `ProviderReady`" needs *state observation*, which none of the options below provide. What a typed error does buy is the other half: Android's bridge suppresses a timeout only when state is already `Ready`/`Stale` and raises a provider error otherwise, and it needs a typed error to make that distinction at all.

</details>

<details>
<summary><b>Three options, all analyzed and tested; each with a gap I name</b></summary>

**1. A log. No API change.** Add `import 'dart:developer' as developer;` to `flags_repository.dart`, then:

```dart
  void _expire() {
    if (_completion.isCompleted) {
      return;
    }
    _expired = true;
    _stopwatch.stop();
    _timer?.cancel();
    _completion.complete();
    developer.log(
      'Flags initialization timed out after ${timeout.inMilliseconds}ms',
      name: 'datadog_flags',
      level: 900,
    );
  }
```

**Measured: `dart analyze` clean, full suite 72/72, on both SDKs.** That the `developer.log` line runs on expiry is control flow, not a measurement — it sits unconditionally after `_completion.complete()`. I did not capture the emitted record itself, and I did not check how it surfaces on every supported runtime. It does add behaviour — a log record — and it reaches a developer tool rather than the app.

**2. A typed error, which is the error half of the sibling shape.** Add a case and a factory in `flags_error.dart`:

```dart
enum FlagsErrorType {
  networkError,
  invalidResponse,
  clientNotInitialized,
  invalidConfiguration,
  initializationTimedOut,
}
```

```dart
  factory FlagsException.initializationTimedOut(
    String message, {
    Object? cause,
  }) {
    return FlagsException(
      FlagsErrorType.initializationTimedOut,
      message,
      cause: cause,
    );
  }
```

Export it, so an app can name the type without importing `src/`:

```dart
export 'src/flags_error.dart'
    show FlagEvaluationError, FlagsErrorType, FlagsException;
```

Then in `_expire`, adding `import 'flags_error.dart';` to `flags_repository.dart` and replacing `_completion.complete()`:

```dart
    _completion.completeError(
      FlagsException.initializationTimedOut(
        'Flags initialization timed out after ${timeout.inMilliseconds}ms',
      ),
    );
```

**Measured: 64/72 — 8 failures, all in `initialization_timeout_test.dart`, none pre-existing.** They `await initialize()` on a timeout path and would need `expectLater(..., throwsA(isA<FlagsException>()))`. One correction to what I posted: I reported `dart analyze` clean for this option and it is not, on either SDK — that was my mistake, not a version difference. The new export makes an existing direct `src/` import redundant, which the analyzer flags as one `info`:

```
info - test/precompute_core_test.dart:11:8 - The import of
'package:datadog_flags/src/flags_error.dart' is unnecessary because all of the used
elements are also provided by the import of 'package:datadog_flags/datadog_flags.dart'.
Try removing the import directive. - unnecessary_import
```

Deleting that import clears it, and it happens to illustrate the next point. The export is not strictly required to name the type — a direct `package:datadog_flags/src/flags_error.dart` import also works — but reaching into `src/` is not a supported import for an app. This is a real behaviour change for any caller that does not catch, and I cannot bound the effect on callers outside this repo. It also gives only the error half of the sibling shape; the observable-state half is a larger change I am not proposing here.

One more thing option 2 needs, which I missed the first time. The core export alone does not reach a Flutter app. `datadog_flags_flutter.dart:8-17` re-exports the core barrel with its own `show` list, and that list names `FlagEvaluationError` but not `FlagsException` or `FlagsErrorType`. With option 2 applied to the core package and nothing else, a file in the Flutter package that imports only the plugin barrel does not compile:

```
error - zz_opt2_probe.dart:4:41 - The name 'FlagsException' isn't defined, so it
can't be used in an 'is' expression. - type_test_with_undefined_name
```

So option 2 is a two-package change: add `FlagsException` and `FlagsErrorType` to the plugin's `show` list too, and raise `packages/datadog_flags_flutter/pubspec.yaml:20` from `datadog_flags: ^1.0.0` to a lower bound that actually contains them. With both of those applied on top of the core change, the Flutter package is clean:

```
$ dart analyze .      # packages/datadog_flags_flutter, probe file importing only the barrel
No issues found!
$ flutter test
00:00 +10: All tests passed!
```

The core package is still 64/72 in that state; option 2's cost is the eight timeout tests, not anything in the plugin. The same `^1.0.0` masking applies here as in option 3's fold below. Published `1.0.1` does contain both types internally, but its barrel exports neither, and it has no `initializationTimedOut` member or factory — so a resolution to `1.0.1` compiles the plugin barrel's `show` clause into an error.

**3. A configuration callback, if changing what `initialize()` returns is unwelcome.** In `flags_configuration.dart`:

```dart
  /// Called when [initializationTimeout] expires before the first
  /// initialization completes. Use it to log or report slow startup.
  final void Function(String clientName, Duration timeout)?
      onInitializationTimeout;
```

with `this.onInitializationTimeout,` in the constructor, forwarded from `datadog_flags.dart`, and an `onExpired` hook on `_InitializationDeadline`. It also needs a line at the Flutter plugin's field-copy site (`datadog_flags_plugin.dart:146-158`), or a Flutter-configured callback is silently dropped:

```dart
      onInitializationTimeout: _flagsConfiguration.onInitializationTimeout,
```

The callback needs containing, though not for the reason I first gave. `_completion.complete()` runs before `onExpired()`, so `initialize()` has already completed normally by then. Without a catch, a synchronous app throw escapes `_expire` through the timer callback as an uncaught zone error:

```
00:00 +1: a throwing app callback does not break initialize()
00:00 +1 -1: a throwing app callback does not break initialize() [E]
  Bad state: app bug
  package:datadog_flags/src/flags_repository.dart 270:16  _InitializationDeadline._expire
```

The trace shows it leaving `_expire`, which calls the app back on the line after `_completion.complete()`. So the catch:

```dart
    _timer?.cancel();
    _completion.complete();
    final onExpired = this.onExpired;
    if (onExpired != null) {
      // An application callback must not break initialization.
      try {
        onExpired();
      } catch (_) {
        return;
      }
    }
```

That catch covers synchronous throws only. The field type `void Function(String, Duration)` also accepts an `async` closure, whose throw becomes a rejected future the catch never sees. I tested that too, and it escapes:

```
00:00 +2 -1: an async app callback that throws is NOT contained by the catch [E]
  Bad state: async app bug
  package:datadog_flags/src/flags_repository.dart 272:18  _InitializationDeadline._expire
```

So option 3 should either document the callback as synchronous-only, or take `FutureOr<void>` and catch what it returns. I did not implement the second.

**Measured on the core package: `dart analyze` clean, the existing suite still 72/72, plus two tests of my own — separate from that 72 — confirming the callback fires with the client name and budget and that a synchronous throwing callback no longer escapes.** Both on Dart 3.6.2 and again on Dart 3.13.3. The plugin line I have now compiled as well; see the last fold.

Of the three I would take 1 or 3. Neither changes what existing callers observe from `initialize()`, though 1 adds a log record and 3 adds a public field. Option 2 is closest to the siblings but only on the error half, and it wants your call on the break.

</details>

<details>
<summary><b>Option 3's plugin line, now compiled — and why it is load-bearing</b></summary>

I retract the caveat I first posted here. I had read `datadog_flags_plugin.dart:57` and reasoned that the reference would resolve, without compiling it. It does compile, and the reasoning held.

One thing that compile hides, and it is a real gap in option 3. I resolved the packages with a hand-written `pubspec_overrides.yaml` pointing `datadog_flags` and `datadog_flutter_plugin` at the local paths, so the analyzer saw this PR's core tree. But `packages/datadog_flags_flutter/pubspec.yaml:20` still asks for `datadog_flags: ^1.0.0`. A consumer can satisfy that with a published 1.0.x that has no `onInitializationTimeout`, and the forwarding line then fails to compile for them. If you take option 3, that lower bound needs raising to the release that ships the field. The core package is at `1.1.0` here.

In `packages/datadog_flags_flutter` with all four production-file changes applied, on Flutter 3.47.4 / Dart 3.13.3:

```
$ dart analyze .
No issues found!
$ flutter test
00:00 +10: forwards onInitializationTimeout to the flags configuration
00:00 +11: All tests passed!
```

That is the package's 10 existing tests plus one of mine, which asserts the forwarding actually happens rather than merely compiling:

```dart
  test('forwards onInitializationTimeout to the flags configuration', () async {
    DatadogFlagsConfiguration? capturedConfiguration;
    final flags = _mockFlags(onEnable: (configuration) {
      capturedConfiguration = configuration;
    });
    final calls = <String>[];
    void onTimeout(String clientName, Duration timeout) {
      calls.add('$clientName/${timeout.inMilliseconds}');
    }

    when(() => mockSdk.configuration).thenReturn(
      DatadogConfiguration(
        clientToken: 'client-token',
        env: 'prod',
        site: DatadogSite.us1,
      ),
    );

    final plugin = _plugin(
      mockSdk,
      flags: flags,
      flagsConfiguration: DatadogFlagsConfiguration(
        onInitializationTimeout: onTimeout,
      ),
    );
    plugin.initialize();
    await plugin.ready;

    final forwarded = capturedConfiguration!.onInitializationTimeout;
    expect(forwarded, isNotNull,
        reason: 'plugin dropped the configured timeout callback');
    forwarded!('shop', const Duration(milliseconds: 2500));
    expect(calls, ['shop/2500']);
  });
```

Then I deleted only the forwarding line, to check the test discriminates. It does — and the more useful half is that the analyzer stays silent:

```
$ dart analyze .        # with the forwarding line deleted
No issues found!
$ flutter test
00:00 +10 -1: forwards onInitializationTimeout to the flags configuration [E]
  Expected: not null
    Actual: <null>
  plugin dropped the configured timeout callback
```

So omitting that line costs nothing at compile time and silently drops the callback for every Flutter caller. If you take option 3, that line wants a test; nothing else at the copy site would catch it.

</details>

packages/datadog_flags/test/initialization_timeout_test.dart
```diff
@@ -0,0 +1,636 @@
...
+      dateProvider: DateTime.now,
+      initializationTimeout: configuration.initializationTimeout,
+      scheduleInitializationTimeout: (_, __) => _TestTimer(),
+    );
+    final context = FlagsEvaluationContext(
+      targetingKey: 'user-123',
+      attributes: {'slow': _SlowIterable(const Duration(milliseconds: 10))},
+    );
+
+    // The fake timer never runs. This proves that the elapsed-time check still
```
aarsilv, line 453: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** Nothing asserts the budget that reaches the timer, so an inflated budget stays green.

I quadrupled the deadline passed to `_InitializationDeadline`. I separately doubled the duration handed to `schedule`. Both left all 72 tests passing, on `dart:3.6.2` and again on Dart 3.13.3. The configured values *are* asserted at lines 19-29, and several tests assert state rather than only completion. But no test checks the budget that arms the timer, and all five injected seams discard the `Duration`.

<details>
<summary><b>Mutations that should have been caught and were not</b></summary>

| Mutation | Result |
| --- | --- |
| `_InitializationDeadline(timeout * 4, scheduleInitializationTimeout)` | **72 green** |
| `schedule(timeout * 2, _expire)` | **72 green** |
| positive control — never apply the timeout at all | 8 red |

The positive control is the reassuring half: the suite genuinely exercises timeout behaviour, and those eight tests do more than watch for completion — they hold body, read and write futures pending and assert readiness and state around them. The narrower gap is that the deadline's magnitude is never pinned.

Being precise about the slack, since I was sloppy about this in an earlier draft. The tests that let a real timer fire use **5 ms** budgets inside a 1 s `.timeout(...)` wrapper; the two fake-timer tests use 1 ms budgets and never fire their callback. Others configure a 5 s budget and invoke the captured callback by hand, so their wrapper is not deadline slack at all. The `timeout * 4` mutation survives because 4x a 5 ms budget is still far inside that 1 s wrapper. The two fake-timer tests do constrain magnitude a little, through their 10 ms synchronous blocks and the stopwatch comparison — but 4x a 1 ms budget still clears 10 ms, so they do not catch it either.

</details>

<details>
<summary><b>The fix needs no wall clock</b></summary>

The seam to assert this deterministically already exists at `flags_repository.dart:29` — the tests just throw away its first argument. Asserting it is instant and exact, and that particular assertion has no timing dependence at all, so slow CI cannot change its result. I understand the current slack absorbs slow runners; this removes the need for slack rather than tightening it.

Where a real timer has to be exercised, a **lower** bound is one-sided: a slow machine makes `elapsed >= budget` more true, never less. It does not catch an inflated budget — only the seam assertion does — so it is complementary coverage, not the mutation killer.

</details>

<details>
<summary><b>Suggested tests</b> — verified in this file, using its existing helpers</summary>

First give `_TestTimer` a counter so cancellation is assertable:

```dart
class _TestTimer implements Timer {
  var _isActive = true;
  var cancelCount = 0;

  @override
  bool get isActive => _isActive;

  @override
  int get tick => 0;

  @override
  void cancel() {
    cancelCount += 1;
    _isActive = false;
  }
}
```

Then three tests:

```dart
  test('schedules the initialization timer for exactly the configured budget',
      () async {
    const budget = Duration(milliseconds: 250);
    final scheduled = <Duration>[];
    final timers = <_TestTimer>[];
    void Function()? timeoutAction;
    final response = Completer<http.Response>();
    final httpClient = MockClient((_) => response.future);
    addTearDown(() {
      if (!response.isCompleted) {
        response.complete(
          http.Response(jsonEncode(_assignmentsResponse()), 200),
        );
      }
    });
    final configuration = _configuration(
      httpClient: httpClient,
      initializationTimeout: budget,
    );
    final repository = FlagsRepository(
      clientName: DatadogFlags.defaultClientName,
      fetcher: FlagAssignmentsFetcher(
        datadogConfig: _datadogConfig,
        configuration: configuration,
        httpClient: httpClient,
      ),
      dateProvider: DateTime.now,
      initializationTimeout: configuration.initializationTimeout,
      scheduleInitializationTimeout: (duration, action) {
        scheduled.add(duration);
        timeoutAction = action;
        final timer = _TestTimer();
        timers.add(timer);
        return timer;
      },
    );

    final initialization = repository.initialize(_context);

    expect(scheduled, const [budget]);

    timeoutAction!();
    await initialization.timeout(const Duration(seconds: 5));

    expect(repository.flagAssignment('show-paywall'), isNull);
    expect(timers.single.cancelCount, 1);
  });

  test('cancels the initialization timer when initialization wins', () async {
    final timers = <_TestTimer>[];
    final httpClient = MockClient((_) async {
      return http.Response(jsonEncode(_assignmentsResponse()), 200);
    });
    final configuration = _configuration(
      httpClient: httpClient,
      initializationTimeout: const Duration(seconds: 30),
    );
    final repository = FlagsRepository(
      clientName: DatadogFlags.defaultClientName,
      fetcher: FlagAssignmentsFetcher(
        datadogConfig: _datadogConfig,
        configuration: configuration,
        httpClient: httpClient,
      ),
      dateProvider: DateTime.now,
      initializationTimeout: configuration.initializationTimeout,
      scheduleInitializationTimeout: (_, __) {
        final timer = _TestTimer();
        timers.add(timer);
        return timer;
      },
    );

    await repository.initialize(_context).timeout(const Duration(seconds: 5));

    expect(repository.flagAssignment('show-paywall'), isNotNull);
    expect(timers.single.cancelCount, 1);
  });

  test('bounds the first initialization with the real timer', () async {
    const budget = Duration(milliseconds: 250);
    final response = Completer<http.Response>();
    final httpClient = MockClient((_) => response.future);
    addTearDown(() {
      if (!response.isCompleted) {
        response.complete(
          http.Response(jsonEncode(_assignmentsResponse()), 200),
        );
      }
    });
    final configuration = _configuration(
      httpClient: httpClient,
      initializationTimeout: budget,
    );
    final repository = FlagsRepository(
      clientName: DatadogFlags.defaultClientName,
      fetcher: FlagAssignmentsFetcher(
        datadogConfig: _datadogConfig,
        configuration: configuration,
        httpClient: httpClient,
      ),
      dateProvider: DateTime.now,
      initializationTimeout: configuration.initializationTimeout,
    );

    final stopwatch = Stopwatch()..start();
    await repository.initialize(_context);
    stopwatch.stop();

    expect(stopwatch.elapsed, greaterThanOrEqualTo(budget));
    expect(repository.flagAssignment('show-paywall'), isNull);
  });
```

The third has no `.timeout()` wrapper on purpose: if the deadline never fires, the suite's own timeout reports a hang rather than an assertion failing for the wrong reason.

</details>

<details>
<summary><b>Measured cost and flake rate of exactly that text</b></summary>

Pasted into this file alone — no production change, and none of the `[P0]` comment's tests — run first in `dart:3.6.2` and then again on Dart 3.13.3:

- `dart analyze` clean, 14/14 tests pass. Both SDKs.
- Kills both mutations above: `timeout * 4` and `schedule(timeout * 2)` each turn it red. Both SDKs.
- **30 consecutive runs with 6 competing CPU hogs: 30 pass, 0 fail on each SDK.** The core counts differ, so the contention does too: 16 cores on the 3.13.3 host, 4 in the 3.6.2 container. The second is genuine saturation. Observed stability under load, not a proof of immunity.
- Cost, re-measured the same way on both SDKs — median of 5 whole-`dart test` process runs, which includes VM startup. `dart:3.6.2`: 737 ms to 1025 ms. Dart 3.13.3: 829 ms to 1133 ms. So roughly 290 ms and 300 ms added. A second correction to what I posted: I gave 167 ms there, from a file-runtime reading I cannot reproduce under any method I can now state. The reporter's own elapsed on `dart:3.6.2` says 423 ms to 686 ms, about 260 ms, so 167 ms was too low. I did not measure the three tests individually, so I cannot attribute the cost to any one of them.

</details>

packages/datadog_flags/lib/src/flags_configuration.dart
```diff
@@ -27,6 +31,22 @@ final class DatadogFlagsConfiguration {
   /// Additional headers sent with precompute assignment requests.
   final Map<String, String>? customFlagsHeaders;
 
+  /// Maximum wall-clock time to wait for the first evaluation context.
```
aarsilv, line 34: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P3]** Two wording fixes here: "maximum", and the absent upper bound.

"Maximum" promises a hard bound the implementation does not give. With a 50 ms budget and 300 ms of synchronous work, I measured `initialize()` returning at 312 ms. The two SDKs agree to within a millisecond. This PR's own `counts synchronous request encoding in the total budget` test relies on that mechanism. Separately, `Duration(minutes: 50)` is accepted verbatim, while `evaluationFlushInterval` in this class is clamped to 1-60 s. Neither sibling clamps, so I suggest wording, not a clamp.

<details>
<summary><b>Why "maximum" overstates it</b></summary>

A timer cannot fire while synchronous encoding or decoding blocks the isolate. The stopwatch checks can run sooner — `observe()` calls `expireIfNeeded()` synchronously, right after `_initialize` returns control — but they still cannot run *during* the blocking chunk. So the deadline takes effect after the budget elapses, late by most of that chunk.

Measured with a real timer, 50 ms budget against a 300 ms block, median of 7 runs on each SDK:

| Scenario | `dart:3.6.2` | Dart 3.13.3 |
| --- | --- | --- |
| 300 ms of synchronous request encoding | 311 ms | 312 ms |
| 300 ms of synchronous JSON decoding | 304 ms | 304 ms |

A correction to the figures I posted. I first reported 326 ms and 311 ms for `dart:3.6.2`. Re-running that harness gives 311 ms and 304 ms, so the earlier pair came from a noisier run and I am replacing it. The conclusion is unchanged and now identical across both SDKs.

This is an isolate-wide limitation rather than something specific to this implementation: I measured three alternative designs and none avoided the overrun. On the decoding path all three return at 300 ms against the current design's 304 ms, on both SDKs. On the encoding path all three are slightly *slower* — 314 ms on `dart:3.6.2` and 316-318 ms on Dart 3.13.3. None of the three escapes it, and all three keep the work on one isolate. I did not measure a design that moves the blocking work off the isolate. The doc comment is what an engineer sizing a splash-screen budget will read, and "maximum" is the word that misleads there.

</details>

<details>
<summary><b>The clamp asymmetry, and why I am not asking for a clamp</b></summary>

`evaluationFlushInterval` is coerced to 1-60 s by the getter at `flags_configuration.dart:96-103` and the README documents that bound. `initializationTimeout` has no equivalent, so a mistyped `Duration(minutes: 50)` is honoured verbatim and the intended 50 ms bound is simply gone. Initialization still runs; if it finishes first the timer is cancelled, and otherwise the deadline fires — normally from the timer, or from an elapsed-time check when synchronous work kept the timer from running.

`Timer` construction did not throw for any of the five `Duration` values I tried, up to `Duration(microseconds: 9223372036854775807)`. That is all I verified — construction on the Dart VM, on those five values. I did not test whether such a timer fires correctly, so I am not claiming there is no hazard, only that I found no construction-time one. iOS does need an explicit `UInt64.max` saturation guard in `initializationTimeoutDeadline` for its own clock arithmetic.

Both siblings leave the value unclamped — iOS disables on `nil`, non-finite, zero and negative only (`FlagsRepository.swift:167-170`); Android coerces with `takeIf { it > 0 }` (`FlagsConfiguration.kt:169-170`). Neither reuses its `evaluationFlushInterval` range here, and neither documents an upper bound either. So parity supports leaving it unclamped; documenting it is my suggestion, not sibling precedent.

</details>

<details>
<summary><b>Suggested prose</b> — four places, not two</summary>

This doc comment:

```dart
  /// Wall-clock budget for the first evaluation context.
  ///
  /// The SDK stops waiting as soon as it can after this budget elapses.
  /// Synchronous encoding or decoding can block the Dart isolate, so the wait
  /// can be longer than the value set here.
  ///
  /// This value is one budget for the complete initialization operation. The
  /// SDK does not restart it for each stage. It includes loading stored
  /// assignments, encoding the request, fetching assignments, reading the
  /// response body, decoding JSON, storing assignments, and making the
  /// assignments available for evaluation. It does not change the HTTP
  /// client's timeout. The assignment operation continues after this timeout
  /// and can make assignments available when it completes.
  ///
  /// The SDK does not limit how large this value can be.
  ///
  /// The timeout applies only to the first
  /// [DatadogFlagsClient.initialize] call for each client. A `null`, zero, or
  /// negative value disables the timeout. If a later call supersedes the first
  /// call, only the first call remains bounded by this timeout.
```

The same word appears in three more places that this fix should cover — `flags_configuration.dart:16` on `defaultInitializationTimeout`, `packages/datadog_flags/README.md:131`, and `packages/datadog_flags_flutter/README.md:94`. For the core README bullet:

```markdown
- `initializationTimeout` is the wall-clock budget for the first context
  initialization. The SDK stops waiting as soon as it can after the budget
  elapses. The default is five seconds, and the SDK does not limit how large
  it can be. Set it to `null`, zero, or a negative value to disable it.
```

Verified on Dart 3.13.3: with that Dart block spliced into `flags_configuration.dart` and nothing else changed, `dart analyze` reports `No issues found!` and the suite is 72/72. That checks syntax, not the prose.

Everything else I checked in these doc blocks matched the code: the stage list, "does not change the HTTP client's timeout", the first-call scope, and the disable semantics. One exception worth naming, because it belongs to the `[P0]` comment rather than here: "a late successful response still makes assignments available" does not hold when a store write never settles.

</details>

fuzzybinary:
> The timeout can complete only when the isolate yields. At that point, the deadline must win before more work starts or assignments become visible. This constraint is why `Future.any` alone does not meet the product goal. A microtask can continue initialization before a delayed timer runs. The stopwatch checks enforce the wall-clock budget at these boundaries.

So, I think I understand the expected requirement, but I'm very curious how much of an issue it actually is in practice, whether it's actually achievable, and whether it's worth the added complexity. 

The only time we check the timeout stopwatch is after an `await` has returned from the microtask queue, and they go behind the Timer expiration. The only way we potentially early out is if one of these tasks completes *just* ahead of the timer, and your stopwatch is showing a time just slightly ahead of the timer expiration. The process then gets *just* enough time to get to the next `await` before the timer expires.

But, if you're concerned with the main isolate being starved because of one of these taking to longer between `await` calls, this can happen in the `stopwatch` situation as well. If you check the stopwatch *just* before one of these long processing events, you're delayed the same amount of time, and the timer triggers during the next drain.

You also can't control what else is in the microtask queue. When you `await Future.delayed(Duration.zero)` you effectively drain the microtask queue, and execution only resumes only after the current micotask queue drains. If something is in front of you that starves the main thread, you've blown your wall clock budget.

Claude puts it this way:
> If the stall happens before a checkpoint, the manual check does nothing extra a real timer wouldn't already do (the timer's overdue and fires the moment the isolate frees up, same as it would without any stopwatch). If the stall happens between a checkpoint and the next await, the manual check already ran and can't retroactively help. So the stopwatch buys protection only in the sliver of cases where completion lands in the same turn the timer was also about to fire — which is indistinguishable, from the customer's perspective, from "initialization finished right around the timeout," a case the product goal already says is fine either way.

Real enforcement of "true" wallclock time would require using actual OS-level concurrency (aka a separate isolate) which I would think is out of scope.

You've already achieved the fact that the customer shouldn't see "in flight" flags by delaying the write to `_state` until initialization is complete. I think using the simpler `Future.any` approach is going to be way simpler and will achieve the expected close to wallclock time delay, and any cases where it doesn't would be possible in the more complicated solution as well.

packages/datadog_flags/lib/src/flags_repository.dart
leoromanovsky: Addressed in a20d8b66. initialize() now throws a public FlagsInitializationTimeoutException with the client name and configured duration when the timeout wins. The assignment operation continues, matching stored assignments remain usable, and late success can still publish assignments. The exception is exported through both the Dart and Flutter APIs, with customer examples and regression coverage.

aarsilv (approved):
Thanks! All issues my review flow had identified have been addressed, thanks for iterating!

Checking the new code i'm just seeing a p2 and p3. I'll post if verified with code execution, but they really are not a big deal, and you can take or leave as you see fit.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -10,41 +10,82 @@ import 'package:meta/meta.dart';
...
+import 'flags_error.dart';
 import 'flags_store.dart';
 import 'json_value.dart';
 
 class FlagsRepository {
   static const defaultStoreReadTimeout = Duration(milliseconds: 100);
 
+  // Keep late writes ordered when SDK reconfiguration replaces a repository.
+  // Expando keeps the queue scoped to the store identity without retaining it.
+  static final Expando<Map<String, _CacheOperationQueue>>
```
aarsilv, line 22: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** A store write that never settles now blocks every later repository for this client, permanently.

The queue is `static` and keyed on `(store, clientName)`, so it outlives the repository that created it. That is the intent — the comment above says it keeps late writes ordered across reconfiguration. The gap is that nothing bounds the wait. If a customer store returns a write future that never completes, the next repository for the same store and client name chains behind it at `flags_repository.dart:147` and its write never reaches the store. On `develop` the chain is a plain instance field (`_cacheOperation`, `:29`, chained in `_enqueueCacheOperation` at `:132-136`), so a replacement repository started with an empty chain. This one does not, for the life of the process.

<details>
<summary><b>Executed evidence</b> — the block is new, shown against a one-line control</summary>

A store whose `write` returns `Completer<void>().future` and counts calls. Repository A initializes, then a **fresh** repository is built for the same store and `clientName` and initializes. `Flutter 3.47.4 • channel stable`, `Darwin arm64`, `dart test`:

```text
00:08 +0 -1: P1 hung write blocks a later repository write on the same store [E]
  Expected: a value greater than <1>
    Actual: <1>
     Which: is not a value greater than <1>
  the second repository's write should reach the store
```

Control, one line changed to give each repository its own chain — `}) : _cacheOperations = _CacheOperationQueue();`, matching `develop`'s per-instance semantics — same test otherwise:

```text
00:08 +1: P1 hung write blocks a later repository write on the same store
00:08 +1: All tests passed!
```

So the shared registry is what causes the block.

</details>

<details>
<summary><b>Suggested fix</b> — applied and run, 74/74 green</summary>

Make ordering best-effort rather than unbounded. In `_CacheOperationQueue`:

```dart
class _CacheOperationQueue {
  // Ordering is best-effort: a store write that never settles must not block
  // every later write for this client for the life of the process.
  static const _predecessorTimeout = Duration(seconds: 5);

  Future<void> _operation = Future<void>.value();

  Future<void> enqueue(Future<void> Function() operation) {
    final next = _operation
        .timeout(_predecessorTimeout, onTimeout: () {})
        .then((_) => operation());
    _operation = next.catchError((_) {});
    return next;
  }
}
```

With this applied: `P1` passes, `dart analyze .` → `No issues found!`, and the package suite is `00:01 +74: All tests passed!`. Ordering coverage still passes by name — `00:01 +3: reset waits for pending store writes before deleting assignments` and `00:00 +1: publishes assignments while persistent storage remains in progress`. I ran this patch on its own, not only alongside my other suggestion. The 5s value is a guess; pick your own.

</details>

One nearby behaviour I checked and would **not** raise: `initialize()` awaiting a hung write is not new. `develop` already sets state then awaits the write in the same order, at `:71-72`.

Scope, measured rather than assumed: a different `clientName` on the same store is unaffected — same probe with `client-a` then `client-b` gives `00:00 +1: P3 a different clientName on the same store is unaffected`. The barrel exports the store interface only (`datadog_flags.dart:20`), and the one internal implementation, `InMemoryDatadogFlagsStore`, has a `write` that always completes — so reaching this needs a customer store whose write future never settles. I did not test a store shared across isolates, and I did not run the iOS or Android integration suites.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -10,41 +10,82 @@ import 'package:meta/meta.dart';
...
+
+    final timeoutCompletion = Completer<void>();
+    final timer = scheduleInitializationTimeout(
+      timeout,
+      () => timeoutCompletion.completeError(
+        FlagsInitializationTimeoutException(
+          clientName: clientName,
+          timeout: timeout,
+        ),
+        StackTrace.current,
```
aarsilv, line 75: 🤖 **Comment from Claude working with Aaron Silverman:**

**[P2]** The timeout's stack trace has no customer frame, because it is captured inside the timer callback.

`StackTrace.current` here runs on the timer's zone callback, long after `initialize()` returned. The trace that reaches the caller is SDK internals and `dart:async` plumbing only. It matters beyond local debugging, because `DatadogSdk.runApp` installs a handler that passes uncaught zone errors to `rum?.addErrorInfo(..., RumErrorSource.source, stackTrace: st)` (`datadog_flutter_plugin.dart:156-162`). By inspection of that code, an un-awaited timeout under `runApp` with RUM enabled would report this trace. I read that path but did not execute it, and did not capture a RUM event.

<details>
<summary><b>Executed evidence</b> — captured trace on head</summary>

A store whose `read` never settles, `initializationTimeout: 50ms`, the error's `StackTrace` printed verbatim:

```text
#0      FlagsRepository.initialize.<anonymous closure> (package:datadog_flags/src/flags_repository.dart:75:20)
#1      _rootRun (dart:async/zone_root.dart:27:47)
#2      _CustomZone.run (dart:async/zone.dart:810:35)
#3      _CustomZone.runGuarded (dart:async/zone.dart:702:7)
#4      _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:743:23)
#5      _rootRun (dart:async/zone_root.dart:35:13)
#6      _CustomZone.run (dart:async/zone.dart:810:35)
#7      _CustomZone.bindCallback.<anonymous closure> (dart:async/zone.dart:726:23)
#8      Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:18:15)
#9      _Timer._runTimers (dart:isolate-patch/timer_impl.dart:423:19)
#10     _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:454:5)
#11     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192:12)
```

The calling test file appears nowhere.

</details>

<details>
<summary><b>Suggested fix</b> — applied and run, 74/74 green</summary>

Capture the stack where the caller is still on it, then pass it into the callback:

```dart
    final timeoutCompletion = Completer<void>();
    final callSite = StackTrace.current;
    final timer = scheduleInitializationTimeout(
      timeout,
      () => timeoutCompletion.completeError(
        FlagsInitializationTimeoutException(
          clientName: clientName,
          timeout: timeout,
        ),
        callSite,
      ),
    );
```

Same probe with this applied:

```text
#0      FlagsRepository.initialize (package:datadog_flags/src/flags_repository.dart:68:33)
#1      main.<anonymous closure> (file:///private/tmp/pr1155-fix/packages/datadog_flags/test/zz_probe_test.dart:29:7)
```

(Test-harness frames follow; the caller is frame #1.)

`dart analyze .` → `No issues found!`, package suite `00:01 +74: All tests passed!`. I ran this patch on its own, not only alongside my other suggestion.

</details>

Cost is one `StackTrace.current` per `FlagsRepository` instance, on its first `initialize()`, and only when a timeout is configured (`_didStartInitialization` is an instance field at `:41`). I did not measure that cost, and I did not check how the native iOS or Android SDKs populate their timeout traces.

packages/datadog_flags/lib/src/flags_repository.dart
```diff
@@ -10,41 +10,82 @@ import 'package:meta/meta.dart';
...
 
+    final timeout = _takeInitializationTimeout();
+    if (timeout == null) {
+      return _initialize(context, token);
+    }
+
+    final timeoutCompletion = Completer<void>();
+    final timer = scheduleInitializationTimeout(
+      timeout,
+      () => timeoutCompletion.completeError(
```
fuzzybinary, line 70: I'm of very split mind here on this throwing, but I'm leaving it to you to make the final decision.

This is not really an error or an exceptional circumstance, and a user failing to catch this exception as part of initialize is going to potentially interrupt their entire initialization flow as a result of this throw.  In a language like Java with checked exceptions, this is less of an issue, but here it's potentially a silent killer.

In my opinion, it might be better for `initialize` to return an boolean or an enum specifying the result of the initialization, with `.timeout` being one of the options. But I'll leave the final design decision up to you.

fuzzybinary (approved):

